PR #1, opened 2023-02-07
Turning the allocator into an overt exercise in data structure composition, part 1

nwf-msr:
I hope that the include/ds files are the kinds of things that might be useful more generally for tiny, embedded systems. The goal here is to offer high-level abstractions without committing to bit-level representations, to some extent mechanizing the semantic gap between diagrams on paper and bit twiddling. I was not aware of an existing library, so I wrote my own; apologies if there's something I should have used instead.

There's still plenty of clean-up possible in the allocator.  In particular, IMHO, it'd be good to get some things on the alloc path more clearly communicating in terms of `MChunkHeader*` / `MChunk*` rather than `void *`s pointing just past the headers.

### DIFF
--- a/sdk/core/allocator/alloc.h
+++ b/sdk/core/allocator/alloc.h
@@ -313,13 +313,6 @@ MChunkHeader
 		return isPrevInUse;
 	}
 
-	MChunkHeader *chunk_plus_offset(size_t s)
-	{
-		CHERI::Capability<void> ptr{this};
-		ptr.address() += s;
-		return ptr.cast<MChunkHeader>();
-	}
-
 	// size of the previous chunk
 	size_t prevsize_get()
 	{
@@ -517,13 +510,6 @@ MChunk
 		return header.is_prev_in_use();
 	}
 
-	// Treat the thing at thischunk + s as a chunk.
-	MChunk *chunk_plus_offset(size_t s)
-	{
-		CHERI::Capability<void> ptr{this};
-		ptr.address() += s;
-		return ptr.cast<MChunk>();
-	}
 	// Returns the next adjacent chunk.
 	MChunk *chunk_next()
 	{
@@ -620,7 +606,7 @@ MChunk
 		return ring.cell_next() == &p->ring;
 	}
 
-	MChunk *split(size_t offset)
+	__always_inline MChunk *split(size_t offset)
 	{
 		return MChunk::from_header(header.split(offset));
 	}
@@ -1773,7 +1759,6 @@ class MState
 
 		if (RTCHECK(ok_address(v->ptr())))
 		{
-			MChunk *r = v->chunk_plus_offset(nb);
 			Debug::Assert(v->size_get() == rsize + nb,
 			              "Chunk {} size is {}, should be {}",
 			              v,
@@ -1782,25 +1767,25 @@ class MState
 			Debug::Assert(v->is_prev_in_use(),
 			              "Free chunk {} follows another free chunk",
 			              v);
-			if (RTCHECK(v->ok_next(r)))
+
+			unlink_large_chunk(v);
+
+			if (rsize >= MinChunkSize)
 			{
-				unlink_large_chunk(v);
-				if (rsize < MinChunkSize)
-				{
-					v->in_use_chunk_set(rsize + nb);
-				}
-				else
-				{
-					v->in_use_chunk_set(nb);
-					/*
-					 * The remainder is big enough to to used by another
-					 * allocation, place it into the free list.
-					 */
-					r->free_chunk_set(rsize);
-					insert_chunk(r, rsize);
-				}
-				return chunk2mem(v);
+				/*
+				 * The remainder is big enough to to used by another
+				 * allocation, place it into the free list.
+				 */
+				auto r = v->split(nb);
+				insert_chunk(r, rsize);
 			}
+
+			/*
+			 * Having split off any residual, mark this chunk in use.
+			 */
+			v->in_use_set();
+
+			return chunk2mem(v);
 		}
 
 		corruption_error_action();
@@ -2096,17 +2081,12 @@ class MState
 				auto   p        = unlink_first_small_chunk(i);
 				size_t rsize    = small_index2size(i) - nb;
 
-				if (rsize < MinChunkSize)
+				if (rsize >= MinChunkSize)
 				{
-					p->in_use_set();
-				}
-				else
-				{
-					p->in_use_chunk_set(nb);
-					auto r = p->chunk_plus_offset(nb);
-					r->free_chunk_set(rsize);
+					auto r = p->split(nb);
 					insert_small_chunk(r, rsize);
 				}
+				p->in_use_set();
 				p->metadata_clear();
 				mem = chunk2mem(p);
 
@@ -2145,15 +2125,14 @@ class MState
 	// Allocate memory with specific alignment.
 	void *mspace_memalign(size_t bytes, size_t alignment)
 	{
-		size_t                  nb;      // padded request size
-		CHERI::Capability<void> m;       // memory returned by malloc call
-		MChunk                 *p;       // corresponding chunk
-		char                   *brk;     // alignment point within p
-		MChunk                 *newp;    // chunk to return
-		size_t                  newsize; // its size
-		size_t  leadsize;                // leading space before alignment point
-		MChunk *remainder;               // spare room at end to split off
-		size_t  remainderSize;           // its size
+		size_t                  nb;   // padded request size
+		CHERI::Capability<void> m;    // memory returned by malloc call
+		MChunk                 *p;    // corresponding chunk
+		char                   *brk;  // alignment point within p
+		MChunk                 *newp; // chunk to return
+		size_t  leadsize;             // leading space before alignment point
+		MChunk *remainder;            // spare room at end to split off
+		size_t  remainderSize;        // its size
 		size_t  size;
 
 		nb = pad_request(bytes);
@@ -2189,47 +2168,34 @@ class MState
 			 * leader, we can move to the next aligned spot -- we've
 			 * allocated enough total room so that this is always possible.
 			 */
-			auto alignedAddress{m};
-			alignedAddress.address() =
-			  (m.address() + alignment - 1) & -static_cast<ssize_t>(alignment);
-			brk = reinterpret_cast<char *>(mem2chunk(alignedAddress));
-			if (ds::pointer::diff(p, brk) < MinChunkSize)
+			size_t alignpad =
+			  alignment - (m.address() & static_cast<ssize_t>(alignment - 1));
+			if (alignpad < MinChunkSize)
 			{
-				brk += alignment;
+				alignpad += alignment;
 			}
 
-			newp     = reinterpret_cast<MChunk *>(brk);
-			leadsize = ds::pointer::diff(p, brk);
-			newsize  = p->size_get() - leadsize;
-
-			// Give back leader, use the rest.
-			newp->in_use_chunk_set(newsize);
-			p->in_use_chunk_set(leadsize);
+			auto r = p->split(alignpad);
 			/*
-			 * This pointer is entirely internal. No need to go through
-			 * quarantine. Same for another call below.
+			 * XXX This is excessive, but because r is marked as in-use, thanks
+			 * to coming out of malloc, it makes some amount of sense to go back
+			 * through the free path.  But we know, for example, that the
+			 * contents are zeroed, the revocation bitmap is not set, that
+			 * coalescing is not possible (because p is the predecessor and is
+			 * in use and because if the successor were free, it would have been
+			 * following another free chunk.
 			 */
 			mspace_free_internal(p);
-			p = newp;
-
-			Debug::Assert(newsize >= nb &&
-			                (chunk2mem(p).address() % alignment) == 0,
-			              "Chunk {} does not meet the size ({}) or alignment "
-			              "({}) requirements",
-			              p,
-			              nb,
-			              alignment);
+			p = r;
 		}
 
 		// Also give back spare room at the end.
 		size = p->size_get();
 		if (size > nb + MinChunkSize)
 		{
-			remainderSize = size - nb;
-			remainder     = p->chunk_plus_offset(nb);
-			remainder->in_use_chunk_set(remainderSize);
-			p->in_use_chunk_set(nb);
-			mspace_free_internal(remainder);
+			auto r = p->split(nb);
+			/* XXX see above */
+			mspace_free_internal(r);
 		}
 
 		ok_in_use_chunk(p);
--- a/sdk/core/allocator/alloc.h
+++ b/sdk/core/allocator/alloc.h
@@ -1123,10 +1123,10 @@ class MState
 	{
 		return true;
 	}
-	void ok_any_chunk(MChunk *p) {}
-	void ok_in_use_chunk(MChunk *p) {}
-	void ok_free_chunk(MChunk *p) {}
-	void ok_malloced_chunk(void *mem, size_t s) {}
+	void ok_any_chunk(MChunkHeader *p) {}
+	void ok_in_use_chunk(MChunkHeader *p) {}
+	void ok_free_chunk(MChunkHeader *p) {}
+	void ok_malloced_chunk(MChunkHeader *p, size_t s) {}
 	void ok_treebin(BIndex i) {}
 	void ok_smallbin(BIndex i) {}
 	void ok_malloc_state() {}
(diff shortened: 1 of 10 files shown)


sdk/include/ds/linked_list.h
```diff
@@ -0,0 +1,642 @@
...
+	{
+		/**
+		 * The primitive, required, abstract interface to our cons cells.
+		 *
+		 * All methods are "namespaced" with `cell_` to support the case where
+		 * the encoded forms are also representing other state (for example,
+		 * bit-packed flags in pointer address bits).
+		 */
+		template<typename T>
+		concept HasPrimOps = requires(T &t)
```
davidchisnall, line 27: Please don't use abbreviations in names (as per the style guide).  Does Prim stand for primitive?  Prime?  Per the comment, I presume primitive, though it's not really clear what 'primitive' means here.  Can it have a more descriptive name (e.g. `CellOperations`)?

sdk/include/ds/linked_list.h
```diff
@@ -0,0 +1,642 @@
...
+	 */
+	template<cell::HasPrimOps T>
+	__always_inline std::enable_if_t<!cell::HasIsSingletonCheck<T>, bool>
+	                is_singleton_check(T *e)
+	{
+		return (e == e->cell_next()) && (e == e->cell_prev());
+	}
+
+	template<cell::HasPrimOps T>
+	__always_inline std::enable_if_t<cell::HasIsSingletonCheck<T>, bool>
```
davidchisnall, line 142: Can this not use a requires clause instead of `enable_if`?

sdk/include/ds/linked_list.h
```diff
@@ -0,0 +1,642 @@
...
+	{
+		auto prev         = curr->cell_prev();
+		elem->cell_next() = curr;
+		elem->cell_prev() = prev;
+		prev->cell_next() = elem;
+		prev              = elem;
+	}
+
+	/**
+	 * Fuse cell initialization with insertion after.  Specifically,
```
davidchisnall, line 265: Is this the same as `emplace` in standard-library terminology?

sdk/include/ds/linked_list.h
```diff
@@ -0,0 +1,642 @@
...
+		}
+	};
+
+	namespace cell
+	{
+
+		/** Cons cell using two pointers */
+		class Pointer
+		{
+			Pointer *prev, *next;
```
davidchisnall, line 503:
```suggestion
			Pointer *prev = this;
                        Pointer *next = this;
```

sdk/include/ds/linked_list.h
```diff
@@ -0,0 +1,642 @@
...
+
+			public:
+			__always_inline void cell_init()
```
davidchisnall, line 506: Why is this not a constructor?

sdk/include/ds/linked_list.h
```diff
@@ -0,0 +1,642 @@
...
+			{
+				return prev == CHERI::Capability{this}.address();
+			}
+
+			__always_inline bool cell_is_doubleton()
+			{
+				return prev == next;
+			}
+
+			__always_inline bool cell_next_is_eq(PtrAddr *p)
```
davidchisnall, line 565: It feels like this would be cleaner if you exposed a `next()` that had an `operator==` (or, since it's the future now, an `operator<=>`) implemented.

sdk/include/ds/pointer.h
```diff
@@ -0,0 +1,256 @@
...
+{
+
+	/**
+	 * Offset a pointer by an unsigned number of bytes.  The return type must be
+	 * explicitly specified by the caller.
+	 */
+	template<typename T, typename U>
+	static inline __always_inline T *offset(U *base, size_t diff)
+	{
+		return reinterpret_cast<T *>(reinterpret_cast<uintptr_t>(base) + diff);
```
davidchisnall, line 23:
```suggestion
		CHERI::Capability c{base};
		c.address() += diff;
		return c;
```

sdk/include/ds/pointer.h
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	/**
+	 * Offset a pointer by a signed number of bytes.  The return type must be
+	 * explicitly specified by the caller.
+	 */
+	template<typename T, typename U>
+	static inline __always_inline T *offset_signed(U *base, ptrdiff_t diff)
```
davidchisnall, line 31: Is this not a different overload for the second parameter?  Does it need a different name?  Are there cases where it's ambiguous whether the `size_t` or `ptrdiff_t` version is intended at the call site (if so, please add a comment)?

sdk/include/ds/xoroshiro.h
```diff
@@ -0,0 +1,75 @@
...
+#pragma once
+
+#include <cstdint>
+#include <cstdlib>
+
+namespace ds::xoroshiro
+{
+	namespace detail
+	{
+		template<typename STATE, typename RESULT, STATE A, STATE B, STATE C>
```
davidchisnall, line 15: clang-tidy should tell you not to use these names, please can you check why not and fix it?  UPPERCASE should be reserved for MACROs (which may have side effects).

sdk/include/ds/xoroshiro.h
```diff
@@ -0,0 +1,75 @@
...
+				y = rotl(y, C);
+				// If both zero, then this does not work
+				if (x == 0 && y == 0)
+					abort();
+				return r >> (StateBits - ResultBits);
+			}
+		};
+	} // namespace detail
+
+	using P128R64 = detail::XorOshiro<uint64_t, uint64_t, 55, 14, 36>;
```
davidchisnall, line 68: What are all of these?  Please give them doc comments.

nwf-msr: I'll ask @mjp41 where he got his constants. :)

sdk/include/ds/xoroshiro.h
nwf-msr: I think it doesn't object to UPPERCASE as a degenerate form of CamelCase?  Even if I set
```
    - key: readability-identifier-naming.TemplateParameterCase
      value: CamelCase
```
`clang-tidy` doesn't object to all caps.

sdk/include/ds/pointer.h
nwf-msr: I was mostly cribbing from `snmalloc`, but I've switched to a single `offset()` that's templated on the type of the offset.

sdk/include/ds/pointer.h
```diff
@@ -0,0 +1,256 @@
...
+			using Type = T;
+
+			__always_inline OffsetPtrAddr(void *c, ptraddr_t &r)
+			  : ctx(c), ref(r)
+			{
+			}
+
+			__always_inline operator T *()
+			{
+				return reinterpret_cast<T *>(
```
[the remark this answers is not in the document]
nwf-msr: This was perhaps a little too golfed.  The issue, tho', is that `PropertyProxyBase` has a `T *` projector (`ptr()`) but no `Capability` projector.  I've rewritten the code on more lines but also, hopefully, with more clarity.

sdk/include/ds/linked_list.h
nwf-msr: You're absolutely right, and these methods predate the refactoring of cons cells to use pointer proxies (originally the cells themselves were also the locus of proxying), so I have moved the `operator==` and `operator<=>` down to the pointer proxies and eliminated these.

sdk/include/ds/linked_list.h
nwf-msr: Because we might want to reset an existing object and, in principle, that object might be bit-packing other state into its representation, so we can't just placement-new a new cell over the old one.  I've added a comment.

sdk/include/ds/linked_list.h
nwf-msr: Yes.  I've renamed things appropriately.

sdk/include/ds/linked_list.h
nwf-msr: Done.

sdk/include/ds/linked_list.h
nwf-msr: Still learning C++.  `enable_if`-s removed.

sdk/include/ds/linked_list.h
nwf-msr: I'm open to being persuaded that cons cells should have constructors (hah), but...
- if we're going to here, we probably should for all three types defined in this file as well as `MChunkHeader` over in `alloc.h`, which we treat as a cons cell with strange pointer proxies.
- we tend not to construct our cells (even using placement new) so much as coerce other pointers into cell types (usually implicitly by coercing a pointer into a cell-containing type like `MChunk`; those, too, do not get constructed in general) which we then `emplace` into their linked lists, so I'm not sure what this buys us.

nwf-msr:
OK, I've folded in most of David's feedback.  For ease of incremental review, I suppose, all the changes are in `fixup!` commits that should be `--autosquash`ed before merging (how I wish TransArc's CM-Lite had escaped).

sdk/include/ds/ring_buffer.h
```diff
@@ -0,0 +1,171 @@
...
+			{
+				/*
+				 * For non-power-of-two, we can use unsigned underflow and
+				 * signed arithmetic right shift to avoid branching.  mask is
+				 * all zeros if next is non-negative and all ones otherwise.
+				 */
+				Ix next   = v - 1;
+				using SIx = std::make_signed_t<Ix>;
+				Ix mask   = static_cast<SIx>(next) >> (8 * sizeof(SIx) - 1);
+				return (mask & (Capacity - 1)) | (~mask & next);
```
mjp41, line 71: It's is unclear to me if this is fast than
```C++
return (v == 0 ? Capacity : v) - 1
```
If there a conditional move on this platform?

sdk/include/ds/ring_buffer.h
```diff
@@ -0,0 +1,171 @@
...
+			return true;
+		}
+
+		/**
+		 * Try to retrieve the index beyond the tail, if there is room.
+		 * Returns true and sets the index if so, or returns false if
+		 * not.
+		 */
+
+		bool tail_next(Ix &v)
```
mjp41, line 145: Not sure I like this name.  It reads like an action or getter, not a predicate.

sdk/include/ds/ring_buffer.h
nwf-msr: I think we do not have a conditional move, no, but this might still be slower than the alternative.

sdk/include/ds/ring_buffer.h
nwf-msr: `head_get` and `tail_next` are conditional-getters: they write to their (out-)parameter if there's a free slot in the buffer and indicate success or failure in their return.  I'm happy to use better names, though!

sdk/include/ds/linked_list.h
davidchisnall: This is skirting dangerously close to UB and probably crossing the line.  If you are changing the type of memory, you must do a placement `new`.  I think the compiler may do some exciting things if you just coerce (in particular, assuming a pointer doesn't alias itself).

sdk/include/ds/pointer.h
```diff
@@ -0,0 +1,268 @@
...
+
+		/**
+		 * Pointer references are pointer proxies, shockingly enough.
+		 */
+		template<typename T>
+		class Pointer
+		{
+			T *&ref;
+
+			__always_inline void set(T *p)
```
davidchisnall, line 93: This looks redundant.  In the proxies in cheri.hh we use a `set` method like this to simplify subclasses, but this is private (not protected) so looks like it's an unnecessary layer of indirection.

sdk/include/ds/pointer.h
```diff
@@ -0,0 +1,268 @@
...
+{
+
+	/**
+	 * Offset a pointer by a number of bytes.  The return type must be
+	 * explicitly specified by the caller.  The type of the displacement offset
+	 * (Offset) is templated so that we can accept both signed and unsigned
+	 * offsets.
+	 */
+	template<typename T, typename Offset = size_t, typename U>
+	static inline __always_inline T *offset(U *base, Offset offset)
```
davidchisnall, line 23:
```suggestion
	template<typename T, typename U>
	static inline __always_inline T *offset(U *base, std::integral auto offset)
```

Not sure if we have the `std::integral` concept, if not then it can be dropped.

sdk/include/ds/xoroshiro.h
davidchisnall: Hmm, it does for other things.  Weird.

sdk/include/ds/linked_list.h
davidchisnall: It makes me uncomfortable that we can have objects that are not in a valid state (especially coming from a Haskell programmer!), rather than requiring that they are constructed in a valid state.  If we can use placement new and ensure that we cannot construct a pointer to a list element without having first placed it in a well-defined state, that would be somewhat less fragile.

If you are concerned about the bit packing, perhaps you could add factory methods that the caller can use that wrap placement new (e.g. `construct_in_place(void*ptr)`)?

sdk/core/allocator/alloc.h
```diff
@@ -130,53 +127,328 @@ static inline size_t leftshifted_val_msb(size_t val)
...
+ *       - Collected in a smallbin ring (using MChunk::ring) or in a treebin
+ *       ring (using either/both the TChunk and MChunk::ring links).
+ */
+struct __packed __aligned(MallocAlignment)
+MChunkHeader
+{
+	/**
+	 * Compressed size of the predecessor chunk.  See cell_prev().
+	 */
+	SmallSize sprev;
```
JerryHsia-MS, line 285: Very nit: `sprev` and `shead` used to contain both size and other metadata bits. Now they are just sizes, we can give them better names like `prevSize` and `currSize` if it doesn't require too many changes.

sdk/core/allocator/alloc.h
```diff
@@ -130,53 +127,328 @@ static inline size_t leftshifted_val_msb(size_t val)
...
+	 *
+	 * This function is marked always-inlined as it is called exactly once.
+	 */
+	__always_inline static MChunkHeader *make(void *base, size_t size)
+	{
+		size -= sizeof(MChunkHeader);
+
+		auto first = static_cast<MChunkHeader *>(base);
+		first->clear();
+		first->shead       = size2head(size);
```
JerryHsia-MS, line 397: Does this overflow into the header word right after `base + size`, instead of placing the fake footer header at the end of `base + size`?

sdk/core/allocator/alloc.h
```diff
@@ -423,53 +680,63 @@ class __packed __aligned(MALLOC_ALIGNMENT) MChunk
...
+	void metadata_clear()
 	{
-		epoch_enq = revoker.system_epoch_get();
+		static_assert(sizeof(*this) == sizeof(MChunkHeader) + sizeof(ring));
+		/*
+		 * This is spelled slightly oddly as using memset results in a call
+		 * rather than a single store instruction.
+		 */
+		static_assert(sizeof(ring) == sizeof(uintptr_t));
+		*reinterpret_cast<uintptr_t *>(&this->ring) = 0;
```
JerryHsia-MS, line 695: Does `capaligned_zero()` help? I wonder if the compiler is clever enough to collapse it into a single cap store.

sdk/core/allocator/alloc.h
```diff
@@ -489,24 +757,127 @@ class __packed __aligned(MALLOC_ALIGNMENT) TChunk : public MChunk
...
+	void mark_root()
+	{
+		parent = reinterpret_cast<TChunk *>(RootParent);
+	}
+
+	bool is_tree_ring()
+	{
+		return parent == reinterpret_cast<TChunk *>(RingParent);
+	}
+	void mark_tree_ring()
```
JerryHsia-MS, line 792: Nit: could benefit from more comments here (although the comments before this PR were already lacking). I think this is called for tree chunks that are not the direct child of a parent?

sdk/core/allocator/alloc.h
```diff
@@ -1618,57 +2009,103 @@ class MState
...
+		{
+			// Consolidate backward
+			MChunk *prev = p->chunk_prev();
+			unlink_chunk(prev, prev->size_get());
+			ds::linked_list::unsafe_remove_link(&prev->header, &p->header);
+			p->header.clear();
+			p = prev;
+		}
+
+		MChunk *next = p->chunk_next();
```
JerryHsia-MS, line 2035: At this point, if `prev` is not in use, have we updated the size field to include both `p` and `prev` so that `p->chunk_next()` can work?

If not, I think we should move this line above `is_prev_in_use()` so that `chunk_next()` is valid.

sdk/core/allocator/alloc.h
```diff
@@ -1618,57 +2009,103 @@ class MState
...
 		}
-	erroraction:
-		corruption_error_action();
-	postaction:
-		heapFreeSize += psizeold;
+
+		auto qring = quarantine_pending_get(oldestPendingIx);
+
+		/*
+		 * It can happen that this ring is empty, because everything that was
```
JerryHsia-MS, line 2079: Is this still true?

nwf-msr: No.  I've updated the comment.

nwf-msr:
At long last, here is a version that, I hope, avoids UB around unconstructed objects and silently re-typed memory.

nwf-msr:
clang-tidy in CI does not like this PR for reasons beyond my understanding (it claims to have been invoked with no arguments) if I tell it to include alloc.h and alloc_config.h (or, rather, to stop excluding those).  I've dropped the commit that does so, which is a little sad.

sdk/core/allocator/alloc.h
nwf-msr: No, that also generates a call, AFAICT.

sdk/core/allocator/alloc.h
Jerryxia32: You are right. Somehow I didn't see the subtraction.

sdk/core/allocator/alloc.h
Jerryxia32: Yep

JerryHsia-MS (approved):

PR #8, opened 2023-02-13
Add a test for a bug in the switcher, alongside a fix

saaramar:
The switcher does not check the validity of the target entry point. Add a check for that and implement a test.

### DIFF
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -141,7 +141,7 @@ compartment_switcher_entry:
 	cunseal			ct1, ct1, cs0
 	// Make sure the export table is valid
 	cgettag			s0, ct1
-	beqz			s0, .Lforce_unwind_with_return_values
+	beqz			s0, .Linvalid_entry
 	// Load the entry point offset.
 	clhu			s0, ExportEntry_offset_functionStart(ct1)
 	// At this point, we known that the cunseal has succeeded (we didn't trap
@@ -315,9 +315,16 @@ exception_entry_asm:
 	reloadRegisters c1, cgp, c4, c5, c6, c7, c8, c9, c10, c11, c12, c13, c14, c15, csp
 	mret
 
-.Lforce_unwind_with_return_values:
-	li		a0, -1
-	li		a1, 0
+.Linvalid_entry:
+// Mark this threads as in the middle of a forced unwind.
+	li             a0, 1
+	csb            a0, TrustedStack_offset_inForcedUnwind(ctp)
+// Make sure we don't leak anything to the compartment.
+// Registers might been used by the call and therefore need zeroing.
+	zeroAllRegistersExcept a0, s0, s1, sp, a2, gp
+// Store an error value in return registers, which will be passed to the
+// caller on unwind. a0 is zeroed by zeroAllRegistersExcept 
+	li          a0, -1
 // We are starting a forced unwind.  This is reached either when we are unable
 // to run an error handler, or when we do run an error handler and it instructs
 // us to return.  This treats all register values as undefined on entry.
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -137,8 +137,12 @@ compartment_switcher_entry:
 	LoadCapPCC		cs0, compartment_switcher_sealing_key
 	li				gp, 9
 	csetaddr		cs0, cs0, gp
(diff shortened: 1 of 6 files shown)


sdk/core/switcher/entry.S
```diff
@@ -311,6 +315,9 @@ exception_entry_asm:
 	reloadRegisters c1, cgp, c4, c5, c6, c7, c8, c9, c10, c11, c12, c13, c14, c15, csp
 	mret
 
+.Lforce_unwind_with_return_values:
```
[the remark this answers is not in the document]
saaramar: Thanks! Done in this commit https://github.com/microsoft/cheriot-rtos/pull/8/commits/6d7ea17057dc4677ebd77c188f069514acd3c008

sdk/core/switcher/entry.S
```diff
@@ -311,6 +315,16 @@ exception_entry_asm:
 	reloadRegisters c1, cgp, c4, c5, c6, c7, c8, c9, c10, c11, c12, c13, c14, c15, csp
 	mret
 
+.Linvalid_entry:
+// Mark this threads as in the middle of a forced unwind.
```
davidchisnall, line 319: Can you expand this to explain *why* (we want to resume rather than unwind if the compartment doesn't have an error handler)?

tests/compartment_calls-test.cc
```diff
@@ -40,3 +40,9 @@ void test_compartment_call()
 	  compartment_call_inner(value, value, &value, value, &value, value, value);
 	TEST(ret == 0, "compartment_call_inner returend {}", ret);
 }
+
+void test_compartment_call()
+{
+	test_number_of_arguments();
+	test_incorrect_export_table(NULL);
```
davidchisnall, line 47:
```suggestion
	test_incorrect_export_table(nullptr);
```

tests/compartment_calls_inner.cc
```diff
@@ -10,6 +10,12 @@
 
 using namespace CHERI;
 
+extern "C" ErrorRecoveryBehaviour
+compartment_error_handler(ErrorState *frame, size_t mcause, size_t mtval)
+{
+	return ErrorRecoveryBehaviour::ForceUnwind;
+}
+
```
davidchisnall, line 18: This should be the default, do we need it to exist?

davidchisnall: I see, we don't unwind here because it's treated as a failure in the callee without the handler.  Please add a comment explaining this.

saaramar: Exactly. Great idea, I'm adding a comment and another inner compartment to tests both cases.

davidchisnall: It would also be good to update the error handling doc for some of these corner cases.  For example, how are each of the following handled:

 - Stack exhaustion
 - Trusted stack exhaustion
 - Invalid return addresses
 - Invalid callee addresses

We don't yet correctly handle the stack exhaustion case, but I'm finishing up the compiler change to generate the correct metadata.

saaramar: True, these corner cases are important and we have a good picture of how they should look like. I'll do that in a separate PR.

tests/compartment_calls_inner.cc
```diff
@@ -95,4 +101,17 @@ int compartment_call_inner(int        x0,
...
+{
+	/*
+	 * Trigger a cross-compartment call with an invalid export entry.
+	 */
+
+	debug_log("test an incorrect export table entry");
+
+	fn();
+
+	TEST(false, "Should be unreachable");
```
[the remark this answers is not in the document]
saaramar: Yep, already done; going to commit this in a few minutes :)

sdk/core/switcher/entry.S
```diff
@@ -336,10 +349,10 @@ exception_entry_asm:
 	LoadCapPCC     cs0, compartment_switcher_sealing_key
 	// ca2 at this point was loaded by .Lpop_trusted_stack_frame from the pcc
 	// in the trusted stack and so should always be sealed as a sentry type.
-	cgettype       gp, cra
+	cgettype       gp, ca2
```
davidchisnall: This looks like it's accidentally reverting a fix?

davidchisnall (approved):
Looks good, thanks!

PR #10, opened 2023-02-14
Use cra instead of ca2 in the switcher.

davidchisnall:
ca2 will contain the wrong value on out-of-stack paths.


JerryHsia-MS (approved):

sdk/core/switcher/entry.S
```diff
@@ -333,10 +336,10 @@ exception_entry_asm:
 	LoadCapPCC     cs0, compartment_switcher_sealing_key
 	// ca2 at this point was loaded by .Lpop_trusted_stack_frame from the pcc
```
rmn30, line 337: Looks like this comment is now out of date. Should it say `cra` instead?

PR #12, opened 2023-02-15
Enable interrupts in the compartment switcher.

davidchisnall:
Preemption during a compartment switch shouldn't matter: the compartment switching code touches only thread-local values (registers, the trusted stack, and the stack).  It currently runs for a potentially long time with interrupts disabled while it zeroes the stack.  I was going to make it just enable interrupts for the zeroing loop, but I'm increasingly confident that it is safe to enable them for the whole switcher path.

### DIFF
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -97,6 +97,23 @@ switcher_scheduler_entry_csp:
 	bnez               t2, .Lforce_unwind
 .endm
 
+/**
+ * This macro sets the interrupt status.  If reg is non-zero, interrupts are
+ * disabled, if it is zero then they are enabled.
+ */
+.macro set_interrupt_state reg
+	// Set the register to 1 on any non-zero value
+	snez               \reg, \reg
+	// Set the register to 8 if reg is 1, 0 if it is 0
+	slli               \reg, \reg, 0x3
+	// Clear the interrupt-enable bit if reg started as 1
+	csrc               mstatus, \reg
+	// If reg is 8, set to 0, if 0, set to 8
+	xori               \reg, \reg, 0x8
+	// Set the interrupt-enable bit if reg started as 0
+	csrs               mstatus, \reg
+.endm
+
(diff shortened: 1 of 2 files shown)


saaramar:
Great point. I agree - while this sounds concerning at first, there isn't any shared state/context to manipulate between context switches. Everything state-related is on the trusted stack, and we have only one single thread with one single trusted stack.

Just to be sure - the places we manipulate the enable-interrupt-bit are during `force_unwind` and `out`. In the latter you changed the logic accordingly, and in the former we clear the bit unconditionally and set it accordingly (so the switcher state is not a factor of the decision).

Based on that, I think this change is safe.

sdk/core/switcher/entry.S
```diff
@@ -180,10 +180,10 @@ compartment_switcher_entry:
 	cjr				cs0
 .Lzero_arguments_start:
 	zeroRegisters	a0, a1, a2, a3, a4, a5, t0
-	// Enable interrupts of the interrupt-disable bit is not set in flags
-	bnez			t1, .Lskip_interrupt_disable
-	csrsi			mstatus, 0x8
-.Lskip_interrupt_disable:
+	// Disable interrupts if the interrupt-disable bit is  set in flags
```
saaramar, line 183: There is an extra space there

JerryHsia-MS (approved):

davidchisnall:
[ Leaving this open for a week or so to give folks a chance to carefully think through any potential security implications ]

sdk/core/switcher/entry.S
rmn30: It's worth pointing out that doing a cross-compartment call from a context with interrupts disabled will now enable interrupts, at least while in the switcher if not at the destination. This is probably fine as I don't think code running with interrupts disabled should make cross-compartment calls, but it is a potential gotcha.

davidchisnall:
@nwf points out that we now have interrupts enabled in places where mscratch does not have a valid value, so I need to rethink this somewhat.

PR #15, opened 2023-02-20
Tweaks to clang-tidy CI

nwf-msr:


rmn30 (approved):
Looks good to me. I can't really comment on the clang-tidy config changes or why `alloc.h` and `alloc_config.h` were excluded previously but would like to see this merged to fix the bug in the clang-tidy invocation.

davidchisnall (approved):

PR #16, opened 2023-02-20
Add support for static sealing types and sealed objects.

davidchisnall:
This adds macros that allow a compartment to request a software-defined sealing capability that is allocated at load time.  These can then be used with another macro that allows compartment A to statically allocate objects sealed with a sealing type owned by compartment B, so that B can request that the allocator unseal them.

This is a building block for things like per-compartment heap quotas, capabilities that authorise network connections to specific hosts, and so on.


sdk/core/loader/boot.cc
```diff
@@ -123,11 +123,23 @@ namespace
...
+		FirstStaticSoftware = 16,
+
+		/**
+		 * The first sealing key in the space that the allocator will
+		 * dynamically allocate for sealing types.
+		 */
+		FirstDynamicSoftware = 0x1000000,
 	};
 
 	// We currently have a 4-bit otype, but we'd like to reduce it to 3.
```
nwf-msr, line 139: While you're in the neighborhood, that comment is stale.

sdk/core/loader/boot.cc
```diff
@@ -245,6 +257,50 @@ namespace
...
+	/**
+	 * Returns a sealing capability to use for statically allocated sealing
+	 * keys.
+	 */
+	uint16_t allocate_static_sealing_key()
+	{
+		static uint16_t nextValue = FirstStaticSoftware;
+		// We currently stash the allocated key value in the export table.  We
+		// could expand this a bit if we were a bit more clever in how we used
+		// that space, but 1^16 static sealing keys will require over 768 KiB
```
nwf-msr, line 269:
```suggestion
		// that space, but 2^16 static sealing keys will require over 768 KiB
```

sdk/core/loader/boot.cc
```diff
@@ -435,6 +502,51 @@ namespace
...
+				  build<uint32_t,
+				        Root::Type::RWGlobal,
+				        PermissionSet{Permission::Load, Permission::Store}>(
+				    target);
+				// TODO: This currently places a restriction that data memory
+				// can't be in the low 64 KiB of the address space.  That may be
+				// too restrictive. If we haven't visited this sealed object
+				// yet, then we should update its first word to point to the
+				// sealing type.
+				if (*sealingType >= 0x10000)
```
nwf-msr, line 522: What is `0x10000`?  I was sort of expecting to see something about `FirstDynamicSoftware` here or somesuch.  (Values larger than that are certainly invalid, so that could be an offset applied to the address of the export table?)

If a dynamic approach is necessary, is there no other metadata in the export table entry that could be used to discriminate processed from unprocessed rows?

sdk/core/loader/boot.cc
```diff
@@ -435,6 +502,51 @@ namespace
...
+			if (contains(sourceCompartment.sealedObjects, target, size))
+			{
+				auto sealingType =
+				  build<uint32_t,
+				        Root::Type::RWGlobal,
+				        PermissionSet{Permission::Load, Permission::Store}>(
+				    target);
+				// TODO: This currently places a restriction that data memory
+				// can't be in the low 64 KiB of the address space.  That may be
+				// too restrictive. If we haven't visited this sealed object
```
nwf-msr, line 519: If we haven't visited yet?  That seems a little worrying as an approach.  From our discussion, I thought we could get away with two passes without the need to re-visit anything?

sdk/core/loader/boot.cc
```diff
@@ -874,10 +1060,8 @@ extern "C" SchedulerEntryInfo loader_entry_point(const ImgHdr &imgHdr,
 	switcherKey.bounds()  = 1;
 	setSealingKey(imgHdr.scheduler(), Scheduler);
 	setSealingKey(imgHdr.allocator(), Allocator);
-	setSealingKey(imgHdr.allocator(),
-	              static_cast<SealingType>(0x1000000),
-	              0xff000000,
-	              sizeof(void *));
+	setSealingKey(
+	  imgHdr.allocator(), FirstDynamicSoftware, 0xff000000, sizeof(void *));
```
nwf-msr, line 1064: `0xff000000` is `2**32 - FirstDynamicSoftware`?

sdk/core/loader/types.h
```diff
@@ -862,9 +868,22 @@ namespace loader
 		static constexpr uint8_t InterruptStatusMask = uint8_t(0b11)
 		                                               << InterruptStatusShift;
 
+		/**
+		 * The flag indicating that this is a fake entry used to identify
+		 * sealing types.  Nothing should refer to this other than an import
+		 * table entry from the same compartment, which will be populated with
+		 * a sealing capability.
+		 */
```
nwf-msr, line 876: Is that right?  Don't the statically sealed objects also refer to the export table entry to specify what type they're sealed under?

sdk/include/compartment.h
```diff
@@ -49,6 +49,112 @@
...
+		      ".ifndef __export." name "\n"                                    \
+		      "  .type     __export." name ",@object\n"                        \
+		      "  .section  .compartment_exports." name                         \
+		      ",\"awG\",@progbits," name ",comdat\n"                           \
+		      "  .globl    __export." name "\n"                                \
+		      "  .p2align  2\n"                                                \
+		      "__export." name ":\n"                                           \
+		      "  .half 0\n"                                                    \
+		      "  .byte 0\n"                                                    \
+		      "  .byte 0b100000\n"                                             \
```
nwf-msr, line 80: Assuming this is `SealingTypeEntry`, could you bring it into the assembler via an `i`-type input rather than as a constant?  Failing that, a comment for `grep` to find?

tests/static_sealing_inner.cc
```diff
@@ -0,0 +1,51 @@
...
+	// Make sure the sealing key has sensible permissions
+	TEST((check_pointer<PermissionSet{Permission::Seal,
+	                                  Permission::Unseal,
+	                                  Permission::Global,
+	                                  Permission::User0}>(key, 1)),
+	     "Incorrect permissions on {}",
+	     key);
+	// Make sure it's in the right range.
+	TEST(
+	  keyCap.address() >= 16,
```
nwf-msr, line 26:
```suggestion
	  keyCap.address() >= FirstStaticSoftware,
```

tests/static_sealing_inner.cc
```diff
@@ -0,0 +1,51 @@
...
+	  "Software sealing key has an address in the hardware-reserved range: {}",
+	  keyCap.address());
+	TEST(keyCap.address() < 0x10000,
```
nwf-msr, line 29:
```suggestion
	TEST(keyCap.address() < FirstDynamicSoftware,
```

tests/static_sealing_inner.cc
```diff
@@ -0,0 +1,51 @@
...
+	     "Software sealing key has an address too large: {}",
+	     keyCap.address());
+	// Make sure that it's a single sealing type
+	TEST(keyCap.bounds() == 1, "Invalid bounds on {}", key);
+
+	// Try to use it
+	Capability unsealed = token_unseal(key, obj);
+	debug_log("Unsealed object: {}", unsealed);
+	// Make sure that the unsealed allocation is the right everything.
+	TEST(unsealed->value == 42, "Unexpected value for static sealed object");
```
nwf-msr, line 39:
```suggestion
	TEST(unsealed->value == 42, "Unexpected value in static sealed object");
```

tests/static_sealing_inner.cc
```diff
@@ -0,0 +1,51 @@
...
+	SKey       key = STATIC_SEALING_TYPE(SealingType);
+	Capability keyCap{key};
+
+	debug_log("Static sealing key: {}", key);
+	// Make sure the sealing key has sensible permissions
+	TEST((check_pointer<PermissionSet{Permission::Seal,
+	                                  Permission::Unseal,
+	                                  Permission::Global,
+	                                  Permission::User0}>(key, 1)),
+	     "Incorrect permissions on {}",
```
nwf-msr, line 22:
```suggestion
	     "Incorrect permissions on static sealing key {}",
```

tests/static_sealing_inner.cc
```diff
@@ -0,0 +1,51 @@
...
+	TEST(unsealed.length() == sizeof(TestType),
+	     "Incorrect length on unsealed capability {}",
+	     unsealed);
+	TEST((check_pointer<PermissionSet{Permission::Load,
+	                                  Permission::Store,
+	                                  Permission::LoadStoreCapability,
+	                                  Permission::LoadMutable,
+	                                  Permission::LoadGlobal,
+	                                  Permission::Global}>(unsealed.get(), 1)),
+	     "Incorrect permissions on {}",
```
nwf-msr, line 49:
```suggestion
	     "Incorrect permissions on unsealed statically sealed object {}",
```

nwf-msr (changes requested):
Nit-picking, but nice functionality in not a lot of code. :)

sdk/core/loader/boot.cc
davidchisnall: Two import entries may point to the same cap (they probably shouldn't but I didn't want to preclude it).

sdk/core/loader/boot.cc
davidchisnall: Yup

sdk/core/loader/boot.cc
davidchisnall: It complicates the software to use values < 2^16, and feels unnecessary given that we'd be using so much RAM just to hold the types to hit that code path - we can't easily test something that requires such a huge firmware image.

sdk/core/loader/boot.cc
nwf-msr: Can it be spelled that way rather than as a "mysterious" hex constant?

sdk/include/compartment.h
```diff
@@ -49,6 +49,113 @@
...
+		  "  .type     __export." name ",@object\n"                            \
+		  "  .section  .compartment_exports." name ",\"awG\",@progbits," name  \
+		  ",comdat\n"                                                          \
+		  "  .globl    __export." name "\n"                                    \
+		  "  .p2align  2\n"                                                    \
+		  "__export." name ":\n"                                               \
+		  "  .half 0\n" /* function start and stack size initialised to 0 */   \
+		  "  .byte 0\n"                                                        \
+		  "  .byte 0b100000\n" /* Set the flag that indicates that this is a   \
+		                          sealing key. */                              \
```
nwf-msr, line 82: GitHub has seemingly eaten my earlier suggestion that this be brought in by name via `"i"(SealingTypeEntry)`

nwf-msr (approved):

PR #17, opened 2023-02-21
Headers of heap chunks must point to set shadow bits

JerryHsia-MS:
We used to set the shadow bit on malloc() and clear it on free() to guard against double free, but this is done by the load barrier anyway. The second free will have the tag cleared which will fail, so we don't have to do that.

Instead, to guard against the case of raising the base of an allocation to the top, which points to the shadow bit of the next header, which can either survive revocation or be used to trick the allocator into thinking it is a valid parameter for free(), we maintain the invariant that malloc headers must always point to shadow bits of 1, avoiding both problems.

### DIFF
--- a/sdk/core/allocator/alloc.h
+++ b/sdk/core/allocator/alloc.h
@@ -387,6 +387,11 @@ MChunkHeader
 	 * system.  In particular, if this is a free chunk, then the result will be
 	 * two free chunks in a row; the caller is expected to fix this by marking
 	 * at least one of the two split chunks as in-use.
+	 *
+	 * Note that we keep the invariant that all headers correspond to shadow
+	 * bits that are set. For this to be true, new headers are assumed to be
+	 * created only by split() after initialisation which is in charge of
+	 * setting the bits.
 	 */
 	MChunkHeader *split(size_t offset)
 	{
@@ -394,6 +399,8 @@ MChunkHeader
 
 		auto newnext = new (newloc) MChunkHeader();
 		newnext->clear();
+		// Invariant that headers must point to shadow bits that are set.
+		revoker.shadow_paint_single(CHERI::Capability{newloc}.address(), true);
 
 		ds::linked_list::emplace_after(this, newnext);
 		newnext->isCurrInUse = newnext->isPrevInUse = isCurrInUse;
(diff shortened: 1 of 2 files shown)


sdk/core/allocator/alloc.h
```diff
@@ -1966,6 +1979,8 @@ class MState
 			unlink_chunk(MChunk::from_header(prev), prev->size_get());
 			ds::linked_list::unsafe_remove_link(prev, p);
 			p->clear();
+			// p is no longer a header. Clear the shadow bit.
+			revoker.shadow_paint_single(CHERI::Capability{p}.address(), false);
```
nwf-msr, line 1983: Maybe move the `shadow_paint_single` into `->clear`?

nwf-msr (changes requested):
LGTM modulo a nit or two.

Maybe add a test for all the shadow bits being clear in the `Debug::Assert(capaligned_range_do(...))` in `mspace_malloc_success`?

sdk/core/allocator/alloc.h
JerryHsia-MS: Slightly tricky because `clear()` is called on both header creation and destruction. We need to have two wrappers, or pass a boolean to `clear()`?

nwf-msr: Ah, right, because we might be `clear()`-ing some existing data under where we're `split()`-ing to make a new header.  I'd be tempted to have a private `zero()` called on construction and by the public `clear()`, but it's probably fine as is.

sdk/core/allocator/alloc.h
```diff
@@ -2155,15 +2173,22 @@ class MState
...
+		 * inside a Capability because that gives exact equal for nullptr.
+		 */
+		Debug::Assert(
+		  capaligned_range_do(p->body(),
+		                      size - sizeof(MChunkHeader),
+		                      [](void **word) {
+			                      CHERI::Capability eachCap{*word};
+			                      return eachCap != nullptr &&
+			                             revoker.shadow_bit_get(
+			                               CHERI::Capability{word}.address());
```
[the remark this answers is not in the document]
nwf-msr: Thanks; sorry to miss the commit message.

nwf-msr (approved):

PR #18, opened 2023-02-21
Implement precise checking of free(mem)

JerryHsia-MS:
We used to only check that cgetbase(mem) is correct. We harden free so that mem needs to be exactly equal to what the user got on malloc(). If there are missing permissions, different bounds and addresses, it is likely a bug from the user.

To implement this, a couple of things need to change:
1. The setbounds in malloc() needs at least capability alignment.
2. On malloc(size), if the returned chunk can be precisely covered by a capability, then always use the whole chunk.
3. If not, use the portion that can be precise, which shares the base with the whole chunk and is exactly MChunkAlignment smaller than the whole thing. If it is >= 2*MChunkAlignment smaller, then we must have split it into two chunks so this cannot happen.

### DIFF
--- a/sdk/core/allocator/alloc.h
+++ b/sdk/core/allocator/alloc.h
@@ -12,6 +12,7 @@
 #include <ds/linked_list.h>
 #include <ds/pointer.h>
 #include <ds/ring_buffer.h>
+#include <errno.h>
 #include <stddef.h>
 #include <stdint.h>
 #include <stdlib.h>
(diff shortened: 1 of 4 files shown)


sdk/core/allocator/main.cc
```diff
@@ -215,15 +215,9 @@ namespace
...
-	 * Since we use the shadow bits to detect valid frees, we need to consult
-	 * the revoker on whether the user cap is valid.
-	 */
-	if (!revoker.is_free_cap_valid(mem))
-	{
-		return -EINVAL;
-	}
-	gm->mspace_free(mem);
+	int rv = gm->mspace_free(mem);
+	if (rv)
```
davidchisnall, line 219: Clang-tidy should complain here.

sdk/core/allocator/alloc.h
```diff
@@ -1056,21 +1079,48 @@ class MState
 	 * @param mem the user cap which has been checked, but has not been
 	 * rederived into an internal cap yet
 	 */
-	void mspace_free(CHERI::Capability<void> mem)
+	int mspace_free(CHERI::Capability<void> mem)
```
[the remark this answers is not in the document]
JerryHsia-MS: Done

sdk/core/allocator/main.cc
JerryHsia-MS: It seems to be okay with it for some reason, but I've fixed the style.

davidchisnall (approved):

PR #19, opened 2023-02-21
Add tests and checks for invalid CSP on switcher entry

saaramar:


sdk/core/switcher/entry.S
```diff
@@ -315,6 +324,14 @@ exception_entry_asm:
 	reloadRegisters c1, cgp, c4, c5, c6, c7, c8, c9, c10, c11, c12, c13, c14, c15, csp
 	mret
 
+.Linvalid_csp_permissions:
+	cjal 			.Lpop_trusted_stack_frame
+	csh zero, 0(cnull)
```
rmn30, line 329: This will trap. Was it committed by mistake? If not how will the subsequent instructions be reached?

saaramar: hehe yes, thanks!

PR #22, opened 2023-02-24
add stack checks

saaramar:


rmn30:
I note that this doesn't test with a sealed stack pointer. This is harder to arrange because you would need a sealing cap to seal the stack pointer with or you could use a sentry but then it would have incorrect permissions too. I think it would fail in the same way as an untagged stack pointer anyway so providing that works it is probably fine.

sdk/core/switcher/entry.S
```diff
@@ -91,11 +91,21 @@ compartment_switcher_entry:
...
+	clhu			tp, TrustedStack_offset_frameoffset(ct2)
+	cgetlen         t2, ct2
+	bgeu           	tp, t2, .Lout_of_trusted_stack
+	// make sure the caller's CSP is valid
+	cgettag         t2, csp
+	beqz            t2, .Lforce_unwind
+	// make sure the caller's CSP has the expected permissions
+	cgetperm        t2, csp
+	li              tp, 0x7e
+	bne	            tp, t2, .Lforce_unwind
```
davidchisnall, line 105: Can these be moved before we access mscratch?  We don't need access to any privileged state to do these checks.

sdk/core/switcher/entry.S
```diff
@@ -415,6 +425,13 @@ exception_entry_asm:
 	li          a1, 0
 	// If we don't have enough space, give up and force unwind
 	bltu        t1, tp, .Lforce_unwind
+	// make sure the caller's CSP is valid
+	cgettag     t1, ct0
+	beqz        t1, .Lforce_unwind
+	// make sure the caller's CSP has the expected permissions
+	cgetperm    t1, ct0
+	li 	        tp, 0x7e
+	bne	        tp, t1, .Lforce_unwind
```
davidchisnall, line 434: These checks are the same as above, can you make them a macro? We should also add a sealing check.

saaramar: Great point, thanks!

sdk/core/switcher/entry.S
saaramar: True, great point. I moved it to the beginning of the trap

sdk/core/switcher/entry.S
```diff
@@ -79,23 +79,47 @@ switcher_scheduler_entry_csp:
...
 #ifndef NDEBUG
 	// XXX: This line is useless, only for mscratch to show up in debugging.
 	cmove			ct2, ct2
 #endif
-	clhu			tp, TrustedStack_offset_frameoffset(ct2)
+
 	// make sure the trusted stack is still in bounds
-	cgetlen                 t2, ct2
-	bgeu                    tp, t2, .Lout_of_trusted_stack
+	clhu			tp, TrustedStack_offset_frameoffset(ct2)
```
davidchisnall, line 117: Can you fix indenting? It looks like you're mixing tabs and spaces.

sdk/core/switcher/trusted-stack-assembly.h
```diff
@@ -44,3 +44,5 @@ EXPORT_ASSEMBLY_SIZE(TrustedStackFrame, (7 * 8))
 
 #define TSTACKOFFSET_FIRSTFRAME                                                \
 	(TrustedStack_offset_frameoffset + TSTACK_HEADER_SZ)
+
+#define COMPARTMENT_STACK_PERMISSIONS 0x7e
```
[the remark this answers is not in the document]
saaramar: Oh right, thanks! Done

sdk/core/switcher/entry.S
saaramar: good catch, thanks :)

davidchisnall (approved):

PR #27, opened 2023-03-03
Start requiring a capability to allocate memory.

davidchisnall:
APIs that allocate and free memory now require a capability that authorises them to do so.  This includes indirect allocation via another compartment.

Each compartment can have a default malloc capability, which is used implicitly with malloc and free.

The current version is a work in progress with a number of limitations:

 - The allocator capability is not actually used.  The allocator checks that you have a valid allocator capability, it does not check anything else or use this capability.
 - The scheduler permits you to destroy objects unconditionally, even if you don't have the rights to deallocate them. This means that an attacker can delete an object with no mechanism for the real owner to recover the memory.
 - Some of the new parameters are not in sensible / consistent places.
 - Some new things have silly names.
 - The examples have not been updated yet.


saaramar (approved):

sdk/include/stdlib.h
```diff
@@ -4,10 +4,37 @@
...
 #include <cdefs.h>
+#include <compartment-macros.h>
 #include <stddef.h>
 #include <stdint.h>
 #include <timeout.h>
 
+#ifndef MALLOC_DEFAULT_QUOTA
+#	define MALLOC_DEFAULT_QUOTA 4096
+#endif
+struct MallocCapability
```
nwf-msr, line 15: Hm. I think I know why you called this a `Capability`, and I don't completely disagree, but I might prefer `MallocCapabilityState` or `MallocAuthorityState` or something along those lines: it's not a CHERI capability, and (an instance of) *this structure* isn't presented as a capability would be, in general: a (sealed) capability to (a wrapper around) an instance of this structure is the actual bearer token.

sdk/include/stdlib.h
```diff
@@ -4,10 +4,37 @@
...
+{
+	size_t    quota;
+	size_t    unused;
+	uintptr_t reserved[2];
```
nwf-msr, line 19: I might reserve more space; we might want to have these...
- know a small identifier for themselves (another `uint16_t`),
- be intrusively indexed on that identifier by a list or tree or something (2 or 3 `uintptr_t`s, a la `MChunk` or `TChunk`)
- have a hierarchical relationship amongst themselves (perhaps as a rose tree with children encoded as a ds::linked_list ring... another 5 `uintptr_t`s)

I don't think we imagine that there will be too many of these, so making them kind of hefty is probably not the worst fate?

sdk/include/stdlib.h
nwf-msr, line 15: It's a bit of a shame that the layout of this `struct` has to be known like this, but doing anything else and supporting static construction is a rabbit hole I don't think we want to fall into.  Though if you come up with some brilliance to avoid it...

sdk/include/stdlib.h
```diff
@@ -52,15 +85,15 @@ void *__cheri_compartment("alloc") heap_allocate(size_t size, Timeout *timeout);
  * Memory returned from this interface is guaranteed to be zeroed.
  */
 void *__cheri_compartment("alloc")
-  heap_allocate_array(size_t nmemb, size_t size, Timeout *timeout);
+  heap_allocate_array(struct SObjStruct *heapCapability, size_t nmemb, size_t size, Timeout *timeout);
```
nwf-msr, line 88: Do you want to have an `ifdef __cplusplus` overload for this and for `heap_free`, too?

sdk/include/compartment-macros.h
```diff
@@ -0,0 +1,174 @@
...
+		      "  auipcc  %0,"                                                  \
+		      "      %%cheri_compartment_pccrel_hi(__import_mem_" #name ")\n"  \
+		      "  clc     %0, %%cheri_compartment_pccrel_lo(1b)(%0)\n"          \
+		      : "=C"(ret));                                                    \
+		ret;                                                                   \
+	})
+
+/**
+ * Macro to test whether a device with a specific name exists in the board
+ * definition for the current target.
```
nwf-msr, line 37: I know it's not part of this PR, but this comment really should be expanded to include something about this macro being useful *only* within a preprocessor expression.  (It's probably best to limit it to preprocessor evaluation contexts rather than pursue terrors like http://hmijailblog.blogspot.com/2016/03/an-isdefined-c-macro-to-check-whether.html ...)

sdk/core/scheduler/common.h
```diff
@@ -155,4 +162,58 @@ namespace sched
...
+
+			public:
+			Deleter(struct SObjStruct *heapCapability)
+			  : heapCapability(heapCapability)
+			{
+			}
+			void operator()(T *object)
+			{
+				object->~T();
+				free(object);
```
nwf-msr, line 180: Should that be `heap_free(heapCapability, object)`?

nwf-msr (changes requested):

sdk/include/stdlib.h
davidchisnall: I was in two minds about whether to reserve state at all.  Until we do an ABI-stable release, I can add fields at will later.

sdk/include/stdlib.h
davidchisnall: This is going away, it was just there so I could start to get things working without updating all of the callers.

sdk/core/scheduler/common.h
davidchisnall: Yup, that's in the next commit.

nwf-msr (approved):

PR #28, opened 2023-03-08
Fix stack tests

saaramar:


tests/stack_integrity_thread.cc
```diff
@@ -8,6 +8,26 @@ using namespace CHERI;
 
 bool *threadStackTestFailed;
 
+/*
+ * Define a macro that gets a __cheri_callback capability and calls it, while
+ * support adding instruction before the call. This is used to avoid code
+ * duplication, in cases we want to call a __cheri_callback in multiple
+ * places while adding additional functionalities.
+ */
+#define CALL_CHERI_CALLBACK(fn, instruction, additional_input)                 \
```
JerryHsia-MS, line 17: Very Nit: `fn` should probably have a better name like `handle` because it is not really a function.

JerryHsia-MS: It also might be less confusing to say that `additional_input` is referenced in the `instruction`. I got really confused when I saw `additional_input` not referenced anywhere in the macro.

tests/stack_integrity_thread.cc
```diff
@@ -89,8 +108,7 @@ void test_stack_invalid_on_call(bool *outTestFailed,
 {
 	threadStackTestFailed = outTestFailed;
 
-	__asm__ volatile("ccleartag      csp, csp\n"
-	                 "csh            zero, 0(cnull)\n");
+	CALL_CHERI_CALLBACK(fn, "move a0, %1\nccleartag csp, csp\n", 0);
```
JerryHsia-MS, line 111: Is the `move` instruction useless? Might be better with a comment.

tests/stack_integrity_thread.cc
saaramar: Great note, thanks @JerryHsia-MS !

tests/stack_integrity_thread.cc
saaramar: That's annoying. I wanted to have one macro for both callsites to avoid repeating code, but I need operand in the first and not in the later. The `move a0, %1` is there just to use the operand.

Do you have a way to make it generic without it?

tests/stack_integrity_thread.cc
```diff
@@ -8,6 +8,31 @@ using namespace CHERI;
...
+		register auto rfn asm("ct1") = handle;                                 \
+		__asm__ volatile(                                                      \
+		  "1:\n"                                                               \
+		  "auipcc ct2, %%cheri_compartment_pccrel_hi(.compartment_switcher)\n" \
+		  "clc ct2, %%cheri_compartment_pccrel_lo(1b)(ct2)\n"                  \
+		  "" instructions "\n"                                                 \
+		  "cjalr ct2\n"                                                        \
+		  : /* no outputs; we're jumping and probably not coming back */       \
+		  : "C"(rfn), "r"(additional_input)                                    \
+		  : "ct2", "memory" /* in case we return */);                          \
```
davidchisnall, line 33: These clobbers are wrong.  The call clobbers everything except cs0 and cs1

davidchisnall (approved):

PR #29, opened 2023-03-10
Add ISA test

rmn30:
This adds a test that attempts to trigger various ISA corner cases and increase test coverage.


tests/isa-test.cc
```diff
@@ -0,0 +1,769 @@
...
+
+extern "C" enum ErrorRecoveryBehaviour
+compartment_error_handler(struct ErrorState *frame, size_t mcause, size_t mtval)
+{
+	debug_log("Test saw error for PCC {} mcause {} mtval {}",
+	          frame->pcc,
+	          mcause,
+	          mtval);
+	TEST(expectedMCause.has_value(), "Unexpected Crash!");
+	TEST(mcause == *expectedMCause, "mcause should be {}", *expectedMCause);
```
davidchisnall, line 29: Maybe log what it was as well as what it should be?

tests/isa-test.cc
```diff
@@ -0,0 +1,769 @@
...
+	void test_restricted_load(bool          invalidateData,
+	                          PermissionSet capPermissions,
+	                          bool          expectValid,
+	                          PermissionSet expectedPermissions)
+	{
+		Capability capToIntPointer = {myGlobalIntPointer};
+		capToIntPointer.permissions() &= capPermissions;
+		Capability capToGlobalInt = &myGlobalInt;
+		if (invalidateData)
+			capToGlobalInt.invalidate();
```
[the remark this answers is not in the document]
rmn30: Oops, fixed.

tests/isa-test.cc
```diff
@@ -0,0 +1,769 @@
...
+	}
+
+	/**
+	 * Returns a PermissionSet that is the set of permissions that the
+	 * hardware would return if asked to encode PermissionSet `ps` by
+	 * CAndPerms.
+	 *
+	 * The returned PermissionSet will always be a subset of `ps`.
+	 *
+	 * TODO make this a method of PermissionSet?
```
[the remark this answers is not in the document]
rmn30: Will check with @davidchisnall on this one.

tests/isa-test.cc
```diff
@@ -0,0 +1,769 @@
...
+	 */
+	PermissionSet permissions_to_representable(PermissionSet ps)
+	{
+		if (ps.contains(Permission::Execute,
+		                Permission::Load,
+		                Permission::LoadStoreCapability))
+		{
+			/* Executable format */
```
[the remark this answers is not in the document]
rmn30: I've switch all non-doc comments to `//`.

tests/isa-test.cc
```diff
@@ -0,0 +1,769 @@
...
+		     "Call with too small PCC bounds did not crash.");
+	}
+
+	/**
+	 * Tests that check that executing off the end of PCC will fault.
+	 */
+	void test_pcc_bounds()
+	{
+		/* Call aligned_nop_slide with different bounds.
+		    Note that the error PC is always the first byte of the instruction
```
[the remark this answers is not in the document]
rmn30: Switched to `//` as per above.

tests/isa-test.cc
```diff
@@ -0,0 +1,769 @@
...
+	 * address of the load. Obviously we can't allow it to be inlined.
+	 */
+	__noinline int *faulting_load(int *volatile *location)
+	{
+		return *location;
+	}
+
+	/**
+	 * Type of (pointer to) Capability -> Capability function that 'filters' a
+	 * capability. `do_load_test` and `do_store_test` take arguments of this
```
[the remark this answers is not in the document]
rmn30: Fixed.

tests/isa-test.cc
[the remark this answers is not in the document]
davidchisnall: Looks generally useful.  Could also wrap it with 

```c++
constexpr bool is_representable() const
{
    return (*this) == to_representable();
}
```

This would be useful in a bunch of static asserts.

tests/isa-test.cc
```diff
@@ -0,0 +1,774 @@
...
+			test_restrict_capability(CapabilityToLocal, permissionMask);
+			test_restrict_capability(CapabilityToGlobal, permissionMask);
+			test_restrict_capability(CapabilityToFunction, permissionMask);
+		}
+	}
+
+	[[cheri::interrupt_state(disabled)]] void *get_interrupts_disabled_sentry()
+	{
+		void *returnCap;
+		asm volatile("cjal %0, 4" : "=C"(returnCap));
```
davidchisnall, line 307: This is beautifully horrifying, can you add a comment explaining it?

tests/isa-test.cc
```diff
@@ -0,0 +1,774 @@
...
+	/*
+	 * Convenience function for getting the address of pointer. Should compile
+	 * down to cgetaddr.
+	 */
+	template<class T>
+	ptraddr_t address_of_pointer(T *p)
+	{
+		Capability cap{p};
+		return cap.address();
+	}
```
davidchisnall, line 343: Is a call to this really less to type than `Capability{p}.address()`?

rmn30: I guess not. Happy to change the two uses of it.

tests/isa-test.cc
rmn30: It is just a jump and link to the next instruction, putting the link cap in `returnCap`. I guess might be better written with local label rather than hardcoded offset?

tests/isa-test.cc
```diff
@@ -0,0 +1,774 @@
...
+
+	/**
+	 * Function that performs a store of capability `value` through `location`.
+	 * We call this with values we expect to fault and want to know the address
+	 * of the faulting instruction. Since it is so simple it will have no
+	 * prelude so we can use the address of the function as the address of the
+	 * store. Obviously we can't allow it to be inlined.
```
nwf-msr, line 350: 1) I'm not sure about the names; these are not _intrinsically_ going to fault, they're just used that way ("He's not evil, he's just written that way")... maybe `perform_store` or something like that?

2) Perhaps this and `faulting_load` should be in assembler rather than relying on the (sensible, probably not going to change, but still not contractual) behavior of the compiler?

tests/isa-test.cc
```diff
@@ -0,0 +1,774 @@
...
+	}
+
+	/**
+	 * Type of (pointer to) Capability -> Capability function that 'filters' a
+	 * capability. `do_load_test`, `do_store_test` and `do_jalr_test` take
+	 * arguments of this type to allow modifying capabilities operands in
+	 * flexible ways before attempting tests.
+	 */
+	template<class T>
+	using CapabilityFilter = Capability<T> (*)(Capability<T>);
```
nwf-msr, line 376: This really _is_ Haskell in C++! 👍

tests/isa-test.cc
```diff
@@ -0,0 +1,774 @@
...
+	}
+
+	/**
+	 * Run a test where we attempt a capability load via a capability to a
+	 * global int* after passing it through `baseFilter`. Asserts that we
+	 * receive CHERI fault `expectedFault`.
+	 */
+	void do_load_test(CauseCode               expectedFault,
+	                  CapabilityFilter<int *> baseFilter,
+	                  size_t                  anExpectedMCause = MCauseCheri)
```
nwf-msr, line 452: Very minor nit, but maybe the argument order should be `baseFilter, expectedFault, anExpectedMCause`, so that it's read as something like "I expect a load with this filter to cause this fault and this code"?

tests/isa-test.cc
```diff
@@ -0,0 +1,774 @@
...
+		debug_log("Attempting to load via capability with too small bounds.");
+		do_load_test(CauseCode::BoundsViolation, restrict_bounds_filter);
+
+		debug_log("Attempting to load via misaligned capability.");
+		do_load_test(
+		  CauseCode::Invalid, misalign_filter, priv::MCAUSE_LOAD_MISALIGNED);
+	}
+
+	/**
+	 * Run a test where we attempt a capability store of a capability to a local
```
nwf-msr, line 486:
```suggestion
	 * Run a test where we attempt a capability store of a local capability to a
```

tests/isa-test.cc
```diff
@@ -0,0 +1,774 @@
...
+		              misalign_filter,
+		              clear_tag_filter,
+		              priv::MCAUSE_STORE_MISALIGNED);
+
+		debug_log(
+		  "Attempting to store untagged capability via data only pointer");
+		// Storing untagged cap via data-only cap is allowed.
+		do_store_test(
+		  std::nullopt,
+		  permission_filter<PermissionSet{Permission::Load, Permission::Store}>,
```
nwf-msr, line 557: Why not `identity_filter`?

tests/isa-test.cc
```diff
@@ -0,0 +1,774 @@
...
+		debug_log("Attempting to jump to capability with too small bounds");
+		do_jalr_test(CauseCode::BoundsViolation, restrict_bounds_filter);
+	}
+
+	/*
+	 * A convenient place to test PCC related errors. Note that we use 4-byte
+	 * nops so that we can test things like half of an instruction being in
+	 * bounds. Clang doesn't seem to emit a prelude for this function. If it
+	 * does (e.g. at -O0) our expected error PC calculations will be wrong due
+	 * to alignment.
```
nwf-msr, line 678: Maybe these, too, should be in assembler?

nwf-msr (commented):
Generally, this is really great.  The combinator approach works out really nicely, even if C++ will never be as nice as Haskell. ;)

A small pile of nits, but nothing big.

tests/isa-test.cc
```diff
@@ -0,0 +1,718 @@
...
+		// TEST(interruptsEnabledSentry.type() == 3,
+		//      "Expected type 3 but got {}",
+		//      interruptsEnabledSentry.type());
+
+		// Capability interruptsDisabledSentry =
+		// {&get_interrupts_disabled_sentry}; debug_log("interrupts disabled
+		// sentry {}", reinterpret_cast<void*>(interruptsDisabledSentry.get()));
+		// TEST(interruptsDisabledSentry.type() == 2,
+		//      "Expected type 2 but got {}",
+		//      interruptsDisabledSentry.type());
```
nwf-msr, line 265: Is there no good way to get these?

rmn30: I got the linker error in the comment. Suspect it might be a bug @davidchisnall ?

tests/isa-test.cc
rmn30: Because the intent of the this is to check that storing an untagged capability through a capability without MC does not fault (whereas storing a tagged capability would fault, and this is tested previously). Therefore we have to strip MC from the authority cap.

tests/isa-test.cc
rmn30: Filed https://github.com/CHERIoT-Platform/llvm-project/issues/57

PR #31, opened 2023-03-13
Enable interrupts for create functions in the scheduler.

davidchisnall:
These don't touch any scheduler data structures, so can happily be preempted.


saaramar (approved):

saaramar:
Great PR!

JerryHsia-MS (approved):

PR #34, opened 2023-03-17
Fix the software revoker linker script.

davidchisnall:
The old one worked but gave a confusing compartment report (some exports that were called didn't exist).

### DIFF
--- a/sdk/software_revoker.ldscript
+++ b/sdk/software_revoker.ldscript
@@ -8,8 +8,10 @@ SECTIONS
 	{
 		# Space for the compartment's PCC and GDC
 		. = . + 16;
-		# Space for the compartment's ID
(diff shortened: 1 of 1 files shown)


sdk/software_revoker.ldscript
```diff
@@ -6,10 +6,9 @@ SECTIONS
 	. = 0;
 	.compartment_export_table : ALIGN(8)
 	{
-		# Space for the compartment's PCC and GDC
+		# Space for the compartment's PCC and GDC and an empty thingy.
```
rmn30, line 9:
```suggestion
		# Space for the compartment's PCC and GDC and -1 for no compartment_error_handler
```

rmn30 (changes requested):

PR #35, opened 2023-03-21
Deconflate the two sealing uses the switcher uses.

davidchisnall:
The scheduler was able to substitute a trusted stack and an export table entry.  Exploiting this is probably hard but not impossible, so let's make it impossible.

### DIFF
--- a/sdk/core/loader/boot.cc
+++ b/sdk/core/loader/boot.cc
@@ -106,12 +106,17 @@ namespace
 		FirstDataSealingType = 9,
 
 		/**
+		 * The sealing type used for sealed export table entries.
+		 */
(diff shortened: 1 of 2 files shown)


nwf-msr (approved):
LGTM; good find.

saaramar (approved):

saaramar:
Great find :)

sdk/core/loader/boot.cc
```diff
@@ -105,13 +105,18 @@ namespace
 		 */
 		FirstDataSealingType = 9,
 
+		/**
+		 * The sealing type used for sealed export table entries.
```
rmn30, line 109: Why does the comment say export table entries but the name is `ImportTableEntries`?

PR #40, opened 2023-03-23
Unroll the zeroing loop

davidchisnall:
Zeroing the stack is a huge part of the total cost of cross-compartment calls.  Running on Sail (reporting retired instructions), the compartment-switcher benchmark (from #37) reports (stack size, call+return, call, return):

0x100   213     127     86
0x200   405     223     182
0x400   789     415     374
0x800   1557    799     758
0x1000  3093    1567    1526

If we skip stack zeroing, the numbers are 95, 67, 28 for all of them. This means that, even with tiny stacks, zeroing is more than half of the cost of a compartment switch and that cost grows with larger compartments.

We currently require three instructions per store in the zeroing loop. This commit unrolls the loop so that we zero 32 bytes at a time.  The results are now:

0x100   185     113     72
0x200   313     177     136
0x400   569     305     264
0x800   1081    561     520
0x1000  2305    1073    1032

512 bytes is probably the smallest stack size that makes sense for a thread and here we see almost a 25% speedup. This has a fairly noticeable impact on the test suite too:

Before:
Test runner: All tests finished in 2385962 cycles
After:
Test runner: All tests finished in 2087477 cycles

### DIFF
--- a/sdk/core/loader/boot.cc
+++ b/sdk/core/loader/boot.cc
@@ -35,15 +35,30 @@ namespace
 	};
 	constexpr ThreadConfig ThreadConfigs[] = CONFIG_THREADS;
 
+	/**
+	 * Round up to a multiple of `Multiple`, which must be a power of two.
+	 */
+	template<size_t Multiple>
+	constexpr size_t round_up(size_t value)
+	{
+		static_assert((Multiple & (Multiple - 1)) == 0,
+		              "Multiple must be a power of two");
+		return (value + Multiple - 1) & -Multiple;
+	}
+	static_assert(round_up<16>(15) == 16);
+	static_assert(round_up<16>(28) == 32);
+	static_assert(round_up<8>(17) == 24);
+
 	template<size_t N>
 	constexpr size_t total_stacksize(const ThreadConfig (&configs)[N])
 	{
 		size_t ret = 0;
 		for (const auto &config : configs)
 		{
-			ret += config.stackSize;
-			ret += sizeof(TrustedStack) +
-			       sizeof(TrustedStackFrame) * config.trustedStackFrames;
+			ret += round_up<16>(config.stackSize);
+			ret +=
+			  round_up<16>(sizeof(TrustedStack) + sizeof(TrustedStackFrame) *
+			                                        config.trustedStackFrames);
 		}
 		return ret;
 	}
@@ -54,8 +69,12 @@ namespace
 	static_assert(
 	  CheckSize<BOOT_TSTACK_SIZE, sizeof(TrustedStackGeneric<0>)>::Value,
 	  "Boot trusted stack sizes do not match.");
-	alignas(TrustedStack) char __section(".thread_stacks")
+	// Stacks must be 16-byte aligned, so ensure that this is 16-byte aligned.
+	alignas(16) char __section(".thread_stacks")
 	  stackSpace[total_stacksize(ThreadConfigs)];
+	// It must also be aligned sufficiently for trusted stacks, so ensure that
+	// we've captured that requirement above.
+	static_assert(alignof(TrustedStack) <= 16);
 	__END_DECLS
 
 	static_assert(
@@ -723,6 +742,8 @@ namespace
 		                       false>(LA_ABS(stackSpace), sizeof(stackSpace));
 		/// Allocate some space from the pool.
 		auto allocate = [&](size_t size) {
+			Debug::log("Rounded up {} to {}", size, round_up<16>(size));
+			size         = round_up<16>(size);
 			auto ret     = stackArea;
 			ret.bounds() = size;
 			stackArea.address() += size;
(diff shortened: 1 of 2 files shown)


sdk/core/switcher/entry.S
```diff
@@ -97,6 +97,25 @@ switcher_scheduler_entry_csp:
 	bnez               t2, .Lforce_unwind
 .endm
 
+.macro zero_stack base top scratch
```
rmn30, line 100: Add a comment saying that `base` has its address changed and `scratch` is clobbered (obvs).

sdk/core/switcher/entry.S
```diff
@@ -97,6 +97,25 @@ switcher_scheduler_entry_csp:
...
+0:
+	csc                cnull, 0(c\base)
+	csc                cnull, 8(c\base)
+	csc                cnull, 16(c\base)
+	csc                cnull, 24(c\base)
+	cincoffset         c\base, c\base, 32
+	blt                \base, \scratch, 0b
+	bge                \base, \top, 2f
+1:
+	csc                cnull, 0(ct2)
```
rmn30, line 113: What is `ct2` referenced here.

sdk/core/switcher/entry.S
davidchisnall: Added.

sdk/core/switcher/entry.S
davidchisnall: Copy and paste fail, fixed.

sdk/core/switcher/entry.S
rmn30: You forgot to mention that `top` is also modified.

sdk/core/switcher/entry.S
```diff
@@ -95,6 +95,42 @@ switcher_scheduler_entry_csp:
...
+	bge                \base, \scratch, 1f
+	// Zero the stack in 32-byte chunks
+0:
+	csc                cnull, 0(c\base)
+	csc                cnull, 8(c\base)
+	csc                cnull, 16(c\base)
+	csc                cnull, 24(c\base)
+	cincoffset         c\base, c\base, 32
+	blt                \base, \scratch, 0b
+	bge                \base, \top, 2f
```
rmn30, line 126: Should this be `bgt`? If base was 16-byte but not 32-byte aligned we would get here `\cbase == \top` but still have 16 bytes to zero.

sdk/core/switcher/entry.S
```diff
@@ -95,6 +95,42 @@ switcher_scheduler_entry_csp:
...
+1:
+	// Zero any tail in 16-byte chunks
+	csc                cnull, 0(c\base)
+	csc                cnull, 8(c\base)
+	cincoffset         c\base, c\base, 16
+	blt                \base, \top, 1b
```
rmn30, line 132: Can this ever execute more than once?

sdk/core/switcher/entry.S
rmn30: While we are at it you should mention the alignment requirements on base and top.

sdk/core/switcher/entry.S
davidchisnall: Hmm, in fact all of the branches were the wrong way around (lt/le).

sdk/core/switcher/entry.S
davidchisnall: Hmm, shouldn't be able to, in fact.

sdk/core/loader/boot.cc
```diff
@@ -35,15 +35,30 @@ namespace
...
 
+	/**
+	 * Round up to a multiple of `Multiple`, which must be a power of two.
+	 */
+	template<size_t Multiple>
+	constexpr size_t round_up(size_t value)
+	{
+		static_assert((Multiple & (Multiple - 1)) == 0,
+		              "Multiple must be a power of two");
+		return (value + Multiple - 1) & -Multiple;
```
rmn30, line 46: An explanation of the bit-twiddling might be nice, maybe with reference if you found it somehwere?

sdk/core/loader/boot.cc
```diff
@@ -35,15 +35,30 @@ namespace
...
+	}
+	static_assert(round_up<16>(15) == 16);
+	static_assert(round_up<16>(28) == 32);
+	static_assert(round_up<8>(17) == 24);
```
rmn30, line 50: Should include a test for already rounded value:
```suggestion
	static_assert(round_up<8>(17) == 24);
	static_assert(round_up<8>(8) == 8);
```

sdk/core/switcher/entry.S
```diff
@@ -95,6 +95,40 @@ switcher_scheduler_entry_csp:
...
+	andi               t2, sp, 0xf
+	bnez               t2, .Lforce_unwind
+.endm
+
+/**
+ * Zero the stack.  The three operands are the base address (modified during
+ * this call, will point at the top at the end), the top address, and a scratch
+ * register to use.  The base must be a capability but it must be provided
+ * without the c prefix because it is used as both a capability and integer
+ * register.  Top and scratch are both clobbered.
```
rmn30, line 112: Should mention that `base` and `top` must be 16-byte aligned as I think this is important to the implementation. A unit test for this macro would be good...

rmn30 (approved):

PR #42, opened 2023-03-24
Small changes to the heap allocator

nwf-msr:
Fix a corner case bug and fold in @JerryHsia-MS's clever tweak to quarantine handling.


davidchisnall:
This makes the allocator benchmark slightly slower for small sizes, which surprised me - I would expect it to be faster, though the delta is very small.  For large (revocation-bound) sizes, before:

```
sail    65536   3918118
sail    131072  3904099
```

After:

```
sail    65536   3918279
sail    131072  3903962
```

So more or less no difference in retired instructions.

PR #43, opened 2023-03-24
Small fixes in timing output.

davidchisnall:
On Sail, use minstret instead of mcycle, since it gives a useful number and mcycle does not.

### DIFF
--- a/benchmarks/allocation/alloc.cc
+++ b/benchmarks/allocation/alloc.cc
@@ -11,8 +11,12 @@ using Debug = ConditionalDebug<DEBUG_ALLOCBENCH, "Allocator benchmark">;
  */
 void __cheri_compartment("allocbench") run()
 {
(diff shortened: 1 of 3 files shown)


benchmarks/allocation/alloc.cc
```diff
@@ -11,8 +11,12 @@ using Debug = ConditionalDebug<DEBUG_ALLOCBENCH, "Allocator benchmark">;
  */
 void __cheri_compartment("allocbench") run()
 {
+	// Make sure sail doesn't print annoying log messages in the middle of the
```
rmn30, line 14: Is it worth noting that this could be removed once Sail makes s release with this fix: https://github.com/rems-project/sail/commit/362254783e2c0a12b5393169d883ebe5621baa6f

rmn30 (approved):

saaramar (approved):

PR #44, opened 2023-03-29
Interrupt latency benchmark

rmn30:


sdk/core/scheduler/main.cc
```diff
@@ -413,19 +413,19 @@ int __cheri_compartment("sched")
 	}
 	return typed_op<Event>(evt, [&](Event &event) {
 		return event.bits_wait(
-		  retBits, bitsToWait, clearOnExit, waitAll, timeout);
+		  retBits, bitsToWait, flags & EventWaitClearOnExit, flags & EventWaitAll, timeout);
```
davidchisnall, line 416: It would be nice to wrap this in the same kind of template that I did for PAL features in snmalloc.  Something like:

```
template<EventWaitFlags Desired>
bool has_event_flag(int eventFlags)
{
    return eventFlags & int(Desired);
}
```

I am always somewhat nervous of raw `&` for checking bitfields because you're relying on the implicit conversion to bool, which can happen in a different place if you don't put enough brackets around the expression.  Putting it in a separate function forces it to happen at the evaluation of the return and better expresses intentionality.

sdk/core/scheduler/main.cc
```diff
@@ -403,8 +404,7 @@ int __cheri_compartment("sched")
                   void     *evt,
                   uint32_t *retBits,
                   uint32_t  bitsToWait,
-                  bool      clearOnExit,
-                  bool      waitAll)
+                  int       flags)
```
davidchisnall, line 407: Can we use a fixed-width type here please?

sdk/include/event.h
```diff
@@ -48,8 +58,22 @@ int __cheri_compartment("sched") event_bits_wait(Timeout  *timeout,
...
+ */
+enum [[clang::flag_enum]] EventSetFlags {
+  /** 
+    * Do not yield immediately if a higher priority thread becomes runnable.
+    * Instead the other thread will run when this thread yields or another
+    * scheduler event occurs (e.g. a timer interrupt). This may be useful to
+    * avoid multiple context switches when setting bits in multiple event
+    * groups from a low priority thread.
+    */
+  EventSetNoYield = (1 << 0)};
```
davidchisnall, line 76: I think GitHub has screwed up rendering the diff here, there's no } for this enum definition in the UI.

rmn30: It is there but on the same line as the `)`. `clang-format` seems happy with this.

sdk/include/event.h
```diff
@@ -98,3 +121,24 @@ int __cheri_compartment("sched")
   event_delete(struct SObjStruct *heapCapability, void *evt);
 
 __END_DECLS
+
+#ifdef __cplusplus
```
davidchisnall, line 125: Not sure these should go in the public header: I don't see them as being useful for anything other than the scheduler (a helper for constructing the flags would be).

davidchisnall (approved):

rmn30:
Marking as draft due to unresolved issues especially around `yield_interrupt_enabled`.

davidchisnall (approved):

PR #45, opened 2023-03-31
Initial plumbing for scheduler capabilities.

davidchisnall:
This should allow the scheduler to export sealing types, which can then be used without the allocator to authorise scheduler operations.  There are no consumers of this yet, but eventually it should be used for statically defining event channel endpoints, especially the ones coupled to interrupt sources, so that we can statically audit who can monitor interrupts.

### DIFF
--- a/sdk/core/loader/boot.cc
+++ b/sdk/core/loader/boot.cc
@@ -544,6 +544,12 @@ namespace
 				        Root::Type::RWGlobal,
 				        PermissionSet{Permission::Load, Permission::Store}>(
 				    target);
+				// Is the software sealing type owned by the scheduler?  If so,
+				// we're going to seal the object with the scheduler's sealing
+				// type, not the allocator's.  This lets the scheduler export
+				// software-defined capabilities without adding the allocator
+				// to the TCB for availability.
+				bool isSchedulerObject = false;
 				// TODO: This currently places a restriction that data memory
 				// can't be in the low 64 KiB of the address space.  That may be
 				// too restrictive. If we haven't visited this sealed object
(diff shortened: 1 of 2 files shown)


davidchisnall:
Probably not worth merging until we have a consumer, but if anyone wants to extend Robert's wake-without-yielding example to require an authorising capability then this is a good starting point.

sdk/core/scheduler/common.h
```diff
@@ -35,7 +35,7 @@ namespace sched
 		/**
 		 * The real type of this subclass.
 		 */
-		enum class Type : uint8_t
+		enum class Type : uint32_t
```
nwf-msr, line 38: Add a comment that this type is deliberately chosen for compatibility with the allocator-based mechanism (xref `sdk/core/allocator/token.h:/^struct SObjStruct`)?

sdk/core/loader/boot.cc
```diff
@@ -568,14 +574,11 @@ namespace
                         }
                         return false;
 					};
-					bool found = false;
-					for (auto &compartment : image.privilegedCompartments)
+					bool found = findExport(image.allocator());
+					if (!found && findExport(image.scheduler()))
```
nwf-msr, line 578: It's possibly a bit of a pity that the scheduler isn't represented as a `PrivilegedCompartment` and so we have to special-case these two.

davidchisnall: They are both `PrivilegedCompartment`s, the old code just iterated over them, but since we're now doing different things for the two privileged compartments that doesn't make sense anymore.

nwf-msr (approved):

PR #49, opened 2023-04-03
Port mibench benchmarks to CHERIoT

JerryHsia-MS:
These benchmarks are the same ones used in the CheriRTOS paper.


davidchisnall:
Thanks for working on this.  I don't think that we should pull in the MIBench source files directly, they should be added as a submodule and wrapped in compatment entry points.  This probably doesn't belong in the test suite, it should go in `benchmarks/mibench`.

PR #59, opened 2023-04-18
Use stack low water mark to optimise zeroing.

rmn30:
This is implemented in Sail but not yet merged:
https://github.com/microsoft/cheriot-sail/pull/8

This is on top of: #57 which fixes a bug discovered while implementing this.

Enabled by xmake option --stack-low-water-mark=true.  We need to think more about potential attacks based on unexpected stack pointer bounds or passing pointers to stack.

### DIFF
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -200,8 +200,9 @@ compartment_switcher_entry:
 	and                gp, gp, ~0xf
 	// Skip zeroing if high water mark >= stack pointer
 	bge                t2, sp, after_zero
-	// Use stack high water mark as base address for zeroing
-	// XXX could be out of bounds / unrepresentable if bad csp?
+	// Use stack high water mark as base address for zeroing.  If this faults
+	// then it will trigger a force unwind.  This can happen only if the caller
+	// is doing something bad.
 	csetaddr           ct2, csp, gp
 #endif
 	zero_stack         t2, s0, gp


sdk/core/switcher/tstack.h
```diff
@@ -44,33 +44,41 @@ struct TrustedStackFrame
...
 	uint8_t inForcedUnwind;
 	// Padding up to multiple of 16-bytes.
-	uint8_t  pad0;
-	uint16_t padding[2];
+#ifdef CONFIG_MSLWM
+#	define TRUSTED_STACK_PADDING 13
+#else
+#	define TRUSTED_STACK_PADDING 5
+#endif
+	uint8_t padding[TRUSTED_STACK_PADDING];
```
davidchisnall, line 81:
```suggestion
uint8_t padding[
#ifdef CONFIG_MSLWM
13
#else
5
#endif
];
```

sdk/core/switcher/entry.S
```diff
@@ -181,8 +180,24 @@ compartment_switcher_entry:
 	cgetbase           s1, csp
 	csetaddr           csp, csp, s1
 	sub                s1, s0, s1
-	csetboundsexact    csp, csp, s1
-	zero_stack         sp, s0, gp
+	csetboundsexact    ct2, csp, s1
+	csetaddr           csp, ct2, s0
+#ifdef CONFIG_MSLWM
+	// read and align the stack low water mark
+	csrr               gp, 0xbc1 // mslwm
```
davidchisnall, line 187: Please can you turn 0abc1 and 0xbc2 into symbolic constants?

sdk/core/switcher/entry.S
```diff
@@ -181,8 +180,24 @@ compartment_switcher_entry:
...
+	and                gp, gp, ~0xf
+	// skip zeroing if low water mark >= stack poitner
+	bge                t2, sp, after_zero
```
davidchisnall, line 190: We probably need to zero the whole thing if the address is not in bounds.  I think that is not needed on the return path though.

rmn30: I think it might be better to fault and unwind to the caller. Here are some scenarios (`hwm` == high water mark):

1. `csp.addr > csp.top` this is possible before this PR. It could happen only if the caller messes with `csp`. It would trap in `csetboundsexact`. According to the comment this will cause a forced unwind because it is after the update to the tstack offset. 
3. `hwm > csp.addr` this can happen in normal operation and is explicitly handled by not zeroing anything (we maintain an invariant that the stack between base and `hwm` is always zero).
4. `hwm > csp.top` this could happen if the caller reduces the length of csp before calling, which would be weird. What happens depends on the value of `csp.addr`. If `hwm > csp.addr` then we skip zeroing as above. This is harmless and the call proceeds with the smaller stack. If `hwm <= csp.addr` we will proceed to zeroing but will trap in `zero_stack`, either because `ct2` became unrepresentable on `csetaddr` and had its tag cleared or because we eventually try to zero above `csp.top`. This trap should cause a forced unwind.
5. `csp.addr < csp.base` the current capability encoding makes this impossible. If `csp.addr` is moved below `csp.base` it's tag will be cleared and this will be detected in `check_compartment_stack_integrity`
6. high water mark `< csp.base` This is possible only if the caller moves `csp.base`, which would be weird. In this case `csetaddr` would make `ct2` unrepresentable, which would clear the tag causing a trap in `zero_stack` again.

If traps are handled as the described in the comment I don't think any special handling is required.

sdk/xmake.lua
```diff
@@ -232,6 +232,11 @@ rule("firmware")
 			add_defines("SIMULATION")
 		end
 
+		if board.stack_high_water_mark then
```
davidchisnall, line 235: At least the Sail json should probably be updated to include this.

rmn30: OK, once the Sail PR is merged I'll do that here. I need to update it to say high water mark instead of low water mark.

rmn30:
I have realised there is a big problem with the error handling path in this PR: if the error handler returns 'install context' then a stale version of the high water mark is restored which doesn't account for the stack use of the call to the error handler. This could result in leaking capabilities on the next compartment call return due to not zeroing enough stack. To fix we could either update the high water mark in the stored context after calling the error handler or skip restoring it on that path. I think the force unwind path is OK because it uses the version of the water mark straight from the CSR.

sdk/core/switcher/entry.S
```diff
@@ -607,6 +646,13 @@ exception_entry_asm:
 
 	// Load the trusted stack pointer to ct1
 	cspecialr          ct1, mtdc
+#ifdef CONFIG_MSHWM
+	// Update the spilled copy of the stack high watermark to ensure that we
```
rmn30, line 650:
```suggestion
	// Update the spilled copy of the stack high water mark to ensure that we
```
Would be nice to be consistent to make searching easier.

davidchisnall: Technically, it should be high-water mark, but this is probably fine.

sdk/core/switcher/entry.S
davidchisnall: I need to think about this a bit more.  It is fine for the caller to restrict the stack before calling (which it might choose to do to avoid some stack clearing on the way out if it knows the callee doesn't need much stack).  I think we check the permissions on csp, so there's no way for it to be anything other than something derived from a valid csp here.

davidchisnall: Thinking a bit more, case 3 is interesting because the correct behaviour here is to zero all of csp but *not* update the SHWM register.  Possibly we should disallow this entirely.

davidchisnall: Updated so that we now just force unwind if we fault anywhere in the switcher.  This is simpler than special casing a load of things.

davidchisnall (approved):

PR #63, opened 2023-04-19
Add a variant of compartment call benchmark that actually uses some s…

rmn30:
…tack.

This is to check the effect of the stack low water mark. Also tests to make sure the stack is being properly cleared.


rmn30:
This is benchmark checks for stack zeroing so could be a useful test for #111 . It is based on a version of the compartment call benchmark before it was updated (in #92) to make it more consistent but this might not matter.

davidchisnall (approved):

PR #65, opened 2023-04-25
Fix the no-temporal-safety mode.

davidchisnall:
This should never be used in production but helps provide a baseline for the cost of various hardware features.


nwf-msr (approved):
LGTM, just one nit.

PR #70, opened 2023-05-11
Remove some checks in the switcher.

davidchisnall:
That sounds like a terrible idea, but now that we are force unwinding on *any* fault in the switcher, we can reduce the number of special-case checks and reduce the complexity.

Most of this is updating (and simplifying) the tests because now we do call the error handler if the stack exists and has sufficient permissions to spill the register frame, but the error handler may fault and force unwind on entry when it tries to load them (or fail in exciting ways later).  Generally, if the stack is broken then there isn't a huge amount that the error handler can do, so it's probably a good idea to just give up.


sdk/core/switcher/entry.S
```diff
@@ -550,19 +547,12 @@ exception_entry_asm:
 	// Fetch the base of compartment stack before cincoffset for later
 	// comparison. The subsequent cincoffset could cause the base to change,
 	// if the capability becomes unrepresentable. Even though that would clear
-	// the tag, which we will detect in check_compartment_stack_integrity.
+	// the tag, will cause a trap later that will force unwind.
 	cgetbase           tp, ct0
-	// Allocate space for the register save frame on the stack.
+	// Allocate space for the register save frame on the stack.  If we didn't
+	// have enough space here, we'll fault in the unwind path, which will
```
rmn30, line 553: Should say fault in the error handler path?

rmn30 (approved):
Bit scary, but looks OK. Another test I'd like to see is corrupt `csp` on return from error handler.

PR #71, opened 2023-05-12
Add a fast accessor for the thread ID.

davidchisnall:
This might be useful for TLS abstractions and will be necessary for priority-inheriting futexes.

### DIFF
--- a/sdk/include/thread.h
+++ b/sdk/include/thread.h
@@ -45,6 +45,13 @@ typedef struct
   thread_id_get(void);
 
 /**
+ * Returns a cacheable (read-only) pointer to a global owned by the scheduler
+ * that contains the current thread ID.  Reading this pointer will return the
+ * thread ID of the currently running thread.
+ */
+__cheri_compartment("sched") uint16_t *thread_id_get_pointer(void);
+
+/**
  * Returns the number of cycles accounted to the idle thread.
  *
  * This API is available only if the scheduler is built with accounting support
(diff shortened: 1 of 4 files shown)


sdk/include/thread.h
```diff
@@ -44,6 +44,12 @@ typedef struct
 [[cheri::interrupt_state(disabled)]] uint16_t __cheri_compartment("sched")
   thread_id_get(void);
 
+/**
+ * Returns a cacheable (read-only) pointer to a global owned by the scheduler
+ * that contains the current thread ID.  Reading this pointer will return the
```
rmn30, line 49: Incomplete sentence.

rmn30 (changes requested):
Just the one incomplete comment. 

Will the _Atomic(uint16_t) compile to normal `lh` / `sh`?

rmn30:
For TLS are you imaging a global array indexed by thread ID? In that case I guess there is nothing stopping one thread from interfering with another thread's TLS.

rmn30:
Maybe it would be better for the scheduler to have a `get_tls_pointer` accessor that would return a pointer to a pointer to the current thread's TLS in a similar way that this PR does thread ID. You could also have an `allocate_tls` scheduler function to allocate TLS for the current thread. TLS should probably not have global perm.

davidchisnall:
TLS must be per-compartment and probably per compartment invocation.  I'm imagining doing emutls with the TLS pointer at the start of the stack on entry and everything cleaned up on compartment return.

rmn30 (approved):

PR #77, opened 2023-05-25
Use imprecise setbounds for the software revoker's regions.

### LINKED ISSUE #76: Large stacks can break the software revoker
A large stack region can cause the software revoker to get an untagged capability for the stacks area.  The loader should round the bounds up a bit to ensure that it has access to the entire range, even if it means having slightly wider access.

davidchisnall:
This ensures that we have a valid tagged capability that spans the regions that must be scanned.  We may also scan some regions that cannot contain revocable capabilities.

Fixes #76

### DIFF
--- a/sdk/core/loader/boot.cc
+++ b/sdk/core/loader/boot.cc
@@ -1214,7 +1214,19 @@ extern "C" SchedulerEntryInfo loader_entry_point(const ImgHdr &imgHdr,
 	// accesses these in the revoker is very small and amenable to auditing
 	// (the only memory accesses are a load and a store back at the same
 	// location, with interrupts disabled, to trigger the load barrier).
-	auto scaryCapabilities = build<void *, Root::Type::RWStoreL>(
+	//
+	// We use imprecise set-bounds operations here because we need to ensure
+	// that the regions are completely scanned and scanning slightly more is
+	// not a problem unless the revoker is compromised.  The software revoker
+	// already has a terrifying set of rights, so this doesn't really make
+	// things worse and is a nother good reason to use a hardware revoker.
+	// Given that hardware revokers are lower power, faster, and more secure,
+	// there's little reason for the software revoker to be used for anything
+	// other than testing.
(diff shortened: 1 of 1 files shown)


sdk/core/loader/boot.cc
```diff
@@ -1214,7 +1214,19 @@ extern "C" SchedulerEntryInfo loader_entry_point(const ImgHdr &imgHdr,
 	// accesses these in the revoker is very small and amenable to auditing
 	// (the only memory accesses are a load and a store back at the same
 	// location, with interrupts disabled, to trigger the load barrier).
-	auto scaryCapabilities = build<void *, Root::Type::RWStoreL>(
+	//
+	// We use imprecise set-bounds operations here because we need to ensure
+	// that the regions are completely scanned and scanning slightly more is
+	// not a problem unless the revoker is compromised.  The software revoker
+	// already has a terrifying set of rights, so this doesn't really make
+	// things worse and is a nother good reason to use a hardware revoker.
```
nwf-msr, line 1222:
```suggestion
	// things worse and is another good reason to use a hardware revoker.
```

sdk/core/loader/boot.cc
```diff
@@ -1214,7 +1214,19 @@ extern "C" SchedulerEntryInfo loader_entry_point(const ImgHdr &imgHdr,
...
+	// Given that hardware revokers are lower power, faster, and more secure,
+	// there's little reason for the software revoker to be used for anything
+	// other than testing.
+	auto scaryCapabilities = build<void *,
+	                               Root::Type::RWStoreL,
+	                               Root::Permissions<Root::Type::RWStoreL>,
+	                               false>(
```
nwf-msr, line 1229: Since we aren't doing the type-level enumeration thing, perhaps
```suggestion
	                               /* Precise: */ false>(
```

nwf-msr (approved):
Two minor nits if you feel like it; otherwise LGTM.

PR #78, opened 2023-05-25
Round the heap region in the loader.

### LINKED ISSUE #75: Heaps are insufficiently aligned on systems with > 256 KiB of RAM
The heap start is 512-byte aligned.  This is fine for heaps up to 256 KiB, but above that causes problems.  The loader should treat the heap import region specially and align the start correctly for the available size.

davidchisnall:
We were always aligning the heap to a 512-byte boundary.  This works on 256 KiB platforms (though is too aggressive if > 50% of the memory is used for code and globals) and works 50% of the time on 512-byte platforms.  Do the rounding in the loader so that it can be based on the total size.

Fixes #75

### DIFF
--- a/sdk/core/loader/boot.cc
+++ b/sdk/core/loader/boot.cc
@@ -460,11 +460,41 @@ namespace
 					    (size == (LA_ABS(__export_mem_heap_end) -
 					              LA_ABS(__export_mem_heap))))
 					{
+						ptraddr_t end = target + size;
 						Debug::log(
-						  "Assigning the heap ({}--{}) to the allocator",
+						  "Rounding heap ({}--{}) region", target, end);
+						size_t sizeMask =
+						  __builtin_cheri_representable_alignment_mask(size);
+						Debug::log("Applying mask {} to size", sizeMask);
+						size_t    roundedSize = size & sizeMask;
+						ptraddr_t roundedBase = end - roundedSize;
+						Debug::log(
+						  "Rounding heap size down from {} to {} (rounded up "
+						  "to {})",
+						  size,
+						  roundedSize,
+						  __builtin_cheri_round_representable_length(size));
+						Debug::Invariant(
+						  (end & ~sizeMask) == 0,
+						  "End of heap ({}) is not sufficiently aligned ({})",
+						  end,
+						  sizeMask);
+						Debug::log(
+						  "Assigning rounded heap (in {}--{}) to the allocator",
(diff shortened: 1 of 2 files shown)


sdk/core/loader/boot.cc
```diff
@@ -460,11 +460,44 @@ namespace
...
+						size_t roundedSize = size & sizeMask;
 						Debug::log(
-						  "Assigning the heap ({}--{}) to the allocator",
+						  "Rounding heap size down from {} to {} (rounded up "
+						  "to {})",
+						  size,
+						  roundedSize,
+						  __builtin_cheri_round_representable_length(size));
+						ptraddr_t end         = target + size;
+						ptraddr_t roundedBase = end - roundedSize;
```
nwf-msr, line 484: This is assuming that `end` is suitably aligned.  We should probably assert that `(end & ~sizeMask) == 0`?

That is, I don't think it's worth trying to compensate for the case that it isn't, because it really should be, but all the same, an assertion wouldn't go amiss.

sdk/core/loader/boot.cc
```diff
@@ -460,11 +460,44 @@ namespace
...
+						static constexpr size_t MantissaBits = 9;
+						static constexpr size_t MantissaMask =
+						  (1 << MantissaBits) - 1;
+						size_t sizeBits = 32 - __builtin_clz(size);
+						size_t sizeMask = (sizeBits <= MantissaBits)
+						                    ? MantissaMask
+						                    : MantissaMask
+						                        << (sizeBits - MantissaBits);
+						Debug::log("Applying mask {} to size", sizeMask);
+						size_t roundedSize = size & sizeMask;
```
nwf-msr, line 476: Isn't that equivalent to
```suggestion
						size_t sizeMask =
							__builtin_cheri_representable_alignment_mask(size);
						Debug::log("Applying mask {} to size", sizeMask);
						size_t roundedSize = size & sizeMask;
```

The intermediate steps are different, in that CRAM is going to give you all ones to the left of your `MantissaMask`. In the end, though, that shouldn't matter, because all those bits in `size` are zero from the `__builtin_clz()`, and so they'll be zero in `roundedSize`.

nwf-msr (changes requested):

sdk/core/loader/boot.cc
davidchisnall: Yup, that works nicely.

PR #79, opened 2023-05-26
Add a design document for allocator claims.

davidchisnall:


docs/AllocatorClaims.md
```diff
@@ -0,0 +1,120 @@
...
+
+This document describes the design of the claim model that the allocator should provide and how it should be implemented.
+
+The problem
+-----------
+
+Consider the scenario where compartment A passes a heap buffer to compartment B.
+Another thread executing in compartment A may free the buffer in the middle of B's execution.
+This will cause B to fault.
+This is a concurrent operation and so is a classic TOCTOU problem.
```
nwf-msr, line 12:
```suggestion
This is a concurrent operation and so sets the stage for classic TOCTOU problems.
```
as we haven't actually said that B has performed any Checks yet.

docs/AllocatorClaims.md
```diff
@@ -0,0 +1,120 @@
...
+The first approach is simply to punt on the problem entirely.
+When an OS kernel interacts with userspace memory, it calls explicit helpers to copy data to and from the kernel.
+This is necessary because the userspace process may provide invalid pointers or update the memory map to invalidate pointers while the kernel runs.
+
+In many situations, the same approach could work with CHERIoT.
+We could provide a safe memcpy as a library function that runs with interrupts disabled, checks the source and destination pointers, and then performs the copy (up to some bounded size) and reports whether the copy succeeded.
+This would be sufficient in a lot of cases but is not a very friendly programmer model.
+In particular, it means that any compartment wrapping existing unmodified code would have to either (deep) copy any objects passed in or would have to make invasive changes to the wrapped code.
+
+Note: This approach can be orthogonal to others and so we should implement it anyway.
```
nwf-msr, line 46: :+1:

docs/AllocatorClaims.md
```diff
@@ -0,0 +1,120 @@
...
+
+Possible approach 1: Hazard pointers
+------------------------------------
+
+Hazard pointers provide inspiration for the first possible solution.
+This approach involves having a per-compartment list of held objects that the allocator must consult before freeing an object.
+The allocator would provide an API that allowed a compartment to register a table containing a (bounded) array of held objects and an allocation capability.
+On deallocation, the allocator would need to traverse all such lists to determine whether an object is on any of them.
```
nwf-msr, line 54: This traversal would have to be done with interrupts off.  In LF SMR's use of hazards, there's a global cooperative assumption that threads will not share their local copies of pointers, so that the number of copies of a given pointer in the hazard lists is strictly decreasing with time.  That's not necessarily true of this use of hazards even without considering malice: compartment B might forward a pointer it's holding via a hazard to compartment C, which might then hold it, even after A has attempted to privatize and deallocate it.

docs/AllocatorClaims.md
```diff
@@ -0,0 +1,120 @@
...
+With a simple reference counting model, the original attack is still possible:
+
+1. A allocates an object.
+2. A passes a pointer to the object to B.
+3. B increments the reference count.
+4. A decrements the reference count twice from another thread.
+5. B traps.
+
+Each reference to the object is now state that must be tracked and, most likely, will require an allocation.
+We currently have 14 bits spare in the allocation header (slightly more if we shrink the allocator ID, which does not actually require a full 16 bits, since a 16-bit allocator ID would require 2 MiB of RAM to hold all of the corresponding allocation capabilities, leaving no space for the heap) and so could store a heap-start-relative (shifted for alignment) pointer to the head of a linked list of claims.
```
nwf-msr, line 80: In the spirit of "O(n) but n is 3", it could suffice to just have a single list of all claims.  If that's too flippant, a small hash table keyed on address could also work.  In either case, we'd just need one bit in the header to bring the (potential) claims to deallocation's attention.

docs/AllocatorClaims.md
```diff
@@ -0,0 +1,120 @@
...
+In the common case, this field will be 0 (the start of the heap cannot be allocated) and so free can skip it.
+
+The proposed solution involves constructing a linked list of structures holding an allocator ID, a saturating reference count, and a next pointer.
```
nwf-msr, line 83: What happens on saturation, the object becomes immortal?

nwf-msr (commented):
The current sketch of the hazard approach is that there's an initialization call to the allocator and then zero calls per object use; the explicit claims approach below has no initialization but two calls per object use.

I wonder how you'd feel about a design where a compartment had to make _one_ call per use, to mark the object as possibly having associated hazards.  This would
- let the allocator forgo hazard handling for objects that had never been so registered
- check whether the object had been the target of deallocation, even if it is still live due to other extant hazards, prior to allowing this caller to add another hazard.

It also possibly addresses the "when do things actually get freed" problem. A compartment can be pinning no more objects than it has hazard pointer slots, and so asking to create a hazard on an object gives the allocator an opportunity to compare the compartment's current hazards with the set of objects whose reclamation had been deferred due to a past hazard found in the calling compartment.

In terms of potential abuse, failing to call that API means that things might get freed out from under you, and calling that API unnecessarily just means that the allocator has to do more work on deallocation.

docs/AllocatorClaims.md
davidchisnall: Yup, if you claim an object 2^32 times (or possibly 2^16 times), it's yours forever and you get to keep it.

For reference, Smalltalk-80 did the same thing, only with 7, not 2^16, and that worked surprisingly well.

docs/AllocatorClaims.md
```diff
@@ -0,0 +1,120 @@
...
+Note: This approach can be orthogonal to others and so we should implement it anyway.
+
+Possible approach 1: Hazard pointers
+------------------------------------
+
+Hazard pointers provide inspiration for the first possible solution.
+This approach involves having a per-compartment list of held objects that the allocator must consult before freeing an object.
+The allocator would provide an API that allowed a compartment to register a table containing a (bounded) array of held objects and an allocation capability.
+On deallocation, the allocator would need to traverse all such lists to determine whether an object is on any of them.
+If the object is found on a list then the metadata should be updated to mark it as owned by the allocation capability associated with the list.
```
nwf-msr, line 55: Ack; I had a comment here and I'm not sure where it went.  :(

In any case: how is compartment B notified that ownership of the object has transferred from compartment A?  When does either compartment B or the allocator know to free the object?

docs/AllocatorClaims.md
```diff
@@ -0,0 +1,120 @@
...
+
+This is attractive for the fast-path case, because the compartment wishing to claim an object for the duration of a call needs only to insert a pointer to it into a pre-registered array.
+It is unclear whether it is possible to use this interface securely.
+In particular, a malicious compartment could allocate a large object and pass a pointer to a small sub-object to the callee.
+The callee has no mechanism (other than calling the allocator) to know whether it holds a pointer to an entire object.
```
rmn30, line 60: The allocators traversal could do a subobject check, although we could really do with an `issubset` instruction for that.

nwf: The CHERIoT `malloc` has no particularly convenient way to map from a sub-object to its allocation, since it's a `dlmalloc`-style inline structure.  It'd be O(n) to walk the heap, uncomfortable with or tricky without holding the lock. :(

docs/AllocatorClaims.md
davidchisnall: That's part of the problem.  Logically, as soon as you insert something to a hazard list, you logically own it: it lazily counts towards your quota.  You must ensure that you have sufficient quota for it, but you can't easily query its size.

docs/AllocatorClaims.md
nwf-msr: Oh, right, yes, I forgot that we marked headers that way.

PR #82, opened 2023-06-02
Add a claims API to the allocator.

davidchisnall:
This makes it possible to claim a heap allocation so that it counts towards your quota (in addition to any other claims and the original owner's quota) and cannot be freed until everyone relinquishes their claims.

Claims can also apply to sub-objects and claim the entire object.

`heap_free` is a unified point for freeing and dropping claims.  This allows freeing heap pointers without worrying about who originally allocated them.

### DIFF
--- a/sdk/core/allocator/main.cc
+++ b/sdk/core/allocator/main.cc
@@ -306,8 +306,7 @@ namespace
 		if (capability->identifier == 0)
 		{
 			static uint32_t nextIdentifier = 1;
-			if (nextIdentifier >
-			    std::numeric_limits<decltype(capability->identifier)>::max())
+			if (nextIdentifier >= (1 << MChunkHeader::OwnerIDWidth))
 			{
 				return nullptr;
 			}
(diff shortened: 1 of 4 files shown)


sdk/core/allocator/main.cc
```diff
@@ -316,29 +315,395 @@ namespace
...
+		 * The identifier of the owning allocation capability.
+		 */
+		uint16_t allocatorIdentifier = 0;
+		/**
+		 * Next 'pointer' encoded as a shifted offset from the start of the
+		 * heap.
+		 */
+		uint16_t encodedNext = 0;
+		/**
+		 * Saturating reference count.  We use a single
```
nwf-msr, line 334: A single...?

sdk/core/allocator/main.cc
```diff
@@ -316,29 +315,395 @@ namespace
...
+		// If there is no claim, fail.
+		if (claim == nullptr)
+		{
+			return false;
+		}
+		if (!reallyDrop)
+		{
+			return true;
+		}
+		// Drop the reference.  If this results in the last claim going away,
```
nwf-msr, line 642:
```suggestion
		// Drop the reference.  If this results in the last reference going away,
```

sdk/core/allocator/main.cc
```diff
@@ -316,29 +315,395 @@ namespace
...
 		{
 			return -EINVAL;
 		}
-		return gm->mspace_free(
-		  mem, capability->quota, capability->identifier, reallyFree);
+		ptraddr_t start    = chunk->body().address();
+		size_t    bodySize = gm->chunk_body_size(*chunk);
+		// Is the pointer that we're freeing a pointer to the entire allocation?
+		bool isPrecise = (start == mem.base()) && (bodySize == mem.length());
+		// If this is a precise allocation, see if we can
```
nwf-msr, line 679: if we can... free it now, because it is otherwise unclaimed?

PR #87, opened 2023-06-26
Attempt to fix problem with command run in dev container.

rmn30:
When creating a new github codespace the postAttachCommand from the devcontainer.json was not being run automatically. This might be because onCreateCommand is more appropriate.


rmn30:
I created a new codespace to test and it seems to work better.

davidchisnall (approved):

davidchisnall:
Approved and rebased.

PR #88, opened 2023-06-27
Fix the allocator benchmark with no revoker.

davidchisnall:
We use the shadow bits to check for object start, when we don't have them just pretend that we do.  This is totally unsafe, but the mode with no shadow bits should never be used for anything other than a baseline in benchmarks.

### DIFF
--- a/sdk/core/allocator/alloc.h
+++ b/sdk/core/allocator/alloc.h
@@ -1223,13 +1223,24 @@ class MState
 		{
 			return nullptr;
 		}
-		if (revoker.shadow_bit_get(address))
-		{
-			return nullptr;
-		}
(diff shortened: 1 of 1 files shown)


sdk/core/allocator/alloc.h
```diff
@@ -1223,13 +1223,24 @@ class MState
 		{
 			return nullptr;
 		}
-		if (revoker.shadow_bit_get(address))
+		// If we don't have shadow bits, skip this check and just assume that
+		// this is the start of a capability.  This mode is for benchmark
```
rmn30, line 1227:
```suggestion
		// this is the start of an allocation.  This mode is for benchmark
```

rmn30 (approved):
Slight brain-o in comment, I think.

PR #89, opened 2023-06-30
Add some docs for writing a device driver

davidchisnall:


docs/WritingADeviceDriver.md
```diff
@@ -0,0 +1,220 @@
...
+    ],
+```
+
+The first part says that the ethernet device's MMIO space is 0x204 bytes, starting at address 0x98000000.
+The second says that interrupt number 16 is used for the ethernet device.
+
+Accessing the memory-mapped I/O region
+--------------------------------------
+
+The `MMIO_CAPABILITY` macros is used to get a pointer to memory-mapped I/O devices.
```
rmn30, line 48:
```suggestion
The `MMIO_CAPABILITY` macro is used to get a pointer to memory-mapped I/O devices.
```

docs/WritingADeviceDriver.md
```diff
@@ -0,0 +1,220 @@
...
+volatile struct EthernetMMIO *ethernet_device()
+{
+    return MMIO_CAPABILITY(struct EthernetMMIO, ethernet);
+}
+```
+
+Note that this macro must be used in code, it cannot be used for static initialisation.
+The macro expands to a load from the compartment's import table and so there is no point assigning the result to a global: you will get smaller code using it directly.
+
+Now that you have a `volatile` pointer to the device's MMIO region, you can access its control registers directly.
```
rmn30, line 69: I hate to nit pick but isn't it the struct that is volatile, not the pointer?

docs/WritingADeviceDriver.md
```diff
@@ -0,0 +1,220 @@
...
+This macro takes four arguments:
+
+ 1. The name that we're going to use to refer to this capability.
+    The name `ethernetInterruptCapability` is arbitrary, you can use whatever makes sense to you.
+ 2. The name of the interrupt, from the board description file (`Ethernet`, in this case).
+ 3. Whether this capability authorises waiting for this interrupt (this will almost always be `true`).
+ 4. Whether this capability authorises acknowledging the interrupt so that it can fire again.
+    This will almost always be true in device drivers but should generally be true for only one compartment (for each interrupt), whereas multiple compartments may wish to observe interrupts for monitoring.
+
+As with the MMIO capabilities, compartments
```
rmn30, line 113: incomplete sentence.

docs/WritingADeviceDriver.md
```diff
@@ -0,0 +1,220 @@
...
+            "provided_by": "build/cheriot/cheriot/release/example-firmware.scheduler.compartment",
+            "symbol": "__export.sealing_type.sched.InterruptKey"
+          }
+```
+
+The sealing type tells you that this is an interrupt capability (it's sealed with the `InterruptKey` type, provided by the scheduler).
+The contents lets you audit what this authorises.
+The first two bytes are a 16-bit (little-endian on all currently supported targets) integer containing the interrupt number, so 1000 means 16 (our Ethernet interrupt number).
+The next two bytes are boolean values reflecting the last two arguments to the macro, so this authorises both waiting and clearing the macro.
+Again, this can form part of your firmware auditing.
```
rmn30, line 131: We really need tooling for this. Manually interpreting binary representation of structs is not fun.

docs/WritingADeviceDriver.md
```diff
@@ -0,0 +1,220 @@
...
+
+Now that you're authorised to handle interrupts, you will need something that you can wait on.
+Each interrupt is mapped to a futex word, which can be used with scheduler waiting primitives.
+You can get the word associated with an interrupt by passing the authorising capability to the `interrupt_futex_get` function exported by the scheduler:
+
+```c
+const uint32_t *ethernetFutex = ethernetFutex = interrupt_futex_get(STATIC_SEALED_VALUE(ethernetInterruptCapability));
+```
+
+The `ethernetFutex` pointer is now a read-only capability (attempting to store through it will trap) that contains a number that is incremented every time that an interrupt fires.
```
rmn30, line 143:
```suggestion
The `ethernetFutex` pointer is now a read-only capability (attempting to store through it will trap) that contains a number that is incremented every time the ethernet interrupt fires.
```

rmn30 (approved):

docs/WritingADeviceDriver.md
davidchisnall:
```suggestion
Now that you have a pointer to a `volatile` object representing the device's MMIO region, you can access its control registers directly.
```

docs/WritingADeviceDriver.md
davidchisnall:
```suggestion
As with the MMIO capabilities, sealed objects appear in compartment reports.
For example, the above macro expands to this in the final report:
```

docs/WritingADeviceDriver.md
davidchisnall: Yup, there are a bunch of auditing things that consume JSON, we should pick one and provide helper functions for it.

PR #90, opened 2023-06-30
revoker: paint ranges backwards

nwf-msr:


sdk/core/allocator/revoker.h
```diff
@@ -183,26 +183,40 @@ namespace Revocation
...
 			else
 			{
-				shadowCap[baseWordIx] &= ~maskLo;
 				shadowCap[topWordIx] &= ~maskHi;
 				midWord = 0;
 			}
 
-			for (size_t shadowWordIx = baseWordIx + 1; shadowWordIx < topWordIx;
-			     shadowWordIx++)
+			for (size_t shadowWordIx = topWordIx - 1; baseWordIx < shadowWordIx;
```
rmn30, line 206: Could `topWordIdx` ever be `0`?

rmn30 (commented):
Looks like a good catch. Just one edge case to think about. I guess it wasn't an issue before because `baseWordIdx` would never by `MAX_SIZE_T`.

davidchisnall:
This looks sensible, can you rerun the allocator benchmarks and see if it introduces a regression?  In particular, the extra branch at the end makes me wonder if we should do a little bit of refactoring.  The compiler will probably unswitch the loop, so we actually have two completely different cases for the two functions but it might be worth making sure and manually unswitching it.

sdk/core/allocator/revoker.h
rmn30: I think we are safe from underflow because it could only happen if `topWordIdx == baseWordIdx` in which case we would already have returned in the case above. Maybe a coment to that effect would be good?

rmn30 (approved):

nwf-msr:
For the release build of the test suite, there is a very small regression relative to 7d71cc28c (of 0.03%) that I've been unable to make better by playing around with unswitching:

```
Before:
Test runner: Allocator finished in 4669423 cycles
Test runner: All tests finished in 5072910 cycles

After:
Test runner: Allocator finished in 4671080 cycles
Test runner: All tests finished in 5074569 cycles
```

If we're OK with 64 bytes of code bloat (changing `.text` from 98098 to 98162; a gain of 0.06%), we can template the method on `fill` and mark it as `__always_inline` (since it is called exactly once in each `fill` disposition) to more than compensate for the induced regression (a speedup of 0.2% relative to 7d71cc28c):
```
Test runner: Allocator finished in 4659207 cycles
Test runner: All tests finished in 5062460 cycles
```

davidchisnall (approved):

davidchisnall:
64 bytes is probably not too bad.  I think we'll save a bit more than that once we enable the bit-manipulation extension.

PR #94, opened 2023-07-06
Add Hongyan's FreeRTOS compat header.

davidchisnall:
We probably don't want to merge this as-is, but it's a useful starting point.


davidchisnall:
Superseded by #151

PR #96, opened 2023-07-10
Enforce stack-size limit.

### LINKED ISSUE #58: Document and enforce stack size limits
Stacks larger than about 8 KiB are problematic for compartment switches because the switcher cannot guarantee that the remainder stack has a representable length.
In particular this [`csetboundsexact`](https://github.com/microsoft/cheriot-rtos/blob/e24f37fafa05d8eb23f81c76ee3b997c6b9415b3/sdk/core/switcher/entry.S#L184) might trap.
Given that stack bounds must be 16-byte aligned and the capability precision is 9 bits stacks up to `16 * 511 = 8176` bytes should be fine.

We should document this limit and possibly enforce it in the xmake and / or linker.

davidchisnall:
The compartment switcher will not behave correctly if subsetting a stack results in something unrepresentable.  We can avoid this by limiting stacks to the maximum representable size for 16-byte alignment.

Eventually this can be removed if we make the switcher round the shared bit of the stack down to avoid overlap.  At the moment, it's not a serious limitation because our stacks are very rarely over 1 KiB and the limit is about 8 KiB.

Fixes #58

### DIFF
--- a/sdk/xmake.lua
+++ b/sdk/xmake.lua
@@ -405,10 +405,19 @@ rule("firmware")
 			"\n\t\tbootTStack = .;" ..
 			"\n\t\t. += " .. loader_trusted_stack_size .. ";" ..
 			"\n\t}\n"
+		-- Stacks must be less than this size or truncating them in compartment
+		-- switch will encounter precision errors.
+		local stack_size_limit = 8176
 		for i, thread in ipairs(threads) do
 			thread.mangled_entry_point = string.format("__export_%s__Z%d%sv", thread.compartment, string.len(thread.entry_point), thread.entry_point)
 			thread.thread_id = i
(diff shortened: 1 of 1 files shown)


sdk/xmake.lua
```diff
@@ -405,11 +405,20 @@ rule("firmware")
 			"\n\t\tbootTStack = .;" ..
 			"\n\t\t. += " .. loader_trusted_stack_size .. ";" ..
 			"\n\t}\n"
+		-- Stacks must be less than this size or truncating them in compartment
+		-- switch will encounter precision errors.
```
rmn30, line 409:
```suggestion
		-- switch will encounter precision errors.
```
```suggestion
		-- switch may encounter precision errors.
```

rmn30 (approved):
Looks fine, but would like to see a test with the maximum stack size.

PR #98, opened 2023-07-11
Token unsealing via libcall

nwf-msr:
This is a (hopefully correct) subset of the overly-ambitious #91.  It gives us a fast-path for unsealing tokens without a full cross-compartment call and enables the use of static ocaps without needing the allocator compartment.  However, the allocator, if present, is in the same trust domain as the token unsealer: they share the same hardware sealing capability, and, as David points out, the allocator can, in principle, change the software type of any ocap to which it has access.  Future work might correct the deficiencies of #91, but this PR is, hopefully, a useful, self-contained subset.

This PR introduces a minor performance delta for the test suite, which is a somewhat unrealistically heavy user of the malloc API.  Using b4b6204e as a baseline, the release test suite compiles to a `104810` bytes in `.text` and `2872` bytes in `.data` and runs in...
```
Test runner: Allocator finished in 4779117 cycles
Test runner: All tests finished in 5182625 cycles
```
After this PR, that's `105504` bytes in `.text` (+ `694` or `0.7%`) and `2896` in `.data` (+ `24` or `0.8%`) and runs...
```
Test runner: Allocator finished in 4803030 cycles
Test runner: All tests finished in 5203722 cycles
```
That is, the release test suite takes `21097` additional cycles to run, `23913` of which are in the allocator test (that is, the rest of the test suite is, on average, _faster_), for a `0.4%` slowdown.


sdk/core/loader/types.h
```diff
@@ -523,6 +534,11 @@ namespace loader
 				return compartments[Allocator];
 			}
 
+			[[nodiscard]] const PrivilegedCompartment &token_server() const
```
davidchisnall, line 537: Maybe add a comment here explaining what it is?  I don't really like the token_server name anymore: it's not a token server, it's a helper library that does a fast path for some allocator functions.  We possibly want to move `heap_can_free` and `heap_quota_remaining` it as well at some point, since these are pure functions that depend only on access to the unseal capability.

sdk/core/loader/boot.cc
```diff
@@ -1066,6 +1080,7 @@ extern "C" SchedulerEntryInfo loader_entry_point(const ImgHdr &imgHdr,
 	switcherKey.bounds()      = 1;
 	setSealingKey(imgHdr.scheduler(), Scheduler);
 	setSealingKey(imgHdr.allocator(), Allocator);
+	setSealingKey(imgHdr.token_server(), Allocator);
```
davidchisnall, line 1083: Perhaps remove PermitSeal from this one?

sdk/core/token_server/token_unseal.S
```diff
@@ -0,0 +1,101 @@
+#include <compartment-macros-asm.S>
+#include <cheri-builtins.h>
+#include "../allocator/token.h"
+
+/*
+ * An in-assembler, libcall-able implementation of
+ *
+ * [[cheri::interrupt_state(disabled)]] void *__cheri_compartment("token_server")
```
davidchisnall, line 8: This should be `__cheri_libcall` not `__cheri_compartment`.

davidchisnall (commented):
A few comments inline but the main thing is that the 'token server' name is now quite misleading (it was accurate in the previous PR).  I think this should be called 'allocator fast paths' or similar.

sdk/core/loader/types.h
nwf-msr: How about `ocap_fast_paths`?

davidchisnall:
The code size changes look pretty large.  Can you compare the allocator code + data to the allocator and the new library's code + data?  A 694 byte increase looks huge.  The allocator is currently 7712 bytes of code and 48 bytes of data, so that's a 9% increase in size.

sdk/core/loader/types.h
davidchisnall: I don't think we use ocap anywhere (also, see the coding style guides about not using abbreviations).  `token_fast_paths` would work for its current use, `allocator_fast_paths` seems better since it is explicitly fast paths for functionality provided by the allocator compartment.

nwf-msr:
Sure, drilling down a bit... relative to 69a1baf9a025ae126a71c3e1a852fd08acd02fb6, I see:
```diff
--- tmp/pre.txt 2023-07-12 17:08:43.633510649 +0100
+++ tmp/post.txt        2023-07-12 17:08:02.943460021 +0100
@@ -1,8 +1,9 @@
    text           data     bss     dec     hex filename
   10208            232      64   10504    2908 tests/build.sail/cheriot/cheriot/release/allocator_test.compartment
      30             32       0      62      3e tests/build.sail/cheriot/cheriot/release/ccompile_test.compartment
-   8348            256      40    8644    21c4 tests/build.sail/cheriot/cheriot/release/cherimcu.allocator.compartment
+   8212            260      40    8512    2140 tests/build.sail/cheriot/cheriot/release/cherimcu.allocator.compartment
     462             32      20     514     202 tests/build.sail/cheriot/cheriot/release/cherimcu.software_revoker.compartment
+     94             48       0     142      8e tests/build.sail/cheriot/cheriot/release/cheriot.ocap_fast_paths.compartment
    2513             84       0    2597     a25 tests/build.sail/cheriot/cheriot/release/compartment_calls_inner.compartment
    1259             56       0    1315     523 tests/build.sail/cheriot/cheriot/release/compartment_calls_inner_with_handler.compartment
    2165            128       0    2293     8f5 tests/build.sail/cheriot/cheriot/release/compartment_calls_test.compartment
@@ -10,14 +11,14 @@
    3556             64       0    3620     e24 tests/build.sail/cheriot/cheriot/release/crash_recovery_outer.compartment
    5000             72       4    5076    13d4 tests/build.sail/cheriot/cheriot/release/crash_recovery_test.compartment
    5729            124       4    5857    16e1 tests/build.sail/cheriot/cheriot/release/futex_test.compartment
-   5689            232      88    6009    1779 tests/build.sail/cheriot/cheriot/release/locks_test.compartment
-  11227            296     120   11643    2d7b tests/build.sail/cheriot/cheriot/release/multiwaiter_test.compartment
+   5649            232      88    5969    1751 tests/build.sail/cheriot/cheriot/release/locks_test.compartment
+  11157            296     120   11573    2d35 tests/build.sail/cheriot/cheriot/release/multiwaiter_test.compartment
    3417            144      16    3577     df9 tests/build.sail/cheriot/cheriot/release/queue_test.compartment
    2091             56      16    2163     873 tests/build.sail/cheriot/cheriot/release/stack_integrity_thread.compartment
    3485            128       1    3614     e1e tests/build.sail/cheriot/cheriot/release/stack_test.compartment
-   4390            100       0    4490    118a tests/build.sail/cheriot/cheriot/release/static_sealing_inner.compartment
+   4380            100       0    4480    1180 tests/build.sail/cheriot/cheriot/release/static_sealing_inner.compartment
      70             68       0     138      8a tests/build.sail/cheriot/cheriot/release/static_sealing_test.compartment
    3496           1480      32    5008    1390 tests/build.sail/cheriot/cheriot/release/test_runner.compartment
    9108            316     632   10056    2748 tests/build.sail/cheriot/cheriot/release/test-suite.scheduler.compartment
    1408             32     288    1728     6c0 tests/build.sail/cheriot/cheriot/release/thread_pool.compartment
-   5175            208      40    5423    152f tests/build.sail/cheriot/cheriot/release/thread_pool_test.compartment
+   5165            208      40    5413    1525 tests/build.sail/cheriot/cheriot/release/thread_pool_test.compartment
- 106194    2872     701  109767   1acc7 tests/build.sail/cheriot/cheriot/release/test-suite
+ 107016    2896     701  110613   1b015 tests/build.sail/cheriot/cheriot/release/test-suite
```

Ah ha:

```diff
- 66 .loader_code                                  00001f50 80018dd0 TEXT
- 67 .loader_data                                  00000048 8001ad20 DATA
+ 67 .loader_code                                  00002300 80018d20 TEXT
+ 68 .loader_data                                  00000048 8001b020 DATA
```

That's a gain of 944 bytes in `.loader_code`.  Drilling down further:
* "Introduce notion of privileged libraries" makes `.loader_code` `0x21e0` bytes long, an increase of 656 bytes.
* "Allow setSealingKey to refine permissions" makes it `0x2260` (128 more bytes)
* "Move token unseal to new privileged library" makes it `0x2300` (160 more bytes)

All of those seem significantly larger than I'd expect, and I'm not sure what's going on there.

sdk/core/loader/types.h
nwf-msr: I didn't want to use `allocator_fast_paths` because _static_ software-defined capabilities aren't really tied to the allocator, despite the current implementation.

davidchisnall:
I'm happy to clean up the loader later.  I think the extra 'oh, but is it a library?' checks are probably bloating the code a bit, we can probably reduce it.  At around 8.5 KiB, the loader is still small in comparison to most heaps, so is fine.

sdk/core/loader/types.h
davidchisnall: I think those should go in a separate `scheduler_fast_paths` privileged library.

davidchisnall: Oh, and I eventually want to rewrite the `event_*` functions on top of futex and move them out of the scheduler, so most of those won't actually need a fast path.  The scheduler will grow  `FUTEX_WAIT_BITSET` and `FUTEX_WAKE_BITSET` analogues and then the event channel things can be a separate (optional) library.

scripts/run_clang_tidy_format.sh
```diff
@@ -37,12 +37,14 @@ echo Sources: ${SOURCES}
 rm -f tidy-*.fail
 
 # sh syntax is -c "string" [name [args ...]], so "tidy" here is the name and not included in "$@"
-echo ${HEADERS} ${SOURCES} | xargs -P${PARALLEL_JOBS} -n5 sh -c "${CLANG_TIDY} -export-fixes=\$(mktemp -p. tidy.fail-XXXX) \$@" tidy
-if [ $(find . -maxdepth 1 -name 'tidy.fail-*' -size +0 | wc -l) -gt 0 ] ; then
-	# clang-tidy put non-empty output in one of the tidy-*.fail files
-	cat tidy.fail-*
-	exit 1
-fi
+[ -n "${CHERIOT_SKIP_CLANG_TIDY-}" ] || {
```
davidchisnall, line 40: What is this about?  I think it might be accidentally included with this PR?

sdk/core/loader/boot.cc
```diff
@@ -1066,6 +1084,11 @@ extern "C" SchedulerEntryInfo loader_entry_point(const ImgHdr &imgHdr,
 	switcherKey.bounds()      = 1;
 	setSealingKey(imgHdr.scheduler(), Scheduler);
 	setSealingKey(imgHdr.allocator(), Allocator);
+	setSealingKey(imgHdr.object_fast_paths(),
```
davidchisnall, line 1087: I still don't really like this name.  Object is very broad.  This is the allocator fast paths library: it holds the allocator's key and adding anything that is not logically part of the allocator to it could break the security.

sdk/core/loader/types.h
```diff
@@ -492,6 +501,8 @@ namespace loader
 				Scheduler,
 				/// The memory allocator
 				Allocator,
+				/// The token server
+				TokenServer,
```
davidchisnall, line 505: This probably wants to be consistently named with the accessor.  I'd suggest `AllocatorLibrary`.

sdk/xmake.lua
```diff
@@ -167,6 +169,17 @@ rule("cherimcu.privileged-compartment")
 		target:add("defines", "CHERIOT_AVOID_CAPRELOCS")
 	end)
 
+rule("cherimcu.privileged-library")
+	add_deps("cherimcu.component")
+	on_load(function (target)
+		-- Mark this target as a CHERIoT compartment.
+		target:set("cherimcu.type", "compartment")
+		target:set("extension", ".compartment")
```
davidchisnall, line 177: Shouldn't these be library?

sdk/xmake.lua
```diff
@@ -188,6 +201,13 @@ target("cherimcu.allocator")
 		target:set('cherimcu.debug-name', "allocator")
 	end)
 
+target("cheriot.object_fast_paths")
+	add_rules("cherimcu.privileged-library", "cherimcu.component-debug")
+	add_files(path.join(coredir, "object_fast_paths/token_unseal.S"))
+	on_load(function (target)
+		target:set("cherimcu.compartment", "object_fast_paths")
```
davidchisnall, line 208: Is this needed?  It's used to override the compartment flag value, but we shouldn't be passing that at all when building a library.

scripts/run_clang_tidy_format.sh
nwf-msr: It's the last commit (presently 6a19ddd); as the message says, `clang-tidy` takes forever but `clang-format` is fast, so it's nice to be able to run just the latter.  I can yank that commit out of this PR, though.

sdk/core/loader/boot.cc
nwf-msr: But static ocaps can be used without even having the allocator, once we have something like this PR to mediate unsealing.  In that sense it happens that the allocator holds _the system ocap key_.

sdk/xmake.lua
nwf-msr: IIRC, it was easier in the loader to treat these like compartments that exported libcalls than it was to treat them like libraries that needed `.rodata` capabilities installed.  I might be / have been wrong.

scripts/run_clang_tidy_format.sh
davidchisnall: I'm not sure how useful it is.  If you just want to run clang-format quickly, either, running clang-format-diff for the whole diff or just `clang-format -i` is easier.

sdk/xmake.lua
nwf-msr: Changing `cherimcu.type` is easy enough, but changing the extension to `".library"`, which I wouldn't expect to make much of a difference, instead prompts the last link to say
```
ld.lld: error: Import table for compartment 'allocator' refers to address 0x800025DC, which is not a valid export
```
(and similarly for other compartments).  `llvm-objdump` of the intermediate object says
```
error: address range table at offset 0x0 has a premature terminator entry at offset 0x10
```

At a quick glance through LLVM, I didn't see why that would be happening.  Suggestions?

davidchisnall: I don't really like the reimplementation of the logic in the library rule, this rule should just adopt that one.  If you just delegate to the library rule, does it do the right thing?

davidchisnall (approved):

davidchisnall:
Don't forget to squash before merging.

sdk/xmake.lua
davidchisnall: This just confuses me.  There is nothing anywhere in the toolchain that should depend on the file extension (except the linker script for the final firmware - did you update it there as well?).

nwf-msr: Yes, see https://github.com/microsoft/cheriot-rtos/pull/98/commits/058c8d79501380bf3e751c0fc2e00ea43976beb9#diff-b2382c519e7cfb672fc11e41afb5a0ac747603b3502b42163a3bf902e974efc3, but that might be wrong?

davidchisnall: All of the bits that say `cheriot.token_library.compartment` need to say `cheriot.token_library.library`.

davidchisnall: The error message is a bit surprising, but understandable, I think.  If a section is referred to but isn't mentioned in the linker script then it gets shoved at the end.  With the wrong name, these things will be in the wrong place and so the firmware will 'link' except that the bit that calculates the report does some additional sanity checking and tells you it's nonsense.

PR #100, opened 2023-07-12
initial software driver for dma

NazerkeT:
[Not a candidate for actually merging yet!]

Thanks for the comments Wes! I have already checked them, however will still create a PR here for David with older version before changing/coding other pieces.

Thanks,
Best wishes

### DIFF
--- a/sdk/include/platform/ibex/platform-dma.hh
+++ b/sdk/include/platform/ibex/platform-dma.hh
@@ -1,0 +1,118 @@
+#pragma once
+
+#define MALLOC_QUOTA 0x100000
+#define TEST_NAME "Allocator"
+#include "tests.hh"
+
+#include <cheri.hh>
+#include <compartment-macros.h>
+#include <cstdint>
+#include <stdint.h>
+#include <utils.hh>
+
+#include <thread.h>
+#include <thread_pool.h>
+
+using thread_pool::async;
+DECLARE_AND_DEFINE_ALLOCATOR_CAPABILITY(dmaDriverHeap, 8144);
+
+namespace Ibex {
+    class GenericDMA {
+        private:
+
+        Timeout noWait{0};
+        SObjStruct *dmaHeap;
+
+        struct StreamingInterface
+        {
+            uint32_t control;
+            uint32_t status;
+            uint32_t address;
+            uint32_t lengthInBytes;
+        };
+
+        struct DMAInterface
+        {
+            /**
+             *  these two streaming interfaces are for 
+             *  [0] memory to target and 
+             *  [1] target to memory directions.
+             *  target here is any I/O device, peripheral or an accelerator
+             */
+            volatile StreamingInterface *streamingInterface[2];
+        };
+
+        volatile DMAInterface *dmaInterface;
+
+        public:
+
+        void init()
+        {
+            dmaInterface->streamingInterface[0] = MMIO_CAPABILITY(StreamingInterface, dma_to_target);
+            dmaInterface->streamingInterface[1] = MMIO_CAPABILITY(StreamingInterface, dma_to_memory);
+        }
+
+        uint32_t read_status(bool index)
+        {   
+            /**
+             *  this statement returns the less signifance bit 
+             *  to show the halted status
+             */
+            return dmaInterface->streamingInterface[index]->status & 0x1;
+        }
+
+        void write_conf(bool index, uint32_t *address, uint32_t lengthInBytes)
+        {
+            /**
+             *  filling address and length fields
+             */
+            dmaInterface->streamingInterface[index]->address = *address;
+            dmaInterface->streamingInterface[index]->lengthInBytes = lengthInBytes;
+
+            /**
+             *  create a heap capability for dma below. 
+             *  todo: we assume that this heap capability can serve 
+             *  for different compartments, irrespective of 
+             *  the origin because this is a driver code?
+             */
+            dmaHeap = STATIC_SEALED_VALUE(dmaDriverHeap);
+
+            /**
+             *  claim the allocated memory 
+             */
+
+            int  claimCount = 0;
+
+            auto claim      = [&]() {
+                size_t claimSize = heap_claim(dmaHeap, address);
+                claimCount++;
+                TEST(claimSize == lengthInBytes,
+                        "{}-byte allocation claimed as {} bytes (claim number {})",
+                        lengthInBytes,
+                        claimSize,
+                        claimCount);
+            };
+
+            claim();
+        }
+
+        void start_dma(bool index)
+        {
+            /**
+             *  setting a start bit
+             */
+            dmaInterface->streamingInterface[index]->control = 0x1;
+        }
+
+        void reset_dma(bool index, uint32_t *address)
+        {
+            /**
+             *  setting a reset bit
+             */
+            dmaInterface->streamingInterface[index]->control = 0x4;
+
+            int ret = heap_free(dmaHeap, address);
+        }
+
+    };
+} // namespace Ibex


sdk/include/platform/ibex/platform-dma.hh
```diff
@@ -0,0 +1,118 @@
+#pragma once
+
+#define MALLOC_QUOTA 0x100000
```
davidchisnall, line 3: This should be in the .cc file for the compartment..

sdk/include/platform/ibex/platform-dma.hh
```diff
@@ -0,0 +1,118 @@
...
+#define TEST_NAME "Allocator"
+#include "tests.hh"
```
davidchisnall, line 5: These are specific to the test suite and should be used only in tests.

sdk/include/platform/ibex/platform-dma.hh
```diff
@@ -0,0 +1,118 @@
...
+
+#include <cheri.hh>
+#include <compartment-macros.h>
+#include <cstdint>
+#include <stdint.h>
+#include <utils.hh>
+
+#include <thread.h>
+#include <thread_pool.h>
```
davidchisnall, line 14: These look unused.

sdk/include/platform/ibex/platform-dma.hh
```diff
@@ -0,0 +1,118 @@
...
+
+using thread_pool::async;
+DECLARE_AND_DEFINE_ALLOCATOR_CAPABILITY(dmaDriverHeap, 8144);
+
+namespace Ibex {
+    class GenericDMA {
+        private:
+
+        Timeout noWait{0};
```
davidchisnall, line 23: You probably don't want this here, allocating a new one on the stack when you use it will be faster than having a global.  It is also not sensibly reentrant - the field in `Timeout` that tracks the amount that's consumed is not updated atomically and so it will end up being nonsense if you use the same timeout from multiple threads.

sdk/include/platform/ibex/platform-dma.hh
```diff
@@ -0,0 +1,118 @@
...
+        SObjStruct *dmaHeap;
+
+        struct StreamingInterface
+        {
+            uint32_t control;
+            uint32_t status;
+            uint32_t address;
+            uint32_t lengthInBytes;
```
davidchisnall, line 31: It would be good to have doc comments for these to explain what they are (but thanks for giving them human-friendly names!).

sdk/include/platform/ibex/platform-dma.hh
```diff
@@ -0,0 +1,118 @@
...
+        struct DMAInterface
+        {
+            /**
+             *  these two streaming interfaces are for 
+             *  [0] memory to target and 
+             *  [1] target to memory directions.
+             *  target here is any I/O device, peripheral or an accelerator
+             */
+            volatile StreamingInterface *streamingInterface[2];
+        };
```
davidchisnall, line 43: I would expect this to be something like:

```c++
struct DMAInterface
{
	StreamingInterface memoryToDevice;
	StreamingInterface deviceToMemory;
};
```

You can then initialise a `volatile DMAInterface *` with `MMIO_CAPABILITY`.

sdk/include/platform/ibex/platform-dma.hh
```diff
@@ -0,0 +1,118 @@
...
+
+        volatile DMAInterface *dmaInterface;
```
davidchisnall, line 45: Caching this will be slower than using the `MMIO_CAPABILITY` directly.  If you want to save some typing, wrapping the `MMIO_CAPABILITY` use in an `__always_inline` function will be better.

sdk/include/platform/ibex/platform-dma.hh
```diff
@@ -0,0 +1,118 @@
...
+
+        public:
+
+        void init()
+        {
+            dmaInterface->streamingInterface[0] = MMIO_CAPABILITY(StreamingInterface, dma_to_target);
+            dmaInterface->streamingInterface[1] = MMIO_CAPABILITY(StreamingInterface, dma_to_memory);
+        }
```
davidchisnall, line 53: This is redundant if you don't cache the pointer.

sdk/include/platform/ibex/platform-dma.hh
```diff
@@ -0,0 +1,118 @@
...
+             */
+            dmaHeap = STATIC_SEALED_VALUE(dmaDriverHeap);
+
+            /**
+             *  claim the allocated memory 
+             */
+
+            int  claimCount = 0;
+
+            auto claim      = [&]() {
```
davidchisnall, line 86: This is called once, so probably doesn't need to be a lambda, but it might want to be if you are claiming the source and destination.

sdk/include/platform/ibex/platform-dma.hh
```diff
@@ -0,0 +1,118 @@
...
+        uint32_t read_status(bool index)
+        {   
+            /**
+             *  this statement returns the less signifance bit 
+             *  to show the halted status
+             */
+            return dmaInterface->streamingInterface[index]->status & 0x1;
+        }
+
+        void write_conf(bool index, uint32_t *address, uint32_t lengthInBytes)
```
davidchisnall, line 64: `bool` seems a strange type for a thing called `index`.  What does it mean?  Looking at the code, I think this is the device to configure - it would probably be better to expose an interface that configures both sides at the same time, or you have fun race conditions if two things both want to start a DMA at the same time.

sdk/include/platform/ibex/platform-dma.hh
```diff
@@ -0,0 +1,118 @@
...
+            /**
+             *  claim the allocated memory 
+             */
+
+            int  claimCount = 0;
+
+            auto claim      = [&]() {
+                size_t claimSize = heap_claim(dmaHeap, address);
+                claimCount++;
+                TEST(claimSize == lengthInBytes,
```
davidchisnall, line 89: This shouldn't be a `TEST`, it should be a check with correct error handling on failure.  In particular, if you have already claimed one then you must call `heap_free` on it when the other claim fails.  You will also need to keep both pointers so that you can call `heap_free` to release your claim at the end of a transaction.

sdk/include/platform/ibex/platform-dma.hh
```diff
@@ -0,0 +1,118 @@
...
+
+        void reset_dma(bool index, uint32_t *address)
+        {
+            /**
+             *  setting a reset bit
+             */
+            dmaInterface->streamingInterface[index]->control = 0x4;
+
+            int ret = heap_free(dmaHeap, address);
+        }
```
davidchisnall, line 115: These would probably make more sense as methods on `StreamingInterface`.

davidchisnall (commented):
Thanks for posting this for early review.  I've left some comments inline, I hope they're helpful.

sdk/include/platform/ibex/platform-dma.hh
```diff
@@ -0,0 +1,140 @@
...
+             *  - it can be either memory buffer or MMAP-ed I/O device (i.e. any peripheral or accelerator)
+             */
+            uint32_t targetAddress;
+            /**
+             * Below is the amount of data IN BYTES that DMA should transfer 
+             * between the source and a target 
+             */
+            uint32_t lengthInBytes;
+            /**
+             * Below is the amount of strides IN 4 BYTE WORDS for a source address.
```
nwf-msr, line 44: - Strides probably want to be in bytes.  I suspect that interacts poorly with the desire to swap on 2- or 4-byte granularities.
- Zero stride probably should mean zero stride, or there should, at least, be a way of spelling zero-stride, for use with MMIO FIFOs.

sdk/include/platform/ibex/platform-dma.hh
```diff
@@ -0,0 +1,140 @@
...
+            dmaInterface = MMIO_CAPABILITY(DMAInterface, dma);
+        }
+
+        uint32_t read_status()
+        {   
+            /**
+             *  this statement returns the less signifance bit 
+             *  to show the halted status
+             */
+            return dmaInterface->status & 0x1;
```
nwf-msr, line 76: Could you repurpose `control`'s bottom bit rather than needing a separate `status` word?

sdk/include/platform/ibex/platform-dma.hh
```diff
@@ -0,0 +1,140 @@
...
+             */
+            dmaInterface->control = 0x1;
+        }
+
+        void reset_dma()
+        {
+            /**
+             *  setting a reset bit, which is bit 3
+             */
+            dmaInterface->control = 0x8;
```
nwf-msr, line 132: What are the semantics of reset?  Transfer aborted and all DMA control registers zeroed?

sdk/core/dma/dma.h
```diff
@@ -0,0 +1,74 @@
...
+		/**
+		 * Initialise a DMA device.
+		 */
+
+		void init()
+		{
+			PlatformDMA<WordT, TCMBaseAddr>::init();
+		}
+
+        void write_conf(uint32_t sourceAddress, uint32_t targetAddress, uint32_t lengthInBytes)
```
nwf-msr, line 47:
```suggestion
        void write_conf(uint32_t *sourceAddress, uint32_t *targetAddress, uint32_t lengthInBytes)
```

sdk/core/dma/dma.h
```diff
@@ -0,0 +1,74 @@
...
+		{
+			device.write_conf()
+			} -> std::same_as<void>;
+		{
+			device.start_dma()
+			} -> std::same_as<void>;
+        {
+			device.reset_dma()
+			} -> std::same_as<void>;
+	};
```
nwf-msr, line 29:
```suggestion
	concept IsDmaDevice = requires(T device, uint32_t *sourceAddress, uint32_t *destinationAddress, size_t length)
	{
		{device.init()};
		{
			device.write_conf(sourceAddress, destinationAddress, length)
			} -> std::same_as<void>;
		{
			device.start_dma()
			} -> std::same_as<void>;
        {
			device.reset_dma(sourceAddress, destinationAddress)
			} -> std::same_as<void>;
	};
```

sdk/core/dma/dma_compartment.cc
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     *  create a heap capability for dma below. 
+     */
+    dmaHeap = STATIC_SEALED_VALUE(dmaDriverHeap);
+
+    /**
+     *  claim the memory 
+     */
+    
+    size_t sourceClaimSize = heap_claim(dmaHeap, sourceAddress);
```
nwf-msr, line 33: I wonder if this is sufficient grounds to want a `heap_claim(SObj heapCapability, void*... objects)` that amortizes the compartment call overhead.  Similarly for `heap_free()`.  @DavidChisnall?

sdk/core/dma/dma_compartment.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include <cdefs.h>
+#include <stdint.h>
+
+/**
+ * A function below claims the source and target addresses of the DMA interface.
+ * While, DMA is in progress, these addresses should not be freed
+ */
+__cheri_compartment("dma") int claim_dma(uint32_t *sourceAddress,
```
nwf-msr, line 10: I think these (claim and free) functions are not what you mean to be exporting from the compartment.  I was expecting something something more like `write_conf` (though probably a fusion of `write_conf`, `write_strides`, and `start`), I think.

To support multiple users of DMA, we probably want that "`memmove_but_with_dma`" call, whatever it's named, to take a `Timeout` for waiting on the DMA engine to become available and for it to use a futex to communicate with the engine's IRQ handler?  But that's perhaps a discussion for later.

nwf-msr (commented):
This looks like a good start.  Good job onboarding enough `xmake` and `__cheri_` annotations to be dangerous. :)

Sorry if any of my comments &| questions below are misguided as a result of having spent too much time with software.

sdk/core/dma/dma.h
```diff
@@ -0,0 +1,50 @@
+#pragma once
```
davidchisnall, line 1: This probably doesn’t belong in the core directory, that’s for core parts of the RTOS, whereas this is optional.

sdk/core/dma/dma.h
```diff
@@ -0,0 +1,50 @@
...
+
+#include "dma_compartment.h"
+#include "platform/ibex/platform-dma.hh"
+#include <concepts>
+#include <cstddef>
+#include <stdint.h>
+#include <utils.hh>
+
+#if __has_include(<platform-dma.hh>)
```
davidchisnall, line 10: Why is this optionally included?

sdk/core/dma/dma.h
```diff
@@ -0,0 +1,50 @@
...
+        {
+			device.reset_dma()
+			} -> std::same_as<void>;
+	};
+
+    template<typename PlatformDMA>
+
+	requires IsDmaDevice<PlatformDMA>
+
+	class GenericDMA : public PlatformDMA
```
davidchisnall, line 35: I’m a bit confused. Is this a generic DMA interface or a wrapper around a non-CHERI DMA interface in a compartment?

sdk/core/dma/dma_compartment.cc
```diff
@@ -0,0 +1,109 @@
...
+#include <thread_pool.h>
+
+#if __has_include(<platform-dma.hh>)
+#	include <platform-dma.hh>
+#endif
+
+// Import some useful things from the CHERI namespace.
+using namespace CHERI;
+
+using thread_pool::async;
```
davidchisnall, line 17: This shouldn’t be using the thread pool.

sdk/core/dma/dma_compartment.cc
```diff
@@ -0,0 +1,109 @@
...
+/**
+ *  creating a global heap below.
+ *  todo: we assume that this heap capability can serve
```
davidchisnall, line 20: Yes, exactly. This bounds claims for the duration of a DMA and so has a large quota but one that does not contribute much to total memory consumption.

sdk/core/dma/dma_compartment.cc
```diff
@@ -0,0 +1,109 @@
...
+using namespace CHERI;
+
+using thread_pool::async;
+/**
+ *  creating a global heap below.
+ *  todo: we assume that this heap capability can serve 
+ *  for different compartments, irrespective of 
+ *  the origin because this is a driver code?
+*/
+DECLARE_AND_DEFINE_ALLOCATOR_CAPABILITY(dmaDriverHeap, 8144);
```
davidchisnall, line 24: If you’re using a named heap capability, why did you provide a large quota to the malloc capability?

sdk/core/dma/dma_compartment.cc
```diff
@@ -0,0 +1,109 @@
...
+
+SObjStruct *dmaHeap;
```
davidchisnall, line 26: What is this?

sdk/core/dma/dma_compartment.cc
```diff
@@ -0,0 +1,109 @@
...
+    if (!check_pointer<PermissionSet{Permission::Load, Permission::Global}>(sourceAddress, lengthInBytes) ||
+           !check_pointer<PermissionSet{Permission::Store, Permission::Global}>(targetAddress, lengthInBytes) )
+    {
+        return -1;
+    }
+
+    /**
+     *  create a heap capability for dma below. 
+     */
+    dmaHeap = STATIC_SEALED_VALUE(dmaDriverHeap);
```
davidchisnall, line 58: Why are you copying this from one global to another?

sdk/core/dma/dma_compartment.cc
```diff
@@ -0,0 +1,109 @@
...
+
+int dma_compartment(uint32_t *sourceAddress, uint32_t *targetAddress, uint32_t lengthInBytes,
+                        uint32_t sourceStrides, uint32_t targetStrides, uint32_t byteSwapAmount)
+{    
+    /**
+     *  return if sufficient permissions are not present 
+     *  and if not long enough 
+     */
+
+    if (!check_pointer<PermissionSet{Permission::Load, Permission::Global}>(sourceAddress, lengthInBytes) ||
```
davidchisnall, line 49: Check pointer should happen after claim. The tag can be cleared by a concurrent free if they happen the other way around.

sdk/core/dma/dma_compartment.cc
davidchisnall: That’s a good idea. In particular, the call could atomically succeed or fail, which simplifies error handling for the caller. I think we’d want to bound the maximum number of objects that can be claimed in a single operation and we can’t used variadic an across compartment calls so it would have to pass an array and a length.

sdk/core/dma/dma_compartment.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#include <cdefs.h>
+#include <stdint.h>
+
+/**
+ * A function below claims the source and target addresses of the DMA interface.
+ * While, DMA is in progress, these addresses should not be freed
```
davidchisnall, line 8: More explicitly: they will be claimed by the DMA compartment and so the memory will not be freed.

sdk/include/platform/ibex/platform-dma.hh
```diff
@@ -0,0 +1,140 @@
...
+             */
+            return device().status & 0x1;
+        }
+
+        void write_conf(uint32_t *sourceAddress, uint32_t *targetAddress, uint32_t lengthInBytes)
+        {
+            /**
+             *  Setting source and target addresses, and length fields
+             */
+            device().sourceAddress = *sourceAddress;
```
davidchisnall, line 82: This line looks wrong. You’re dereferencing the pointer and using whatever it happens to point to as the address. I think you want to be using the address from the capability. Same comment for the next line.

sdk/include/platform/ibex/platform-dma.hh
```diff
@@ -0,0 +1,140 @@
...
+        void write_strides(uint32_t sourceStrides, uint32_t targetStrides)
+        {
+            /**
+             *  Setting source and target strides
+             */
+            device().sourceStrides = sourceStrides;
+            device().targetStrides = targetStrides;
+        }
+
+        int byte_swap_en(uint32_t swapAmount)
```
davidchisnall, line 97: Is en short for something? English?

sdk/include/platform/ibex/platform-dma.hh
```diff
@@ -0,0 +1,140 @@
+#pragma once
+
```
davidchisnall, line 2: This file looks like it’s describing the interface to the specific DMA controller for use in the compartment. I think this should be local to the DMA compartment, not a public header. The public header is the one that describes the public interface for other code to talk to any DMA engine, including this one: it should have the concept for a DMA engine and the interface to the compartment.

examples/10.dma_test/xmake.lua
```diff
@@ -0,0 +1,39 @@
...
+    add_files("dma_test.cc")
+
+-- Firmware image for the example.
+firmware("dma_test")
+    add_deps("crt", "freestanding", "atomic_fixed")
+    add_deps("dma", "dma_app")
+    -- add_deps("dma_app")
+    on_load(function(target)
+        target:values_set("board", "$(board)")
+        target:values_set("threads", {
```
davidchisnall, line 30: You don’t have any worker threads for the thread pool, so your async blocks will never run.

sdk/core/dma/dma_compartment.cc
```diff
@@ -0,0 +1,196 @@
...
+     */	
+	LockGuard g{dmaOwnershipLock};
+
+    /**
+     *  If dma is already launched, we need to check for the interrupt status.
+	 *  No need for validity and permissions checks though for the scheduler 
+	 *  once futex is created   
+     */
+    
+	const uint32_t *dmaFutex = interrupt_futex_get(STATIC_SEALED_VALUE(dmaInterruptCapability));
```
davidchisnall, line 66: This should be static

sdk/core/dma/dma_compartment.cc
```diff
@@ -0,0 +1,196 @@
...
+    uint32_t currentInterruptCounter = *dmaFutex;
+    
+    /** 
+     * sleep a bit, so that dmaIsLaunched can be flipped successfully
```
davidchisnall, line 70: What does this comment mean?

sdk/core/dma/dma_compartment.cc
```diff
@@ -0,0 +1,196 @@
...
+     *  cleared by every DMA operation,
+     *  that is why we are explicitely resetting 
+     *  them at this exit function.
+     */
+
+    /**
+     *  Flip the dmaIsRunnning value to show that launch is finished,
+     *  in case finishing thread wakes from the interrupt first 
+     */
+    dmaIsLaunched = 0;
```
davidchisnall, line 173: This needs to hold the lock or a concurrent call here will corrupt state with a launch.

sdk/core/dma/dma_compartment.cc
NazerkeT: I forgot to delete this, will delete it 👌

nwf-msr:
I swear I hit "cancel review" to get rid of some stale notes.  Sorry for the noise.

PR #102, opened 2023-07-31
Add `heap_free_all`.

davidchisnall:
This API takes a heap capability and frees *all* memory that can be freed by that capability.

The current implementation holds the heap lock for the entire heap walk. This may not be ideal but it's somewhat complex to avoid.  We would need to ensure that the chunk that we are currently inspecting is not coalesced when we yield.

### DIFF
--- a/sdk/core/allocator/main.cc
+++ b/sdk/core/allocator/main.cc
@@ -435,9 +435,8 @@ namespace
 			 */
 			Iterator &operator++()
 			{
-				Claim *next = **this;
-				encodedNextPointer =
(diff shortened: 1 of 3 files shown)


davidchisnall:
I want to leave the long-held lock to a future commit.  We should have a baseline 'make it work' before we add a 'make it fast' on top.  I'd also quite like to see some timing results with a realistic heap because I don't think the lock is actually held very long (for things with no claims, we scan them very quickly, the worst case is if we have a lot of objects with multiple claims on them).

PR #108, opened 2023-08-28
Add support for edge-triggered interrupts.

davidchisnall:
These are automatically completed in the scheduler and so do not require explicit completion.

### DIFF
--- a/sdk/core/scheduler/plic.h
+++ b/sdk/core/scheduler/plic.h
@@ -71,6 +71,12 @@ namespace sched
 			 * The priority for this interrupt.
 			 */
 			uint32_t priority;
+			/**
+			 * True if this interrupt is edge triggered, false otherwise.  Edge
+			 * triggered interrupts are automatically acknowledged, level
+			 * triggered interrupts must be explicitly acknowledged.
+			 */
+			bool isEdgeTriggered;
 		};
 
 		/**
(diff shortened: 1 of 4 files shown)


sdk/xmake.lua
```diff
@@ -361,7 +361,11 @@ rule("firmware")
 			local interruptConfiguration = "CHERIOT_INTERRUPT_CONFIGURATION="
 			for _, interrupt in ipairs(board.interrupts) do
 				interruptNames = interruptNames .. interrupt.name .. "=" .. math.floor(interrupt.number) .. ", "
-				interruptConfiguration = interruptConfiguration .. "{" .. math.floor(interrupt.number) .. "," .. math.floor(interrupt.priority) .. "},"
+				interruptConfiguration = interruptConfiguration .. "{"
+					.. math.floor(interrupt.number) .. ","
+					.. math.floor(interrupt.priority) .. ","
+					.. (interrupt.edge_triggered and "true" or "false")
+					.. "},"
```
nwf, line 368: While you're in the neighborhood, could you add a comment that these are initializers for `sched::InterruptController::Interrupt` structures?

sdk/core/scheduler/plic.h
```diff
@@ -139,6 +145,10 @@ namespace sched
 			{
 				if (ConfiguredInterrupts[i].number == uint32_t(source))
 				{
+					if (ConfiguredInterrupts[i].isEdgeTriggered)
```
nwf, line 148: Is this really the right place to do this?  `futex_word_for_source` is called by `interrupt_futex_get`, which is (AIUI) meant to be used in initialization, not necessarily when an interrupt has happened.  Perhaps this belongs in the other caller of `futex_word_for_source`, namely `do_external_interrupt`?

Separately, because we're in IRQ context, and so running with IRQs disabled, I don't think there's concern about the order in which completing the interrupt and incrementing the futex word happens, but it'd be good to document that fact (assuming it is, in fact, a fact).

nwf-msr (approved):
LGTM

PR #109, opened 2023-08-30
Add a script for building all of the configurations of a benchmark.

davidchisnall:
This takes a base board.json, modifies it to include the different variations, then builds the specified benchmark, once for each version.

Also fixes some bugs for configurations we haven't tested.

### DIFF
--- a/scripts/build_benchmark_configs.sh
+++ b/scripts/build_benchmark_configs.sh
@@ -1,0 +1,105 @@
+#!/bin/sh
+
+# Helper script to build multiple configurations of a benchmark.  Patches the
+# board description files to enable and disable different combinations of
+# hardware features.  Also patches include paths to keep track of the original
+# locations.
+
+set -e
+
+SCRIPTPATH=$(dirname $(readlink -f $0))
+
+BOARD=$(realpath $1)
+BENCHMARK=$(realpath $2)
+SDK=$(realpath $3)
+BOARDPATH=$(dirname ${BOARD})
+BOARDNAME=$(basename -s .json ${BOARD})
+
+if [ ! -f "${BOARD}" ] ; then
+	echo "First argument must be the path of a board description file.  '${BOARD}' is not a valid file"
+	exit 1
+fi
+if [ ! -f "${BENCHMARK}/xmake.lua" ] ; then
+	echo "Second argument must be the path of a benchmark file.  '${BENCHMARK}' does not contain an xmake.lua file"
+	exit 1
+fi
+if [ ! -d "${SDK}" ] ; then
+	echo "Third argument must be the path of the sdk.  '${SDK}' does not exist"
+	exit 1
+fi
+if [ ! -f "${SDK}/bin/clang" ] ; then
+	echo "Third argument must be the path of the sdk.  '${SDK}/bin/clang' does not exist"
+	exit 1
+fi
+
+# Set up a temporary directory to work in.
+DIR=$(mktemp -d benchmark.XXX) || exit 1
+echo Working in ${DIR}...
+DIR=$(pwd)/${DIR}
+cd  ${DIR}
+
+
+# Create a directory for the board files
+
+mkdir boards
+cd boards
+
+
+# Our JSON files are slightly extended JSON, parse them with xmake's JSON
+# parser and spit them out as standard JSON so that jq can consume them.
+cp ${BOARD} basic.json
+cat <<EOF > xmake.lua
+target('fake')
+	on_load(function (target)
+		import("core.base.json")
+		local board = json.loadfile(path.join(os.scriptdir(), 'basic.json'))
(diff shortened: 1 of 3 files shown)


davidchisnall:
@microsoft-github-policy-service agree company="Microsoft"

scripts/build_benchmark_configs.sh
```diff
@@ -0,0 +1,103 @@
...
+
+# Helper script to build multiple configurations of a benchmark.  Patches the
+# board description files to enable and disable different combinations of
+# hardware features.  Also patches include paths to keep track of the original
+# locations.
+
+set -eo pipefail
+
+SCRIPTPATH=$(dirname $(readlink -f $0))
+SDKPATH=$(realpath ${SCRIPTPATH}/../sdk)
```
nwf-msr, line 11: Stale?

scripts/build_benchmark_configs.sh
```diff
@@ -0,0 +1,103 @@
+#!/bin/sh
+
+# Helper script to build multiple configurations of a benchmark.  Patches the
+# board description files to enable and disable different combinations of
+# hardware features.  Also patches include paths to keep track of the original
+# locations.
+
+set -eo pipefail
```
nwf-msr, line 8: Debian's `dash`, which it uses as `/bin/sh`, apparently doesn't like `-o pipefail`, which seems surprising.

scripts/build_benchmark_configs.sh
```diff
@@ -0,0 +1,103 @@
...
+jq 'del(.revoker) | .stack_high_water_mark=false | .defines[.defines| length]|= .+"CHERIOT_FAKE_REVOKER"' < standard.json > ${BOARDNAME}-fake-revoker.json
+rm standard.json
+
+CONFIGS="${BOARDNAME}-software-revoker ${BOARDNAME}-software-revoker-shwm ${BOARDNAME}-hardware-revoker ${BOARDNAME}-hardware-revoker-shwm ${BOARDNAME}-no-revoker ${BOARDNAME}-fake-revoker"
+
+# Patch the include directories so that relative paths become absolute relative
+# to the board file.
+for I in ${CONFIGS}; do
+	# Fix up includes if they expect to be relative to the board file path
+	jq ".driver_includes = (.driver_includes | map_values(. = if . | startswith(\"/\") then . else \"${BOARDPATH}/\" + . end))" < $I.json > $I.json.fixed
```
nwf-msr, line 82: I think this also needs to probe for starting with `$`, as in `${sdk}`, which will be expanded by `xmake`?
```suggestion
	jq ".driver_includes = (.driver_includes | map_values(. = if . | startswith(\"/\") or startswith(\"\$\") then . else \"${BOARDPATH}/\" + . end))" < $I.json > $I.json.fixed
```

scripts/build_benchmark_configs.sh
davidchisnall: Yes, that’s left over from when I was doing ${sdk} substitution, but I don’t think we need it so I removed the complicated jq expression.

scripts/build_benchmark_configs.sh
davidchisnall: No, that will do the wrong thing. If include paths are sdk-relative then we need to do sdk substitution. I had an if then else thing in the jq that substituted ${SDKDIR} for the literal ${sdk} in that case. If you have a file that uses sdk that you can test with, can you try adding that back?

scripts/build_benchmark_configs.sh
davidchisnall: Hmm, not sure if it’s better to require bash or skip the pipe fail here. It should be fine with zsh and, I think, the FreeBSD POSIX shell.

scripts/build_benchmark_configs.sh
davidchisnall: You need to have it inside the map_values bit.  This line is trying to replace the array, not the elements of the array.

davidchisnall: Yes, that's correct.  I was misremembering why I didn't do this.  We only do ${sdk} substitution for include files and treat relative paths as relative to the board source location.  For the simulator, we also allow ${board}.  That's the thing that needs updating when we move the board file.  Ho hum.

scripts/build_benchmark_configs.sh
nwf-msr: Are we actually doing anything in a pipeline in this script?  The only `|`s I see are either part of `||`s or inside `jq` expressions.

davidchisnall: Hmm, maybe not, so just -eo should be fine.

nwf-msr: Just `-e`, I think?  And`-u` is also generally a good plan, but I haven't looked for where we might rely on not-that at the moment.

nwf-msr (approved):

PR #111, opened 2023-09-25
Enable hardware stack zeroing.

davidchisnall:
This replaces the software stack zeroing path with offload to a hardware engine, when available.  The hardware zeroing pipeline is controlled by the ZTOP SCR (currently assumed to be number 27).  This zeroes from the top of the capability in ZTOP to the bottom.

We assume that short-lived cross-compartment calls or returns from deeply nested calls will result in ranges where each is a subset of another, so we try to coalesce.  We could probably do better aggregating overlapping but not subset ranges.


sdk/core/switcher/entry.S
```diff
@@ -347,11 +398,20 @@ exception_entry_asm:
...
 	csrr               t1, CSR_MSHWM
 	csw                t1, TrustedStack_offset_mshwm(csp)
 	csrr               t1, CSR_MSHWMB
 	csw                t1, TrustedStack_offset_mshwmb(csp)
+#endif
+#ifdef CHERIOT_HAS_ZTOP
+	// Stop zeroing and capture the current zeroing value.  Note: cspecialr is
+	// encoded as cspecialrw and so we can't just use cnull as the source.  We
+	// probably could use the current ct1 value, since it's guaranteed to be
+	// untagged at this point in the code.
```
nwf-msr, line 411: Can we leave the zeroizer running until the `cspecialw` below (making that a load of the new thing, a `cspecialrw`, and a store of the old thing)?  We'd need to keep at least access to `TrustedStack_offset_ztop(csp)` (for *this* `csp`) around, but we have ample registers in which to do so.

sdk/core/switcher/entry.S
```diff
@@ -347,11 +398,20 @@ exception_entry_asm:
...
 	csrr               t1, mstatus
 	csw                t1, TrustedStack_offset_mstatus(csp)
-#ifdef CONFIG_MSHWM
+#ifdef CHERIOT_HAS_MSHWM
 	csrr               t1, CSR_MSHWM
 	csw                t1, TrustedStack_offset_mshwm(csp)
 	csrr               t1, CSR_MSHWMB
 	csw                t1, TrustedStack_offset_mshwmb(csp)
+#endif
+#ifdef CHERIOT_HAS_ZTOP
```
nwf-msr, line 407: You may as well put this `ifdef` underneath the one for `CHERIOT_HAS_MSHWM`, too, since the feature is dependent.

sdk/core/switcher/entry.S
davidchisnall: I pondered that, but I don't like the idea of slowing down the scheduler, since that impacts interrupt latency (switching to a high priority thread on interrupt will be slower if the interrupted thread is doing zeroing).  We can't leave it in a register while we invoke the scheduler and so we'd have to get it back, which might break the code that handles the double fault case.

nwf-msr: That's a good point about the slowdown, but I'd still kind of like to measure it, at least to know.

But yeah, juggling things around the (untrusted) scheduler is more of a challenge, yes. :(

sdk/core/switcher/entry.S
```diff
@@ -126,6 +135,46 @@ switcher_scheduler_entry_csp:
...
+	// relative to ztop will trap.
+	cgettag            \top, c\scratch
+	beqz               \top, 2f
+	// If we've already reached the bottom then do nothing.
+	// NOTE: This can be deleted if ZTOP's tag is cleared when zeroing finishes.
+	cgetbase           \top, c\scratch
+	beq                \top, \scratch, 2f
+	// If the current zeroing range is a subset of range that we're about to
+	// zero, don't bother waiting, just zero from the start.
+	ctestsubset        \top, c\base, c\scratch
```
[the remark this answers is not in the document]
davidchisnall: We could do it with checks on the top and bottom, since we already have checks that CSP is sane (correct permissions, correctly aligned top and bottom) on entry into the compartment switcher.  I'm not sure if we could do that without adding another temporary register, which would require some small tweaks to the users of this macro.

sdk/core/switcher/entry.S
```diff
@@ -18,6 +18,15 @@
...
 #define CSR_MSHWMB 0xbc2
 
+/**
+ * The special capability register for the fast zeroing control.
+ * Writing a capability here starts zeroing from the address of this register
+ * down to the base.
+ *
+ * TODO: This is a made up number, it might need to be a different one.
+ */
+#define SCR_ZTOP 0x1c
```
nwf-msr, line 28: 0x1c is `mtcc`.  Ibex is using `0x1b` for this.  That was an exciting thing to discover in testing.

nwf-msr:
This runs the `compartment-call-benchmark` test, generating
```
#board     stack size      full call       call    return
ibex.json 0x100   186     127     59
ibex.json 0x200   186     127     59
ibex.json 0x400   188     129     59
ibex.json 0x800   188     129     59
ibex.json 0x1000  188     129     59
```
We'll see if it survives the full test harness!

nwf-msr:
It does survive the full test harness, huzzah!  Reported cycle numbers for 452d404df65c27a58fc8b3566777aeaa7000880a + this PR, without and with the zeroizer support enabled:

| Test                             | Without  | With     |
|----------------------------------|----------|----------|
| Static sealing                   |   449592 |   449425 |
| Crash recovery                   |  7438235 |  7438011 |
| Compartment calls                |  1378192 |  1378099 |
| Stack exhaustion in the switcher |  2129788 |  2162198 |
| Thread pool                      |   867080 |   865635 |
| Global constructors              |    85663 |    85522 |
| Queue                            |  6053459 |  6052851 |
| Futex                            |  2943148 |  2943848 |
| Locks                            |  5001834 |  5001183 |
| Multiwaiter                      |  4623924 |  4623760 |
| Allocator                        |  7858154 |  7688631 |
| All tests                        | 40007035 | 39868625 |

sdk/core/switcher/entry.S
```diff
@@ -126,6 +133,46 @@ switcher_scheduler_entry_csp:
  * and integer register.  All three registers are clobbered.
  */
 .macro zero_stack base top scratch
+#ifdef CHERIOT_HAS_ZTOP
+	// Derive the capability to the range that should be zeroed.
+	// This is stored in base, leaving top as a second scratch register to use
+	csub               \top, c\top, c\base
+	csetboundsexact    c\base, c\base, \top
```
nwf-msr, line 140: As per discussions out of band, this should probably be
```suggestion
	csetbounds    c\base, c\base, \top
	cincoffset    c\base, c\base, \top
```

- We need to set the address to the top of the region to be zeroed, since the state machine counts down.
- Using inexact bounds allows us to tolerate large (> 2KiB) stacks without risk, and...
  - Because the stack is guaranteed to be a superset representable region, we'll never try to grow beyond it
  - Because the state machine counts down from the (precise) address, it won't zeroize parts of the stack in use
  - It might zero a little bit past the HWM, but that's fine.

sdk/core/switcher/trusted-stack-assembly.h
```diff
@@ -19,28 +19,39 @@ EXPORT_ASSEMBLY_OFFSET(TrustedStack, c12, 12 * 8)
...
 EXPORT_ASSEMBLY_OFFSET(TrustedStack, c14, 14 * 8)
 EXPORT_ASSEMBLY_OFFSET(TrustedStack, c15, 15 * 8)
-EXPORT_ASSEMBLY_OFFSET(TrustedStack, mstatus, 16 * 8)
-EXPORT_ASSEMBLY_OFFSET(TrustedStack, mcause, (16 * 8) + 4)
-#ifdef CONFIG_MSHWM
-EXPORT_ASSEMBLY_OFFSET(TrustedStack, mshwm, 17 * 8)
-EXPORT_ASSEMBLY_OFFSET(TrustedStack, mshwmb, (17 * 8) + 4)
+#ifdef CHERIOT_HAS_ZTOP
+EXPORT_ASSEMBLY_OFFSET(TrustedStack, ztop, 16 * 8)
+#	define TSTACK_HAS_ZTOP 1
```
nwf-msr, line 24: This is _extremely_ minor, but the differing "types" of `CHERIOT_HAS_ZTOP` and `TSTACK_HAS_ZTOP` is bugging me.  Maybe `TSTACK_ZTOP_WORDS` or such?

davidchisnall:
Test failure is interesting.  The SHWM is reporting a lot more stack usage in this mode, but it shouldn't.  There may be a hardware bug in the SHWM's interaction with the zeroing engine.

PR #115, opened 2023-10-08
Fix CI to use the correct devcontainer

davidchisnall:


davidchisnall:
And this taught me something about Docker (pushing aarch64 and x86_64 images with the same tag did not, in fact, give me x86_64 and aarch64 images.  It gave me whichever one took longer to build and ended up being pushed second).  This is now fixed and both the AArch64 and x86_64 images are present, so CI doesn't fail trying to pull the wrong image.

rmn30 (approved):

PR #117, opened 2023-10-16
Update LLVM build instructions

marnovandermaas:
First is a minor fix of directory name from `llvm` to `cheriot-llvm`.
Second is some fixes to the `cmake` command:
- `${LLVM_PATH}` needs to be followed by `llvm` to find the correct folder.
- Remove some extra spaces in the command argumetns.
- Remove `LLVM_ENABLE_LLD` to resolve the following error on GCC
platforms:
```
CMake Error at cmake/modules/HandleLLVMOptions.cmake:298 (message):
  Host compiler does not support '-fuse-ld=lld'
```

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -118,14 +118,14 @@ Next create a directory for the build and enter it:
 
 ```sh
 $ mkdir -p builds/cheriot-llvm
-$ cd builds/llvm
+$ cd builds/cheriot-llvm
 ```
 
 This can be inside the source checkout but you may prefer to build somewhere else, for example on a different filesystem.
 Next, use `cmake` to configure the build:
 
 ```sh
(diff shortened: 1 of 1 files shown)


davidchisnall:
The first change looks good, but I don't like forcing a specific compiler (it should be fine to build LLVM with whatever your system compiler is).

marnovandermaas:
> I don't like forcing a specific compiler (it should be fine to build LLVM with whatever your system compiler is).

Thanks, I've removed the forcing of the compiler.

davidchisnall:
We might want to remove `LLVM_ENABLE_LLD` as well.  I'm not sure if this works on GCC platforms (I don't know if it checks whether LLD is supported by the compiler, or just unconditionally passes the flag).

marnovandermaas:
> We might want to remove `LLVM_ENABLE_LLD` as well. I'm not sure if this works on GCC platforms (I don't know if it checks whether LLD is supported by the compiler, or just unconditionally passes the flag).

Good point, that was the source of the cmake error I reported earlier.

PR #120, opened 2023-10-17
Work around a bug in Ibex exception priorities.

davidchisnall:
We exit the examples with success if they reach the end of the main thread.  We detect that exit by checking that the return address register is hitting a tag fault and we are at the top of the stack.

With Ibex, we relax this and ignore the exception cause.


sdk/include/fail-simulator-on-error.h
```diff
@@ -58,7 +58,11 @@ compartment_error_handler(ErrorState *frame, size_t mcause, size_t mtval)
 		  frame->get_register_value<CHERI::RegisterNumber::CRA>()};
 		if (registerNumber == CHERI::RegisterNumber::CRA &&
 		    returnCapability.address() == 0 &&
+		// Ibex currently has a bug with exception priorities and so
+		// reports these with the wrong error code.
```
rmn30, line 62:
```suggestion
		// reports these with the wrong error code, therefore relax the condition a bit.
```

rmn30 (commented):
Looks good except maybe explain the workaround a bit more.
In fact since we know Ibex currently gives a bounds exception maybe we should check for that and then we'll remember to remove the workaround when it is fixed?

rmn30 (approved):

PR #121, opened 2023-10-23
Update microvium

davidchisnall:
The newer version in npm has some bytecode format changes and so this is needed for people to be able to run the JavaScript examples.

The newer version introduced a snprintf dependency for float -> string and int -> string conversion.  We compile in no-float mode, so this is a huge dependency for something quite simple.  I have upstreamed a hook to allow users to provide a 32-bit integer to float conversion routine and added an implementation of it in microvium_port.h.

While poking things upstream, I also added hooks to set bounds and permissions on pointers.  These are now used when getting a pointer to a string or a buffer from Microvium, which means that the JavaScript heap is now fully memory safe when used from C++ code.


examples/09.javascript/bytecode.inc
```diff
@@ -1 +1 @@
...
...0x01,0x00,0x01,0x00,0x01,0x00,0x51,0x00,0x6d,0x00,0x61,0x00,0x45,0x00,0x3d,0x00,0x05,0x40,0x70,0x75,0x73,0x68,0x00,0x00,0x04,0x40,0x6c,0x6f,0x67,0x00,0x00,0x00,0x02,0x60,0x00,0x00,0x0e,0x40,0x48,0x65,0x6c,0x6c,0x6f,0x2c,0x20,0x57,0x6f,0x72,0x6c,0x64,0x21,0x00,0x07,0x40,0x61,0x72,0x72,0x61,0x79,0x5b,0x00,0x00,0x00,0x00,0x05,0x40,0x5d,0x20,0x3d,0x20,0x00,0x00,0x04,0x50,0x31,0x30,0x30,0x88,0x1d,0x00,0x6b,0x12,0x6f,0x67,0x01,0x60,0x00,0x00,0x06,0x50,0x01,0x01,0x7e,0x00,0xa1,0x06,0xa0,0x10,0x89,0x02,0x00,0xe0,0x70,0x02,0x11,0x60,0x89,0x01,0x00,0x08,0x6c,0x10,0x8c,0x01,0x00,0x67,0x89,0x01,0x00,0x12,0x12,0x12,0x6f,0x67,0x10,0x07,0x6c,0x10,0xa1,0x67,0x76,0xdd,0x09,0x50,0x01,0x01,0x01,0x89,0x00,0x00,0x10,0x88,0x45,0x00,0x6b,0xa1,0x88,0x51,0x00,0x78,0x82,0x89,0x03,0x00,0x01,0x78,0x01,0xa1,0x06,0xa0,0x06,0x10,0xa1,0x67,0x10,0x89,0x02,0x00,0xe0,0x70,0x02,0x01,0x60,0x01,0x89,0x00,0x00,0x10,0x88,0x45,0x00,0x6b,0xa1,0x88,0x61,0x00,0x13,0x88,0x6d,0x00,0x16,0x16,0x6b,0x78,0x85,0x10,0x07,0x6c,0x10,0xa1,0x67,0x76,0xd9,0x00,0x0c,0x00,0x03,0x00,0x17,0x00,0x85,0x00,0x19,0x00,0x02,0x00,0x19,0x00,0x01,0x00,0x08,0xc0,0x05,0x00,0x05,0x00,0x3d,0x00,0x75,0x00,0x08,0xc0,0x05,0x00,0x05,0x00,0x45,0x00,0x4d,0x00}
```
nwf-msr, line 1: I'm just going to trust that that's correct, since I am not a cylon.

sdk/include/microvium/microvium_port.h
```diff
@@ -165,13 +166,13 @@ fixes and improvement from the original github or npm repository.
  *
  * Offset may be negative
  */
-#define MVM_LONG_PTR_ADD(p, s) ((MVM_LONG_PTR_TYPE)((uint8_t*)p + (intptr_t)s))
+#define MVM_LONG_PTR_ADD(p, s) ((MVM_LONG_PTR_TYPE)((uint8_t *)p + (intptr_t)s))
```
nwf-msr, line 169: Maybe, while you're here,
```suggestion
#define MVM_LONG_PTR_ADD(p, s) ((MVM_LONG_PTR_TYPE)((uint8_t *)p + (ptrdiff_t)s))
```

sdk/include/microvium/microvium_port.h
```diff
@@ -291,10 +294,70 @@ static uint16_t crc16(MVM_LONG_PTR_TYPE lp, uint16_t size) {
  * The `context` passed to these macros is whatever value that the host passes
  * to `mvm_restore`. It can be any value that fits in a pointer.
  */
-#define MVM_CONTEXTUAL_MALLOC(size, context) ({ Timeout t = {0, 0}; heap_allocate(&t, context, size); })
+#define MVM_CONTEXTUAL_MALLOC(size, context)                                   \
+	({                                                                         \
+		Timeout t = {0, 0};                                                    \
```
nwf-msr, line 299: I realize this hunk is just reformatting, but...

Should this tie into the "gas counter" below somehow?  I know that counts in units of VM instructions, and maybe we want to consider `malloc` to be O(1) within a VM instruction, so perhaps not, but OTOH maybe we want to tell `malloc` to give up after `gas_left * rough_time_per_vm_instr` time?

davidchisnall: I don't think there's a good way of doing that.  This is currently a non-blocking malloc and the VM will stop if it can't allocate.  We might want to provide a hook that allows users to specify a timeout to use (possibly make the context a pointer to a timeout and an allocation capability, so they can set a per-invocation cumulative timeout).

examples/09.javascript/bytecode.inc
davidchisnall: > I am not a cylon

I will defer fixing that for a later PR.

nwf-msr (approved):

PR #122, opened 2023-10-23
Add first version of a compartmentalisation exercise.

davidchisnall:
This is intended for the DSbD all hands and (hopefully) for other people to try.

The firmware example provided in this uses Microvium to simulate an attacker who has managed to acquire arbitrary-code execution via a code reuse attack.  It loads JavaScript bytecode via the UART and executes it, with access to some FFI functions that allow reading and modifying a set of capabilities.

Users then have three steps to work through:

 1. Prevent the attacker from crashing the system.
 2. Prevent the attacker from exhausting memory.
 3. Prevent the attacker from exfiltrating a secret.


exercises/01.compartmentalisation/js.cc
```diff
@@ -0,0 +1,568 @@
...
+					return c - 'a' + 10;
+				case 'A' ... 'F':
+					return c - 'A' + 10;
+			}
+			return 0;
+		};
+
+		// Read values in the form generated by microvium's hex output: a
+		// series of hex bytes in braces.
+		skipUntil('{');
```
nwf-msr, line 501: Maybe we should be warning if we discard something other than whitespace?

exercises/01.compartmentalisation/js.cc
```diff
@@ -0,0 +1,568 @@
...
+	};
+
+	/// Constant for the say_hello function exposed to C++->JavaScript FFI
+	static constexpr mvm_VMExportID ExportRun = 1234;
+
+	/**
+	 * Type used for the set of capabilities that JavaScript has complete
+	 * control over.
+	 */
+	using VMRegisterState = std::array<void *, 8>;
```
nwf-msr, line 93: This name is a little confusing since there are two layers of virtual machines here: JS and the in-JS capability machine.

exercises/01.compartmentalisation/js.cc
```diff
@@ -0,0 +1,568 @@
...
+
+	/**
+	 * Address of the `VMRegisterState` on the stack.  We store this on the
```
nwf-msr, line 96:
```suggestion
	 * Address of the `VMRegisterState` on the stack.  That structure is on the
```

exercises/01.compartmentalisation/cheri.js
```diff
@@ -0,0 +1,98 @@
+// FFI Imports
+// Each function imported from the host environment needs to be assigned to a
+// global like this and identified by a constant that the resolver in the C/C++
+// code will understand.
+
+
+/**
+ * Log function, writes all arguments to the UART.
+ */
+export const print = vmImport(1);
```
nwf-msr, line 10: Document that these constants are elements of the `Exports` `enum` down in `js.cc`?

exercises/01.compartmentalisation/js.cc
```diff
@@ -0,0 +1,568 @@
...
+			case 10:
+			{
+				void *pcc;
+				asm("auipcc %0, 0\n" : "=C"(pcc));
+				return pcc;
+			}
+		}
+	}
+
+	/**
```
nwf-msr, line 158: Could all of this, through `call_export` be factored out into a microvium glue header or even pushed upstream?  It feels a little generic and/or out of place here.

exercises/01.compartmentalisation/README.md
```diff
@@ -0,0 +1,222 @@
...
+enum ErrorRecoveryBehaviour
+compartment_error_handler(struct ErrorState *frame,
+                          size_t             mcause,
+                          size_t             mtval);
+```
+
+If you implement this function in a compartment, it will be invoked when any synchronous fault (such as a CHERI exception) occurs.
+This is a flexible mechanism that lets you rewrite the register file and resume, or just give up and return to the compartment that invoked this one.
+
+Implement an error handler that frees all memory associated with the JavaScript compartment.
```
nwf-msr, line 204:
```suggestion
Implement an error handler that does not attempt to recover from the error but instead frees all memory associated with the JavaScript compartment and bails to the caller.
```

nwf-msr (commented):
I really like this.  Some nits here and there, but this is great.

exercises/01.compartmentalisation/js.cc
```diff
@@ -0,0 +1,568 @@
...
+{
+	/// The secret value.
+	int32_t secret;
+
+	/**
+	 * Set the secret the first time.
+	 *
+	 * This blocks until there is a character available on the UART.  This is
+	 * *not* a good way of getting entropy, but it's the only one that we have
+	 * in the simulator.
```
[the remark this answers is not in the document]
davidchisnall: I think most real silicon versions will have a separate hardware entropy source.

exercises/01.compartmentalisation/js.cc
davidchisnall: If this were a real program, maybe, but it's already a bit overcomplicated since it's meant to be used only with the load script, which does the right thing.

davidchisnall:
Addressed all review comments except in places where I disagree over phrasing.

I've reordered the exercises so that they fix the most critical issues first:

1. Fix the confidentiality problem.
2. Fix the system-wide availability problem.
3. Fix the per-compartment availability problem.

The first exercise is now very simple because it's compartmentalising along an existing boundary.  This is the version with a solution:

```
$ git diff --stat
 exercises/01.compartmentalisation/secret.h  | 4 ++--
 exercises/01.compartmentalisation/xmake.lua | 4 +++-
 2 files changed, 5 insertions(+), 3 deletions(-)
```

Hopefully this shows that adding compartments is nice and easy!

nwf-msr (approved):

PR #126, opened 2023-10-25
Add Sail patch instruction

marnovandermaas:
Add missing command that I found in the Sail README: https://github.com/microsoft/cheriot-sail/blob/19ad1e982c0f876bbb3f21f39b3cee6935278adc/README.md?plain=1#L28


rmn30 (approved):
Thanks!

PR #127, opened 2023-10-27
Exercise 1 readme tweaks

rmn30:
Attempt to clarify some things, fix a typo and incomplete sentence. Also add instructions to configure before attempting to build.

### DIFF
--- a/exercises/01.compartmentalisation/README.md
+++ b/exercises/01.compartmentalisation/README.md
@@ -28,7 +28,7 @@ First we need to configure the `xmake` build to run on the Ibex simulator and bu
 If you are using the [dev container](../../docs/GettingStarted.md) you should run:
 
 ```
-xmake config --sdk=/cheriot-tools/ --board=ibex-safe-simulator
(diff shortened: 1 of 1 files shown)


exercises/01.compartmentalisation/README.md
```diff
@@ -112,7 +119,7 @@ Exercise 1: Confidentiality
 ---------------------------
 
 First, we want to protect the secret.
-For the this exercise, we want to move the 
+For the this exercise, we want to move it into a separate comparment from the JavaScript interpreter: we want to minimise the attack surface of the compartment that handles the secret, so we definitely don't want it to share with the potentially vulnerable interpreter.
```
davidchisnall, line 122: I don't think we want to talk about attack surface here.  We are doing this to isolate the secret, not to reduce the attack surface.  We could talk about blast radius, but that's probably too much detail.

rmn30: Better? I just wanted to fix the incomplete sentence.

davidchisnall (approved):

PR #132, opened 2023-11-02
Run apt-update before installing things.

davidchisnall:
If you don't, apt will use stale indexes and then fail in the installation because the package repo state doesn't match what it expects.


nwf-msr (approved):
Ah, apt.  Yup.

Issue #133, opened 2023-11-03
Would be nice to support hazard pointer like claims on heap objects


nwf-msr:
We have a notion of claims of heap objects, but they're fairly heavy, requiring a full compartment call to either acquire or drop.  We'd like a lighter-weight option for a small number of objects per thread: hazard pointers understood by the heap.

Hazard pointers originate in the world of lock/wait-free data structures, where pointers have a four-stage lifecycle: unpublished, public, withdrawn, and freed.  Hazard pointers mediate the transition from withdrawn to freed, deferring it until there are no more hazardous, local copies of once-public pointers.  Because everyone involved is presumed cooperative, hazard lists need only be scanned once: a request to free memory must follow that pointer being withdrawn from the shared state, and so it cannot become newly hazardous.

The story for potentially uncooperative threads is slightly different, as it may be impossible to ensure that a pointer is completely withdrawn from the shared state (and is not cached somewhere without being declared hazardous) when it comes time to `free()` it.  This is, after all, why we have sweeping revocation!  As such, our story needs to be a little different.  In particular, we need a ratchet mechanism to advance an object through these states:
1. allocated and hazard-able (and possibly hazardous)
2. allocated but no longer hazard-able, with N hazards registered
3. ...
4. allocated but no longer hazard-able, with 1 hazard registered
5. No longer hazardous, and so free (well, quarantined)

Additionally, hazards are intended to be ephemeral.  As such, we do not expect them to survive cross-compartment calls.  (And anyway, a callee cannot assume that its caller has declared any hazards appropriately, and so must do so itself; if we ensured that hazards persisted across calls, there would be redundant registrations.)

Here is a design sketch.  This is going to need to be a fair bit of carefully crafted assembler but I believe it's entirely possible.

There is a Hazard Registry maintained on behalf of the allocator by a privileged _library_.  A Hazard Registry Entry (or just "Hazard") is an optional pair of a capability to a (sub)object and a taint flag.  The `None` branch of the option is indicated by a `null` cap.  Hazard entries are, we imagine, stored in static sealed objects, with handle type `SealedHeapHazard`, each of which has room for some number of hazards.  To avoid linked-list traversal, we imagine that these are collected in a dedicated section, forming a "linker set".

The privileged library exposes a single call to clients (and more to the allocator itself):
`void *heap_hazard_register(SealedHeapHazards *hh, size_t ix, void *p)`, with interrupts deferred,
  - unseals `hh`,
[...]

PR #136, opened 2023-11-19
Add support for importing MMIO capabilities reduced permissions.

davidchisnall:
This is most useful for read-only access, but the mechanism should be sufficiently general to permit other uses (in particular, once we have CHERI-aware devices, it may be useful to restrict the ability to store capabilities and the ability to load mutable capabilities from devices).

### DIFF
--- a/sdk/core/loader/boot.S
+++ b/sdk/core/loader/boot.S
@@ -47,7 +47,10 @@ start:
 	la_abs			a3, __compart_headers_end
 	sub				a3, a3, a1
 	csetaddr		ca1, ca2, a1 // ca2 still has the G root.
-	csetboundsexact	ca1, ca1, a3
+	// FIXME: This should be a set bounds exact, but we currently don't have a
+	// 'pad to capability alignment' command in the linker script and it needs
+	// to span two sections.
+	csetbounds		ca1, ca1, a3
 	// Set up $cra to be the loader's C++ entry point.
 	// We are safe to clobber $cra here because this is the root function on
 	// the call stack.
--- a/sdk/core/loader/boot.cc
+++ b/sdk/core/loader/boot.cc
@@ -405,36 +405,40 @@ namespace
 	 */
 	void *find_export_target(const ImgHdr &image,
 	                         const auto   &sourceCompartment,
-	                         ptraddr_t     target,
-	                         size_t        size)
+	                         ImportEntry  &entry)
 	{
 		// Build an MMIO capability.
 		auto buildMMIO = [&]() {
-			Debug::log("Building mmio capability {} + {}", target, size);
+			Debug::log("Building mmio capability {} + {} (permissions: {})",
+			           entry.address,
+			           entry.size(),
+			           entry.permissions());
 			if constexpr (std::is_same_v<
 			                std::remove_cvref_t<decltype(sourceCompartment)>,
 			                ImgHdr::PrivilegedCompartment>)
 			{
 				if (&sourceCompartment == &image.allocator())
 				{
-					if (target == LA_ABS(__export_mem_heap) &&
-					    (size == (LA_ABS(__export_mem_heap_end) -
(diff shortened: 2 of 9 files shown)


davidchisnall:
This is generally useful and has been on my to-do list for a while, but is specifically useful as a building block for #133.

sdk/core/loader/boot.cc
```diff
@@ -405,36 +405,41 @@ namespace
...
+	                         ImportEntry  &entry)
 	{
 		// Build an MMIO capability.
 		auto buildMMIO = [&]() {
-			Debug::log("Building mmio capability {} + {}", target, size);
+			Debug::log("Building mmio capability {} + {} (raw: {})",
+			           entry.address,
+			           entry.size(),
+			           entry.sizeAndPermissions);
+			Debug::log("Permissions: {}", entry.permissions());
```
nwf-msr, line 416: Why two `Debug::log`s back to back?

davidchisnall: Because the second one was crashing in an earlier version. Probably not needed now.

sdk/include/compartment-macros.h
```diff
@@ -5,33 +5,78 @@
...
 		      "  auipcc  %0,"                                                  \
-		      "      %%cheri_compartment_pccrel_hi(__import_mem_" #name ")\n"  \
+		      "      %%cheri_compartment_pccrel_hi(" mangledName ")\n"         \
 		      "  clc     %0, %%cheri_compartment_pccrel_lo(1b)(%0)\n"          \
-		      : "=C"(ret));                                                    \
+		      : "=C"(ret)                                                      \
+		      : "i"(((permitLoad) ? (1 << 31) : 0) +                           \
+		            ((permitStore) ? (1 << 30) : 0) +                          \
+		            ((permitLoadStoreCapabilities) ? (1 << 29) : 0) +          \
+		            ((permitLoadMutable) ? (1 << 28) : 0)));                   \
```
nwf-msr, line 39: Can you use symbolic constants here?

```suggestion
		      : "i"(((permitLoad) ? (1 << ImportEntry::PermitLoad) : 0) +      \
		            ((permitStore) ? (1 << ImportEntry::PermitStore) : 0) +    \
		            ((permitLoadStoreCapabilities) ? (1 << ImportEntry::PermitLoadStoreCapabilities) : 0) + \
		            ((permitLoadMutable) ? (1 << ImportEntry::PermitLoadMutable) : 0))); \
```

If this needs to be pure C, can you move those constant definitions out of `ImportEntry` and make that refer to the new definition site?

davidchisnall: It needs to be pure C, and public headers should not include private headers from the loader.  

This is part of the ABI and we have tests for it, so I'm not too worried about things being mismatched.  Tests will fail if these constants are wrong.

davidchisnall: I also don't want anything else using these constants and, at some point, this logic will all go into the compiler, so I don't want a symbolic constant to suggest to users that this is something that they should actually depend on.

sdk/core/loader/boot.S
```diff
@@ -47,7 +47,10 @@ start:
 	la_abs			a3, __compart_headers_end
 	sub				a3, a3, a1
 	csetaddr		ca1, ca2, a1 // ca2 still has the G root.
-	csetboundsexact	ca1, ca1, a3
+	// FIXME: This should be a set bounds exact, but we currently don't have a
+	// 'pad to capability alignment' command in the linker script and it needs
+	// to span two sections.
```
rmn30, line 52: What are the consequences of inexact bounds?

davidchisnall: If we have memory safety bugs in the loader then it will read out of bounds when accessing the headers or possible write out of bounds when initialising the threads.

PR #137, opened 2023-11-20
Initial version of the hazard mechanism proposed in #133

### LINKED ISSUE #133: Would be nice to support hazard pointer like claims on heap objects

davidchisnall:
This provides a lightweight claims mechanism where each thread has two slots that can be used to store pointers to objects that have been temporarily claimed.  These are automatically cleared on cross-compartment call.

A Switcher function allows code to retrieve a (local) pointer to the current thread's space.  This is local so that it cannot be shared between threads (doing so would allow one thread to clear another thread's hazard slots while that thread entered another compartment). It is store-only to provide some defence in depth.

The hazard slots are synchronised with the allocator via an epoch mechanism.  The allocator increments the epoch when it starts scanning the list and again at the end.  If the value of the epoch counter changes during insertion or is odd then, the thread trying to insert needs to retry.  There is currently an unconditional futex wake on the allocator's free path to allow callers to sleep on the futex.  Given how rare contention is expected to be, it's possibly better to just sleep for one tick when this happens and avoid the notify.  This is a simple change to make.

This structure is not intended to be manipulated directly.  The `heap_claim_fast` library function provides a user-friendly wrapper around it.

Along the way, fixes a bug where quotas could be amplified.

Fixes #133


davidchisnall:
Some optimisation later:

```
	Original	New		Overhead
32	110578928	123680994	11.85%
64	58579947	65044143	11.03%
128	30624678	34142932	11.49%
256	16749468	18369632	9.67%
512	9623171		10526866	9.39%
1024	6562620		6945096		5.83%
2048	4268676		4531561		6.16%
4096	3342736		3479603		4.09%
8192	2618123		2667895		1.90%
16384	2184942		2209965		1.15%
32768	1955510		1966652		0.57%
65536	2554743		1910479		-25.22%
131072	2616802		3092058		18.16%
```

This version uses a futex_wait on the waiting side, but with a one-tick timeout and with priority boosting, so it should always kick the freeing thread out of free block in one tick.  It doesn't require the freeing side to do anything (we rely on the one-tick timeout to exit).  

With this speed, we can manage around 3 MiB/s allocation rates for 1024 byte allocations and over 300 KiB/s for 64-byte allocations at 20 MHz, which seems plenty.  In particular, it should be sufficient for 10 Mb/s networking with one malloc per packet.

sdk/core/allocator/main.cc
```diff
@@ -782,6 +797,13 @@ namespace
 		  *capability, *chunk, bodySize, isPrecise, reallyFree);
 	}
 
+	bool timeout_is_valid(Timeout *timeout)
+	{
+		return !heap_address_is_valid(timeout) &&
```
rmn30, line 802: This is worthy of a comment, at least and should be documented on the APIs that take a `Timeout`. By the way I observe that `heap_address_is_valid` returns true for pointers into the hazardQuarantine but I think that is fine as they should never escape the allocator.

sdk/core/loader/boot.cc
```diff
@@ -758,6 +758,16 @@ namespace
...
+		Capability<void *> hazardPointers =
+		  build<void *,
+		        Root::Type::RWGlobal,
+		        PermissionSet{Permission::Store,
+		                      Permission::LoadStoreCapability}>(
+		    LA_ABS(__export_mem_hazard_pointers),
+		    LA_ABS(__export_mem_hazard_pointers_end) -
+		      LA_ABS(__export_mem_hazard_pointers));
+		// Space per thread for hazard pointers.
+		static constexpr size_t HazardPointerSpace = 2 * sizeof(void *);
```
rmn30, line 770: Could we have a symbolic name for the number of hazard pointers per thread? Could it be configurable?

davidchisnall: Do you imagine ever supporting more than 2?  I think 2 is both the minimum number to be useful and the maximum number to be performant.

sdk/core/allocator/main.cc
davidchisnall: Agreed.  I think that should go in timeout.h in the generic coverage of timeouts.  In a follow-on, I am also going to add a check-timeout helper like this, but I want to see with respect to code size whether it makes more sense in the helpers library or as a static inline.

sdk/core/loader/boot.cc
rmn30: Perhaps not, but it would make the code more self-documenting rather than a magic constant. Bonus points if the same name could be used in the linker script.

davidchisnall: Yup, I agree it should be a symbolic constant, just not sure what scope it needs.  If you agree that we won't change 2, then it can go here.  

Unfortunately, the linker script is generated from lua and we don't have a mechanism for sharing constants between the build system and the code.  It would be nice to add one.

sdk/core/allocator/main.cc
davidchisnall: Comment added in timeout.h

sdk/core/loader/boot.cc
davidchisnall: Symbolic constant added.

sdk/boards/ibex-arty-a7-100.json
```diff
@@ -23,6 +23,14 @@
         "gpio_led0" : {
             "start": 0x8f00f000,
             "length": 0x800
+        },
+        "dmb_cfg": {
```
rmn30, line 27: What's this?

sdk/core/allocator/alloc.h
```diff
@@ -891,6 +902,46 @@ class MState
...
+	 */
+	bool hazard_quarantine_is_empty()
+	{
+		return hazardQuarantineSize == 0;
+	}
+
+	/**
+	 * Returns true of `ptr` is already in the quarantine list.
+	 */
+	bool hazard_quarantine_contains(Capability<void> ptr)
```
rmn30, line 921: There don't seem to be any calls to this function? I was a bit worried about the `is_subset_of`.

sdk/boards/ibex-arty-a7-100.json
```diff
@@ -23,6 +23,14 @@
...
+            "start": 0x8f0f0000,
+            "length": 0x1000
+        },
+        "dmb": {
+            "start": 0x90000000,
+            "length": 0x100000
```
nwf-msr, line 33: This hunk looks unrelated; am I missing something?

sdk/core/allocator/alloc.h
```diff
@@ -1160,26 +1250,135 @@ class MState
...
+			 * that they are waiting for us to walk the epoch list will block
+			 * until we're done.
+			 *
+			 * It must also happen before zeroing because the allocator is
+			 * running with interrupts enabled.  Were we to zero and then paint,
+			 * there would be a window in which preemption could allow a store
+			 * through a copy of the user capability (or its progeny) that undid
+			 * our work of zeroing!
+			 */
+			revoker.shadow_paint_range<true>(mem.address(), chunk.cell_next());
```
nwf-msr, line 1380: Perhaps a comment of `guard` going out of scope after this line, just so there's _something_ written that makes it look like this can't be CSE'd with the then branch above.

sdk/core/allocator/main.cc
```diff
@@ -732,12 +735,22 @@ namespace
...
+				if (ret == 0)
+				{
+					owner.quota += chunkSize;
+				}
+				return ret;
 			}
-			chunk.ownerID = 0;
+			// We've removed the owner, so refund the quota immediately, the
+			// free won't happen until the last claim goes away, but this is no
+			// longer the owner's responsibility.
```
rmn30, line 752: A bit unfortunate that this means there is memory that is not free (could remain allocated indefinitely) but isn't attributed to a quota. With the existing claim mechanism does it get charged to the claimer's quota?

sdk/core/allocator/alloc.h
```diff
@@ -891,6 +902,46 @@ class MState
 	size_t heapFreeSize;
 	size_t heapQuarantineSize;
 
+	/**
+	 * The number of entries currently in the `hazardQuarantine` array.
+	 */
+	size_t hazardQuarantineSize = 0;
```
nwf-msr, line 908:
```suggestion
	size_t hazardQuarantineOccupancy = 0;
```

since you use `hazardQuarantineSize` for the actual size of the reserved region, below?

sdk/lib/compartment_helpers/claim_fast.cc
```diff
@@ -0,0 +1,73 @@
...
+		values--;
+	}
+	if (!heap_address_is_valid(ptr2))
+	{
+		ptr2 = nullptr;
+		values--;
+	}
+	// If neither pointer refers to heap memory, set without synchronizing.  It
+	// doesn't matter if the revoker sees these or not, they cannot extend the
+	// lifetime of a heap object.
```
rmn30, line 31: Is this comment a bit stale? I don't understand it and I'm not sure it describes the code.

sdk/include/stdlib.h
```diff
@@ -145,6 +145,27 @@ void *__cheri_compartment("alloc")
...
+ * A null pointer can be used as a not-present value.  This function will
+ * treat operations on null pointers as unconditionally successful.  It returns
+ * `-ETIMEDOUT` if it failed to claim before the timeout expired or `EINVAL` if
+ * one or more of the arguments is neither null nor a valid pointer at the end.
+ *
+ * In the case of failure, neither pointer will have been claimed.
+ *
+ * This function is provided by the compartment_helpers library, which must be
+ * linked for it to be available.
+ */
```
rmn30, line 164: Is the idiomatic way to release a claim to call with `nullptr`?

nwf-msr (commented):
Nits, but generally LGTM

rmn30 (approved):

sdk/boards/ibex-arty-a7-100.json
davidchisnall: Yup, that was accidentally added.

sdk/core/allocator/alloc.h
davidchisnall: Does the comment above not say this?  It starts with 'Paint the shadow bitmap with the hazard epoch odd'.  I've added a comment to the guard.

sdk/boards/ibex-arty-a7-100.json
davidchisnall: Accidental things from the network stack bits.

sdk/lib/compartment_helpers/claim_fast.cc
davidchisnall: No, this refers directly to the code immediately below it.  Above, we've checked whether either pointer refers to heap memory.  Below, if they don't, we null out the hazard slots and return immediately.

sdk/include/stdlib.h
davidchisnall: Yup.  Or simply return from a cross-compartment call (or call another cross-compartment call).  For example, if you're yielding in a loop, you implicitly release them before each loop iteration, so you don't need the explicit call on retry.

sdk/core/allocator/alloc.h
nwf-msr: Sorry, I'd meant a comment _after_ the call to `shadow_paint_range` so that someone like future-me, reading upwards from the ends of the then and `else` scopes wouldn't forget that `guard` going out of scope was important.

PR #138, opened 2023-11-24
Fix doc comment for switcher_current_thread.

rmn30:


rmn30:
As an aside: do we have a convention for return values from library calls and cross compartment calls? This library call returns false on failure, which is fine, but that is not a good idea for compartment calls since they return -1 on forced unwind. This could cause confusion if we ever change a library call into a compartment call or vice versa, so wouldn't it be better to be consistent?

rmn30:
I also notice that the doc comment for `trusted_stack_has_space` references `trusted_stack_index`, which doesn't exist.

davidchisnall (approved):

PR #139, opened 2023-11-24
Remove message queues and semaphores from the scheduler.

davidchisnall:
Message queues are now provided by a shared library (for communication within a compartment) and a compartment (for communication between mutually distrusting compartments).  The version in the library can also be used between compartments with some care.

Along the way:

 - Allow LockGuard to work nicely with timeouts.
 - Make the C11/C++11 atomics sufficiently complete that we can claim to support them.
 - Allow check_pointer to skip the stack tests, for use in shared libraries.

The multiwaiter can still be used with message queues, but requires a call to set up the wait object.


davidchisnall:
First commit in the series.  The next will remove event groups and build them on top of futex.  We can then build some other locking abstractions (C versions of the ones in locks.hh) in shared libraries and have a rich set of communication primitives, all optional, all outside the scheduler.

sdk/include/cheri.hh
```diff
@@ -1058,13 +1058,20 @@ namespace CHERI
...
 	template<PermissionSet Permissions = PermissionSet{Permission::Load},
-	         typename T                = void>
+	         typename T                = void,
+	         bool CheckStack           = false>
 	__always_inline inline bool check_pointer(T *ptr, size_t space = sizeof(T))
 	{
-		return detail::check_pointer_internal<!Permissions.contains(
-		  Permission::Global)>(ptr, space, Permissions.as_raw());
+		return detail::check_pointer_internal < CheckStack &&
+		       !Permissions.contains(Permission::Global) >
```
nwf-msr, line 1073: Wow that... that sure is a formatting decision.  Would you consider...

```
        constexpr bool stackish = CheckStack && !Permissions.contains(Permission::Global);
		return detail::check_pointer_internal<stackish>
```

sdk/include/locks.hh
```diff
@@ -337,5 +344,26 @@ class LockGuard
...
 		}
 	}
+
+	/**
+	 * Conversion to bool.  Returns true if this guard owns the lock, false
+	 * otherwise.  This allows lock guards to be used with a timeout in
+	 * conditional blocks, such as:
+	 *
+	 * ```
+	 * if (LockGuard g{lock, timeout})
```
nwf-msr, line 264: The guard object is in scope until the end of the `if` and `else` blocks or the end of the enclosing scope?

sdk/lib/queue/queue.cc
```diff
@@ -0,0 +1,437 @@
...
+#else
+typedef uint32_t MessageQueueCounter;
+#endif
+
+#include <assembly-helpers.h>
+
+namespace
+{
+	/**
+	 * Helpers for the queue.  The queue uses two counters that wrap on double
```
nwf-msr, line 26: Might be worth making https://github.com/microsoft/cheriot-rtos/blob/main/sdk/include/ds/ring_buffer.h use that trick rather than the separate `bool empty` and then use that here?

sdk/lib/queue/queue.cc
```diff
@@ -0,0 +1,437 @@
...
+	 *
+	 * The remaining bits are free for other uses, but should not be modified
+	 * while the lock is not held.
+	 */
+	struct HighBitFlagLock
+	{
+		/**
+		 * A reference to the word that holds the lock.
+		 */
+		atomic<uint32_t> &lockWord;
```
nwf-msr, line 93: Perhaps an `atomic_ref`, if we support those, rather than requiring that the word always be `atomic`?

sdk/lib/queue/queue.cc
```diff
@@ -0,0 +1,437 @@
...
+		size_t index =
+		  counter >= handle.queueSize ? counter - handle.queueSize : counter;
+		auto             offset = index * handle.elementSize;
+		Capability<void> pointer{handle.buffer};
+		pointer.address() += offset;
+		return pointer;
+	}
+
+	/**
+	 * Flag lock that uses the high bit of a word for the lock.
```
nwf-msr, line 83: high _two_ bits!

sdk/lib/queue/queue.cc
```diff
@@ -0,0 +1,437 @@
...
+		/**
+		 * Try to acquire the lock.  Returns true on success, false on failure.
+		 */
+		bool try_lock(Timeout *t)
+		{
+			uint32_t value;
+			do
+			{
+				value = lockWord.load();
+				Debug::log("Lock word: {}", value);
```
nwf-msr, line 121: Not sure about the merits of keeping these `Debug::log`s around, but if we do, they should probably also include `&lockWord`, not just `value`.

sdk/lib/queue/queue.cc
```diff
@@ -0,0 +1,437 @@
...
+	if (overflow)
+	{
+		return -EINVAL;
+	}
+	// We need the counters to be able to run to double the queue size without
+	// hitting the high bits.  Error if this is the case.
+	//
+	// This should never be reached: a queue needs to be at least 512 MiB
+	// (assuming one-byte elements) to hit this limit.
+	if (((elementCount * 2) & HighBitFlagLock::reserved_bits()) != 0)
```
nwf-msr, line 238: For maximum paranoia, I think that could be `(elementCount | (elementCount * 2)) & HighBitFlagLoc::reserved_bits()`, lest that `* 2` push the offending bit off the top and badness result.  (The `__builtin_mul_overflow` above won't object to such `elementCount`s with `elementSize == 1`.)

nwf-msr (commented):
This looks good so far; it's pretty big so I haven't yet read all of it, but thought a "comment early, comment often" approach might be better than holding fire until the end.

davidchisnall:
FYI, in terms of reduction of scheduler size, the before version is 0x2380 bytes, after is 0x11b0 (no change to global size, but I think we now have some dead globals).  So this removes more than half the code in the global TCB for availability.  The entire scheduler is now just under 4.5 KiB.

The globals library is 0x1600 bytes (0xdc0 for the wrapper compartment) and event groups 0xff0 (but should not be used because they're a stupid abstraction), so somewhat larger in total, but with a much smaller TCB.

sdk/include/cheri.hh
```diff
@@ -1058,13 +1058,20 @@ namespace CHERI
...
+	 *
+	 * If the permissions do not include Global, then this will also check that
+	 * the capability does not point to the current compartment's stack.  This
+	 * behaviour can be disabled (for example, for use in a shared library) by
+	 * passing `false` for `CheckStack`.
 	 */
 	template<PermissionSet Permissions = PermissionSet{Permission::Load},
-	         typename T                = void>
+	         typename T                = void,
+	         bool CheckStack           = false>
```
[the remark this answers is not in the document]
davidchisnall: Yup.

sdk/include/locks.hh
davidchisnall: Guards are in scope for the duration of the statement (if statements take a block as a body).  They are not in scope in the `else` clause.

sdk/lib/queue/queue.cc
davidchisnall: Please do.  I mostly wanted you to check that this makes sense.  I haven't found anywhere else that uses this trick from a quick skim of things explaining how to design a ring buffer, but I can't quite believe I invented it because it seems too obvious.  I generalised this from the power-of-two version.  15 years ago I worked through a proof of why that worked because I didn't initially believe it, and I think it is basically the same for this.  The nice thing about this is that it requires one bit of extra state (which is otherwise padding because we don't have variable-bit-width memory) and doesn't require any divide operations (they're 15 cycles on Ibex).  That seems too nice to be real, so I suspect I got some maths wrong.

sdk/lib/queue/queue.cc
davidchisnall: We don't have those, but I think we want all accesses to be atomic anyway.

sdk/lib/queue/queue.cc
davidchisnall: Ooops, stale comment.

sdk/include/cheri.hh
davidchisnall: Ugh, clang-format seems to have lost the fact that this is a template parameter.  Fixed.

davidchisnall:
After the last commit, the size is now 0xfa0 (4000) bytes, which is more than a 10% additional saving.

sdk/include/cheri.hh
```diff
@@ -1058,13 +1058,23 @@ namespace CHERI
 	 * Checks that `ptr` is valid, unsealed, does not overlap the caller's
 	 * stack, and has at least `Permissions` and has at least `Space` bytes
 	 * after the current offset.
+	 *
+	 * If the permissions do not include Global, then this will also check that
+	 * the capability does not point to the current compartment's stack.  This
```
rmn30, line 1063:
```suggestion
	 * the capability does not point to the current thread's stack.  This
```

sdk/lib/queue/queue.cc
```diff
@@ -0,0 +1,435 @@
...
+		return counter;
+	}
+
+	/**
+	 * Returns true if and only if the `size`, `producer`, and `consumer`
+	 * counters indicate a full queue.
+	 */
+	constexpr bool is_full(uint32_t size, uint32_t producer, uint32_t consumer)
+	{
+		return producer - consumer == size;
```
rmn30, line 53: Shouldn't this check for `producer < consumer`?

sdk/lib/queue/queue.cc
```diff
@@ -0,0 +1,435 @@
...
+			{
+				if (consumer->wait(timeout, consumerCounter) == -ETIMEDOUT)
+				{
+					Debug::log("Timed out on futex");
+					return -ETIMEDOUT;
+				}
+				consumerCounter = counter_load(consumer);
+			}
+			auto entry = buffer_at_counter(*handle, producerCounter);
+			if (int claim = heap_claim_fast(timeout, handle->buffer, src);
```
rmn30, line 295: Is this claim held indefinitely, until the next claim or compartment call? Same for the one in `queue_receive`.

sdk/lib/locks/locks.cc
```diff
@@ -0,0 +1,230 @@
...
+			{
+				// If there are not already waiters, set the waiters flag.
+				if ((old & Flag::LockedWithWaiters) == 0)
+				{
+					Debug::Assert((old & 0xffff0000) == Flag::Locked,
+					              "Unexpected flag value: {}",
+					              old);
+					uint32_t addedWaiters = Flag::LockedWithWaiters;
+					addedWaiters |= (old & 0xffff);
+					lockWord.compare_exchange_strong(old, addedWaiters);
```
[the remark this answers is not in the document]
davidchisnall: Good catch.  Weirdly, I got this right in the high-bits flag lock.  Ho hum.

sdk/lib/queue/queue.cc
davidchisnall: The logic was also wrong in the items-remaining check.  I have now added some exciting recursive template magic to do compile-time exhaustive checking.

sdk/include/locks.h
```diff
@@ -0,0 +1,185 @@
...
+	/**
+	 * The count of the times the lock has been acquired by the same thread.
+	 * This must be initialised to 0.
+	 */
+	uint16_t depth __if_cxx(= 0);
+	/**
+	 * The ID of the thread that currently holds the lock.  This must be
+	 * initialised to -1 (or some other value that is not a valid thread ID).
+	 */
+	uint16_t owner __if_cxx(= -1);
```
nwf-msr, line 58: If the underlying `FlagLockState` is PI, do we need to track this again here?

davidchisnall:
After all of this and #142, the final sizes are:

Scheduler: 3904 bytes
Message queue library: 1768 bytes
Message queue compartment: 1552 bytes
Event groups library: 1080 bytes

This adds up to a total of 8304 bytes, which is *less* than the 9088 that we started with.  In addition:

 - The TCB for availability is less than half its initial size.
 - Message queues are more flexible (faster if within a compartment).
 - Event groups are a terrible idea and any firmware image that doesn't use them can save an additional 1 KiB.
 - The locks library (1112 bytes) reduces a lot of the size of other code that's using locks.  For example, the allocator drops from 9152 bytes to 8704 bytes and there's likely to be a similar saving across other things that use locks.

PR #142, opened 2023-11-30
Move thread_id_get() into the switcher.

davidchisnall:
This is reading a single value per thread, which never changes, and so adds only three instructions (including cret) to the switcher.  We previously had two ways of getting the thread ID:

 - thread_id_get() returned the ID, but required a full compartment switch into the scheduler.
 - thread_id_get_pointer() retrieved a cacheable pointer that could be dereferenced to give the current thread ID.

The latter form was fast, but could not be used in shared libraries because it required capturing the returned pointer.

This simplifies the locks APIs and means that the message-queue compartment can be entirely stateless (no mutable globals) and so a compromise cannot persist.

### DIFF
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -777,7 +777,7 @@ exception_entry_asm:
 	// Restore the stack pointer.  All other spilled values are spilled there.
 	clc                csp, TrustedStackFrame_offset_csp(ct1)
 	// Update the current frame offset.
-	csw                t2, TrustedStack_offset_frameoffset(ctp)
+	csh                t2, TrustedStack_offset_frameoffset(ctp)
 	// Do the loads *after* moving the trusted stack pointer.  In theory, the
 	// checks in `check_compartment_stack_integrity` make it impossible for
 	// this to fault, but if we do fault here then we'd end up in an infinite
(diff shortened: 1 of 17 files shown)


sdk/core/switcher/entry.S
```diff
@@ -777,7 +777,7 @@ exception_entry_asm:
 	// Restore the stack pointer.  All other spilled values are spilled there.
 	clc                csp, TrustedStackFrame_offset_csp(ct1)
 	// Update the current frame offset.
-	csw                t2, TrustedStack_offset_frameoffset(ctp)
+	csh                t2, TrustedStack_offset_frameoffset(ctp)
```
rmn30, line 780: Was this previously a bug where this write was overflowing into the padding?

rmn30 (approved):
Nice! Not sure whether a library call and `csrr` read would be faster or slower than inlined `thread_id_get_fast`.

nwf-msr (approved):
This very tiny addition to ASR-wielding code seems more than welcome.

sdk/core/switcher/entry.S
davidchisnall: Yes, but it was benign because it stored zeroes into padding, so embarrassing, but not a security issue.

PR #147, opened 2023-12-06
Fixes to C headers and debug output.

davidchisnall:
Most of these are fixes for either:

 - Macros that we don't test using don't actually work in C code.
 - C headers contain things that don't work if you include them in more than one compilation unit.

At the same time, add explicit overloads for printing uint16_t / uint32_t.  Without this, integer promotion promotes them to int and we print them as signed (decimal) numbers, not as hex.

### DIFF
--- a/sdk/lib/string/strchr.c
+++ b/sdk/lib/string/strchr.c
@@ -1,0 +1,14 @@
+#include <string.h>
+
+char *strchr(const char *s, int intChar)
+{
+	char c = (char)intChar;
+	while (*s != c)
+	{
+		if (!*s++)
+		{
+			return NULL;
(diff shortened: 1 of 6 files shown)


sdk/lib/string/strchr.c
```diff
@@ -0,0 +1,14 @@
+#include <string.h>
+
+char *strchr(const char *s, int intChar)
+{
+	char c = (char)intChar;
+	while (*s != c)
+	{
+		if (!*s++)
```
rmn30, line 8: I had to look up operator precedence for this but I think it's correct...

rmn30 (approved):

PR #148, opened 2023-12-13
Expose sleep-wait functions.

davidchisnall:
We didn't have generic functions for spinning / sleeping and they're quite important.

### DIFF
--- a/sdk/include/thread.h
+++ b/sdk/include/thread.h
@@ -5,8 +5,10 @@
 
 #include <cdefs.h>
 #include <compartment.h>
+#include <riscvreg.h>
 #include <stddef.h>
 #include <stdint.h>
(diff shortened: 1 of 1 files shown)


nwf-msr:
These are fine, I guess, even if I generally think large waits should be CPS hung off timers and short waits shouldn't be a thing.  Where did you find you needed them?

davidchisnall:
> These are fine, I guess, even if I generally think large waits should be CPS hung off timers and short waits shouldn't be a thing. Where did you find you needed them?

Short sleeps are normally needed (in interrupts-disabled things) for bit banging.  The longer one is needed when initialising the Ethernet PHY (requires a 167ms sleep, which can be a wait for a few ticks and then spin for any remaining time).

PR #149, opened 2023-12-13
Make the error messages more helpful.

davidchisnall:
The fail-on-simulator-error error handler is useful for debugging, but without the PCC that caused the crash it's somewhat less useful.


nwf-msr (approved):
LGTM

PR #152, opened 2023-12-13
Disable interrupts while printing debug log messages.

davidchisnall:


davidchisnall:
This breaks the build of the loader with debug logs enabled.  Not sure if it's worth fixing or whether we should close this.

nwf-msr:
This looks reasonable to me, but it looks like you have angered the CI gods by using this in the (non-compartmentalized) boot loader.  Perhaps the loader, which runs with IRQs deferred anyway, should have its own `log()` method?

davidchisnall:
The loader already has a different template specialisation of this, so I just added another template parameter for it.  You can't have too many template parameters...

PR #153, opened 2023-12-13
Add ethernet bits to Arty A7 board description.

davidchisnall:
This includes a driver for Kunyan's ethernet adaptor.


davidchisnall:
This depends on #148.

rmn30 (approved):
There are a few magic constants that might be nice to have symbolic names for (e.g. buffer length and addresses) although mostly they are only used once.

PR #154, opened 2023-12-15
Small fixes to the FreeRTOS compat layer:

davidchisnall:
- The order of clear-on-exit and wait-for-all flags was incorrect in the event-wait function.
 - Make sure that we can kick all waiters out of a sleep state when we delete an event group.
 - Make the queue error values match the macros that calling code compares them against.

### DIFF
--- a/sdk/lib/event_group/event_group.cc
+++ b/sdk/lib/event_group/event_group.cc
@@ -176,3 +176,20 @@ int eventgroup_get(EventGroup *group, uint32_t *outBits)
 	*outBits = group->bits;
 	return 0;
 }
+
+int eventgroup_destroy(SObjStruct *heapCapability, EventGroup *group)
+{
+	group->lock.lock();
+	// Force all waiters to wake.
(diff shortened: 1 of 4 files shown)


sdk/lib/event_group/event_group.cc
```diff
@@ -176,3 +176,20 @@ int eventgroup_get(EventGroup *group, uint32_t *outBits)
 	*outBits = group->bits;
 	return 0;
 }
+
+int eventgroup_destroy(SObjStruct *heapCapability, EventGroup *group)
+{
+	group->lock.lock();
```
rmn30, line 182: What will happen to any threads waiting on this lock once we free `group`? I guess they will timeout and then crash? Oh well.

rmn30 (approved):
I think rather than waking and cleanly returning when the group is destroyed any waiting threads will dereference an invalid capability and crash. Given this API is not supposed to be used and groups are probably never destroyed I'm not sure if this matters...

PR #157, opened 2024-01-02
Do not zero-pad in strncpy.

davidchisnall:
We have a strcpy implementation that wraps strncpy, so we end up trying to pad 4 GiBs of zeroes.


nwf-msr (approved):
LGTM

PR #158, opened 2024-01-03
Fix defining static sealing values with structures.

davidchisnall:
The test and malloc-capability macros were inconsistent in how they expected these to be use:

 - The test wanted to pass all arguments in braces.
 - The malloc macro wanted to pass them as separate arguments.

The implementation expected the former, but broke when you actually tried to pass multiple values.  This happened to work with malloc capabilities because we were zero-initialising the subsequent fields and so ignoring the explicit zeroes was fine.

This is now fixed to use a variadic macro (hopefully) correctly.

### DIFF
--- a/sdk/include/compartment-macros.h
+++ b/sdk/include/compartment-macros.h
@@ -187,8 +187,7 @@
 	  name = /* NOLINT(bugprone-macro-parentheses) */                          \
 	  {(uint32_t)&__sealing_key_##compartment##_##keyName,                     \
 	   0,                                                                      \
-	   initialiser,                                                            \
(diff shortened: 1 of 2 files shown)


nwf-msr (approved):
Ah, the C preprocessor.  While not the spec, https://en.cppreference.com/w/c/preprocessor/replace makes it sound like only balanced parentheses work to make comma tokens not count as function-like macro argument separators.

In any case, if CI is happy I think I'm happy, but perhaps `, ##__VA_ARGS__` wants to be `__VA_OPT__(,) ##__VA_ARGS__` in both places?

PR #161, opened 2024-01-15
Add a concept for a platform entropy source and a weak implementation.

davidchisnall:
The Ibex / Arty A7 uses the cycle count and the number of Ethernet interrupts to seed.  The first of these starts counting when the first-stage (ROM) loader runs and so depends on the speed at which we load the firmware over the UART, the second after the scheduler starts and depends on external network traffic.  These are not even slightly cryptographically secure, but they do provide enough entropy that we get different values from the PRNG across different runs and so can pretend for the purpose of a prototyping platform.

### DIFF
--- a/sdk/include/ds/xoroshiro.h
+++ b/sdk/include/ds/xoroshiro.h
@@ -7,11 +7,15 @@
 
 #include <cstdint>
 #include <cstdlib>
+#include <debug.hh>
+#include <initializer_list>
 
 namespace ds::xoroshiro
 {
+
 	namespace detail
 	{
+		using Debug = ConditionalDebug<false, "xoroshiro">;
 
 		/**
 		 * The xoroshiro+ (not ++) generator(s) of Blackman and Vigna's
@@ -22,7 +26,15 @@ namespace ds::xoroshiro
 		 * variables used internally (State), the type of each sample (Result),
 		 * and the seed state values (A, B, C).
 		 */
-		template<typename State, typename Result, State A, State B, State C>
+		template<typename State,
+		         typename Result,
+		         State A,
+		         State B,
+		         State C,
+		         State Jump0     = 0,
+		         State Jump1     = 0,
+		         State LongJump0 = 0,
+		         State LongJump1 = 0>
 		class XorOshiro
 		{
 			private:
@@ -45,12 +57,33 @@ namespace ds::xoroshiro
 				return (x << k) | (x >> (StateBits - k));
 			}
 
+			void jump(State jump0, State jump1)
+			{
+				const State Jump[] = {jump0, jump1};
+				uint64_t    s0     = 0;
+				uint64_t    s1     = 0;
+				for (int i : {0, 1})
+				{
+					for (int b = 0; b < 64; b++)
+					{
+						if (Jump[i] & uint64_t(1) << b)
+						{
+							s0 ^= x;
+							s1 ^= y;
+						}
+						next();
+					}
+				}
+				x = s0;
+				y = s1;
+			}
+
 			public:
 			XorOshiro(State x = 5489, State y = 0) : x(x), y(y)
 			{
 				// If both zero, then this does not work
-				if (x == 0 && y == 0)
-					abort();
+				Debug::Invariant((x != 0) || (y != 0),
+				                 "Invalid state, both x and y are zero");
 
 				next();
 			}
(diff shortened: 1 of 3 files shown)


sdk/include/platform/arty-a7/platform-entropy.hh
```diff
@@ -0,0 +1,65 @@
...
+
+	  /// Reseed the PRNG
+	  void reseed()
+	  {
+		  // Start from a not very random seed
+		  uint64_t seed = rdcycle64();
+		  prng.set_state(seed, seed >> 24);
+		  uint32_t interrupts = *ethernetInterruptCount;
+		  // Permute it with another not-very-random number
+		  for (uint32_t i = 0 ; i<((interrupts & 0xff) >> 8); i++)
```
nwf-msr, line 46: I am pretty sure you didn't mean `(interrupts & 0xff) >> 8`.

```suggestion
		  for (uint32_t i = 0 ; i<((interrupts & 0xff00) >> 8); i++)
```

sdk/include/ds/xoroshiro.h
```diff
@@ -85,7 +142,15 @@ namespace ds::xoroshiro
...
-	using P128R64 = detail::XorOshiro<uint64_t, uint64_t, 24, 16, 37>;
+	using P128R64 = detail::XorOshiro<uint64_t,
+	                                  uint64_t,
+	                                  24,
+	                                  16,
+	                                  37,
+	                                  0xdf900294d8f554a5,
+	                                  0x170865df4b3201fc,
+	                                  0xd2a98b26625eee7b,
+	                                  0xdddf9b1090aa7ac1>;
```
nwf-msr, line 153: "Let's see... types, small numbers... omg".  I trust that those are the right big numbers.

sdk/include/platform/concepts/entropy.h
```diff
@@ -0,0 +1,36 @@
...
+	 * secure random number.
+	 */
+	{
+		T::IsSecure
+		} -> std::convertible_to<const bool>;
+
+	/**
+	 * Must return a random number.  All bits of the value type are assumed to
+	 * be random, this should use a narrower type if it guarantees less
+	 * entropy.
```
nwf-msr, line 23: Requiring distillation in the application-facing source is probably fine, but I'm not sure it's worth it if one starts stacking/pooling sources a la Fortuna.

sdk/include/platform/arty-a7/platform-entropy.hh
```diff
@@ -0,0 +1,65 @@
...
+DECLARE_AND_DEFINE_INTERRUPT_CAPABILITY(                               
+  RevokerInterruptEntropy, InterruptName::EthernetReceiveInterrupt, true, false)                                         
+
+
+/**
+ * A simple entropy source.  This wraps a few weak entropy sources to seed a
+ * PRNG.  It is absolutely not secure and should not be used for anything that
+ * depends on cryptographically secure random numbers!  Unfortunately, there is
+ * nothing on the Arty A7 instantiation of CHERIoT SAFE that can be used as a
+ * secure entropy source.
```
nwf-msr, line 18: Huh!  That's kind of surprising.  Maybe something like https://dl.acm.org/doi/10.1145/3610295 is worth adding to the FPGA image?  But either way, this looks fine for testing and demo purposes.

davidchisnall: There are a bunch of ways you can do entropy sources (free-running ring oscillators are a favourite) on FPGAs.  The problem is not that it's impossible, it's that no one has done it for CHERIoT SAFE.

davidchisnall:
Reviewer feedback should be addressed and clang-format appeased.

sdk/include/platform/concepts/entropy.h
davidchisnall: I think that real silicon will be likely to come with a decent hardware random number generator (entropy source fed through some running hash construct) and so I don't expect to need Fortuna long term.

sdk/include/ds/xoroshiro.h
davidchisnall: Updated the comment to provide a citation for the magic numbers.  I am 100% confident that they are, in fact, numbers.

nwf-msr (approved):
LGTM

PR #162, opened 2024-01-17
A few small fixes:

davidchisnall:
- Add a missing extern "C"
 - Expose vsnprintf (we have it, it's useful, let's expose it).
 - Add explicit_bzero (it's just a different name for memset for us, since our memsset is in a shared library, but the compiler treats it differently).

### DIFF
--- a/sdk/lib/freestanding/memset.c
+++ b/sdk/lib/freestanding/memset.c
@@ -88,3 +88,9 @@ void *__cheri_libcall memset(void *dst0, int c0, size_t length)
 		} while (--t != 0);
 	return (dst0);
 }
+
+void __cheri_libcall explicit_bzero(void *s, size_t n)
+{
+	memset(s, 0, n);
+}
+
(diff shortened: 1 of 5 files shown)


sdk/lib/freestanding/memset.c
```diff
@@ -88,3 +88,9 @@ void *__cheri_libcall memset(void *dst0, int c0, size_t length)
 		} while (--t != 0);
 	return (dst0);
 }
+
+void __cheri_libcall explicit_bzero(void *s, size_t n)
+{
+	memset(s, 0, n);
```
nwf-msr, line 94: Does this result in an extra call/return within the library?

nwf-msr (approved):
LGTM; one question that's not reason to delay merging.

sdk/lib/freestanding/memset.c
davidchisnall: It should be a tail call. I don’t think we can merge the entries, but the compiler will lower calls within the same compilation unit as direct. I think it may not do tail calls right now, but that’s a simple fix soon.

PR #164, opened 2024-01-25
Fix erroneous check for empty trusted stack.

rmn30:
The check in `pop_trusted_stack_frame` needs to compare the trusted stack offset against the offset of frames field in `TrustedStack` to check whether we have reached the top of the trusted stack  (which grows upwards). This is the same as the check in `handle_error`.

Spotted because tests were raising several exceptions before eventually exiting due to failure to spot empty trusted stack.

### DIFF
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -769,7 +769,7 @@ exception_entry_asm:
 	clear_hazard_slots ctp, ct2
 	// make sure there is a frame left in the trusted stack
 	clhu               t2, TrustedStack_offset_frameoffset(ctp)
-	li                 tp, TrustedStackFrame_size
+	li                 tp, TrustedStack_offset_frames
 	bgeu               tp, t2, .Lset_mcause_and_exit_thread
 	cspecialr          ctp, mtdc
 	addi               t2, t2, -TrustedStackFrame_size


rmn30:
This makes me suspicious of the decrement of the frame offset at line 689 because it does not have a similar check...

davidchisnall:
The one at 689 looks fine, it's looking inside the *current* trusted stack frame to find something.  This must exist because if it didn't then we wouldn't be running any code.  We should be checking for underflow when we pop (which, apparently, we're doing incorrectly).

davidchisnall (approved):

davidchisnall:
Hmm, interesting.  We probably should push a trusted stack frame when we call into `thread_exit`.

rmn30:
Or maybe prevent popping to the dummy stack frame and call `thread_exit` earlier?

davidchisnall:
> switcher re-enters error handler and this time detects fault in switcher (via PCC base) so attempts to unwind again. Now there is no trusted stack frame to unwind so call scheduler's thread_exit. Note that thread_exit runs with trusted_stack_frameoffset == TrustedStack_offset_frames.

Hmm, I'm not sure that's actually the case.  The scheduler's thread-exit is not an entry point, it's just an exception code in the scheduler's exception handler.  The exception handler is not allowed to make cross-compartment calls or to trap (if it does, it should die).  While it's running, the trusted stack register [should be null](https://github.com/microsoft/cheriot-rtos/blob/e4dc3234eac2b81984242dcc7140f52e15931d6c/sdk/core/switcher/entry.S#L356), which means that entering the switcher via the exception path is treated as a double fault.  Once we have a mechanism for performing a reset, that path will reset the CPU.

rmn30:
To see what happened added a trap to the scheduler in the `thread_exit` case. The switcher immediately noticed the null `mtdc` and entered an exception loop. I guess that's fine as the scheduler is already trusted for availability.

davidchisnall:
Yup, that's the expected behaviour.  Once we have a way of asserting reset from the core, that's what we'll do.

PR #165, opened 2024-01-29
Enable linker dead-code elimination.

davidchisnall:
This is essential for BearSSL, which wants to compile everything but then throw away all of the things that are not referenced by your specific use case.  It's nice to have for everything else.  It reduces the size of the FreeRTOS network stack by around 7.5%.

Savings in the test suite are very small (around 1%) and mostly come from some functions in headers that should be `inline` and aren't.

### DIFF
--- a/sdk/compartment.ldscript
+++ b/sdk/compartment.ldscript
@@ -71,10 +71,3 @@ SECTIONS
 		*(.comment);
 	}
 }
-# Export the exported function table.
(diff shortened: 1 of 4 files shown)


davidchisnall:
This is failing because a compiler fix to mark exports as roots was not yet merged.  Currently waiting for the PR to run then I can update the dev container.

Issue #168, opened 2024-02-03
Destruction lock


davidchisnall:
We currently have a nice priority inheriting mutex, but for a lot of use cases we want a lock that can be upgraded to a destruction mode.  This needs the following properties:

 - It sets an extra bit in the futex word that causes other attempts to acquire it to fail.
 - It wakes any pending waiters and forces them to fail to acquire the lock immediately.

This makes it easy to have an object containing a lock and the lock used to protect the object during destruction.  Normal operations acquire the lock, operate over the object, and release the lock.  Destruction operations acquire the lock, upgrade it to destruction mode, destroy the object.  As soon as the object is in destruction mode, all threads waiting on the lock are woken and fail, all future attempts to lock fail either by trapping (the object is gone, don't try to acquire it) or by detecting that the state is 'going away'.

PR #169, opened 2024-02-06
Implement lock destruction feature + cleanup in locking-related code

hlef:
This PR contains three commits:

- First one fixes some stale comments which I found somewhat confusing
- Second one implements lock destruction (see issue #168)
- Last one does some cleanup in the locking tests for coherence and consistency

It would be nice to implement `test_destruct_flag_lock_acquire` with templating, unfortunately we are hitting a compiler bug which prevents us from doing that. Let's add templating to that function once the compiler bug is reported and fixed.

### DIFF
--- a/sdk/lib/locks/locks.cc
+++ b/sdk/lib/locks/locks.cc
@@ -31,7 +31,9 @@ namespace
 			/// The lock is held.
 			Locked = 1 << 16,
 			/// The lock is held and one or more threads are waiting on it.
-			LockedWithWaiters = 1 << 17
+			LockedWithWaiters = 1 << 17,
+			/// The lock is held and set in destruction mode.
+			LockedInDestructMode = 1 << 18
 		};
 
 		public:
@@ -50,6 +52,19 @@ namespace
 				{
 					return 0;
 				}
+
+				// We are not setting the LockedInDestructMode
+				// bit in our expected value (`old`), so the
+				// CAS will always fail if the bit is set.
+				// Being locked for destruction is unusual (it
+				// should happen at most once for any lock) and
+				// so we don't bother to check for it unless
+				// we've failed to acquire the lock.
+				if ((old & Flag::LockedInDestructMode) != 0)
+				{
+					return -ENOENT;
+				}
+
 				if (!timeout->may_block())
 				{
 					return -ETIMEDOUT;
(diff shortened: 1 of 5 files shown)


sdk/lib/locks/locks.cc
```diff
@@ -136,14 +132,33 @@ namespace
 				lockWord.notify_all();
 			}
 		}
+
+		/**
+		 * Set the destruction bit in the flag lock word and wake
+		 * waiters. Assumes that the lock is hold by the caller.
```
davidchisnall, line 138:
```suggestion
		 * waiters. Assumes that the lock is held by the caller.```

**Comment by hlef** [comment]
@microsoft-github-policy-service agree company="SCI Semiconductor Ltd."

**Comment by hlef** [comment]
Thanks for the review @davidchisnall !

I have updated the PR with fixes to your comments. Two things to note:
- I did not do exactly what you suggested with `TEST(lock.lock())` in the async, because the `lock.lock()` API does not return a bool. To reach a compromise, I added an assert on `try_lock()` before `lock.lock()`. Not perfect, but better than nothing?
- I did not add an assert for the API misuse `lock.unlock()`. Would you like me to add the assert? If so, what would be the preferred approach to reset the lock?

**Comment by hlef** [comment]
For the record, the compiler bug which prevents us from using templating in the second test is this one: https://github.com/CHERIoT-Platform/llvm-project/issues/16

File: sdk/lib/locks/locks.cc [hunk 2]
```diff
@@ -63,6 +52,17 @@ namespace
 				{
 					return 0;
 				}
+
+				// Do not attempt to lock if the destruct mode
+				// flag is set. Note that the lock is in
+				// destruction mode iff it is already locked,
```
davidchisnall, line 58: I don't think that's the reason: we are not setting the locked-for-destruction bit in our expected value, so the CAS will always fail if the lock-for-destruction bit is set.  Being locked for destruction is unusual (it should happen exactly once for any lock) and so we don't bother to check for it unless we've failed to acquire the lock.

sdk/lib/locks/locks.cc
```diff
@@ -136,14 +136,33 @@ namespace
 				lockWord.notify_all();
 			}
 		}
+
+		/**
+		 * Set the destruction bit in the flag lock word and wake
+		 * waiters. Assumes that the lock is held by the caller.
+		 *
+		 * Note: This does not check that the lock is owned by the
```
davidchisnall, line 144: It might be nice to have an assert that does check.  There's a `Debug::Assert` variant that takes a lambda as the argument and so will not compile the check unless we've turned on lock debugging.

tests/locks-test.cc
```diff
@@ -65,6 +65,81 @@ namespace
...
+	void test_destruct_lock_wake_up(Lock &lock)
+	{
+		modified = false;
+
+		Timeout t{1};
+		lock.try_lock(&t);
+
+		// Create an async that tries to grab the lock
+		async([&]() {
+			lock.lock();
```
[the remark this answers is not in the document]
davidchisnall: The lock with no timeout will block this thread forever if the wake fails.  Subsequent tests will then deadlock.

hlef:
@davidchisnall

> The lock with no timeout will block this thread forever if the wake fails. Subsequent tests will then deadlock.

I don't think that this is still a problem in the current version. If the wake fails, we are guaranteed to fail `TEST(modified == true, ...)`, which will prevent other tests from being executed anyways.

sdk/lib/locks/locks.cc
hlef: I agree, these would be nice to have!

I suggest to add them in a separate commit. Currently, most methods in this file make this assumption without any assert. The commit which adds the assert could do so on all methods at once.

hlef:
@davidchisnall I updated the PR to address your comment on "Do not attempt to lock if the destruct mode...".

Regarding the assert, I suggest to do that as part of another PR since most of the code in `locks.cc` would benefit from this.

davidchisnall:
> @davidchisnall
> 
> > The lock with no timeout will block this thread forever if the wake fails. Subsequent tests will then deadlock.
> 
> I don't think that this is still a problem in the current version. If the wake fails, we are guaranteed to fail `TEST(modified == true, ...)`, which will prevent other tests from being executed anyways.

If any of the asserts fire on the main thread, they will force unwind out and the unlock will not be reached. The next test will then run. Try it and see.

hlef:
> > @davidchisnall
> > > The lock with no timeout will block this thread forever if the wake fails. Subsequent tests will then deadlock.
> > 
> > 
> > I don't think that this is still a problem in the current version. If the wake fails, we are guaranteed to fail `TEST(modified == true, ...)`, which will prevent other tests from being executed anyways.
> 
> If any of the asserts fire on the main thread, they will force unwind out and the unlock will not be reached. The next test will then run. Try it and see.

Alright, I have reworked the code a bit, now it should be good?

hlef:
Thank you for the comments, pushed a new version to address them.

davidchisnall (approved):
Looks great, thank you!

PR #171, opened 2024-02-12
Bug fixing and asserts in the locking subsystem

hlef:
This PR contributes:
- Bug fixing for an overflow bug in ticket locks
- Add asserts to catch potential bugs in the locking subsystem
- Expose debugging options for the locking subsystem through the build system


davidchisnall (approved):
Looks great, thanks!

PR #172, opened 2024-02-13
Add a script to measure firmware image sizes

hlef:
We need a standard and common way of measuring firmware images sizes to optimize this metric.

Add a helper script to `scripts/` to measure this.

Re-use the LLVM tool finding bit of the Ibex script by factoring it out in its own file.

### DIFF
--- a/scripts/output_code_size.sh
+++ b/scripts/output_code_size.sh
@@ -1,0 +1,98 @@
+#!/bin/sh
+
+SCRIPT_DIRECTORY="$(dirname "$(realpath "$0")")"
+. ${SCRIPT_DIRECTORY}/includes/helper_find_llvm_install.sh
+
+OBJDUMP=$(find_llvm_tool_required llvm-objdump)
+
+MACHINE_READABLE=0
+
+print_compartment_size() {
+	# Print all sections that match passed compartment name (potentially
+	# multiple ones, as compartments covers code and various data
+	# sections), and sum their sizes.
+	#
+	# Note that this does not validate compartment names. An invalid
+	# compartment name may return either 0kB, or an arbitrary number if the
+	# provided name matches legitimate compartments. This would be nice to
+	# fix at some point.
+	SIZE=$(${OBJDUMP} --headers $1 | grep -i $2 |
+		awk '{ sum += "0x"$3 } END { print sum}')
+
+	if [ "$MACHINE_READABLE" -eq 0 ]; then
+		KB_SIZE=$(echo $SIZE | awk '{ print ($1 / 1024) "kB"}')
+		echo "Size of compartment '$2': ${KB_SIZE}"
+	else
+		echo "${SIZE}"
+	fi
+}
+
+print_full_code_size() {
+	# Print all sections between the start of the firmware (`loader_start`)
+	# and the end of `__cap_relocs`, and sum the sizes of sections
+	SIZE=$(${OBJDUMP} --headers $1 |
+		awk '/loader_start/{f=1} /__cap_relocs/{f=0;print} f' |
+		awk '{ sum += "0x"$3 } END { print sum}')
+
+	if [ "$MACHINE_READABLE" -eq 0 ]; then
+		KB_SIZE=$(echo $SIZE | awk '{ print ($1 / 1024) "kB"}')
+		echo "Size of the full binary: ${KB_SIZE}"
+	else
+		echo "${SIZE}"
+	fi
+}
+
+help() {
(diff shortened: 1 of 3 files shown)


scripts/helper_find_llvm_install.sh
```diff
@@ -0,0 +1,25 @@
+# Constant: location of the custom tool binaries in the dev container
```
davidchisnall, line 1: I was confused by the missing shbang in a .sh file.  Maybe change the extension to .inc to indicate that this should not be run directly?  Or move it to an `includes` directory?

scripts/output_code_size.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
```
davidchisnall, line 1: Nope.  This should either be:

```
#!/bin/sh
```

```
#!/usr/bin/env bash
```

And we should not be using bash extensions in portable scripts, so it should be the former.

scripts/output_code_size.sh
```diff
@@ -0,0 +1,80 @@
...
+
+SCRIPT_DIRECTORY="$(dirname "$(realpath "$0")")"
+. ${SCRIPT_DIRECTORY}/helper_find_llvm_install.sh
+
+OBJDUMP=$(find_llvm_tool llvm-objdump)
+
+if [ ! -x ${OBJDUMP} ] ; then
```
davidchisnall, line 8: It would be nice to have a second argument to `find_llvm_tool` for whether to fail if the tool is not found.  I think in all of our current uses we want to print an error and have the script exit with a failure exit code if it can't find the tool.

scripts/output_code_size.sh
```diff
@@ -0,0 +1,80 @@
...
+	echo "Size of compartment '$2': ${SIZE}"
+}
+
+print_full_code_size() {
+	# Print all sections between the start of the firmware (`loader_start`)
+	# and the end of `__cap_relocs`, and sum the sizes of sections
+	SIZE=$(${OBJDUMP} --headers $1 |
+		awk '/loader_start/{f=1} /__cap_relocs/{f=0;print} f' |
+		awk '{ sum += "0x"$3 } END { print (sum / 1024) "kB"}')
+	echo "Size of the full binary: ${SIZE}"
```
davidchisnall, line 33: This doesn't need to be separate, just put it in the `END` block of the awk program.

scripts/output_code_size.sh
```diff
@@ -0,0 +1,80 @@
...
+	# multiple ones, as compartments covers code and various data
+	# sections), and sum their sizes.
+	#
+	# Note that this does not validate compartment names. An invalid
+	# compartment name may return either 0kB, or an arbitrary number if the
+	# provided name matches legitimate compartments. This would be nice to
+	# fix at some point.
+	SIZE=$(${OBJDUMP} --headers $1 | grep -i $2 |
+		awk '{ sum += "0x"$3 } END { print (sum / 1024) "kB"}')
+	echo "Size of compartment '$2': ${SIZE}"
```
davidchisnall, line 24: Same as below.

scripts/output_code_size.sh
```diff
@@ -0,0 +1,80 @@
...
+	echo "Size of the full binary: ${SIZE}"
+}
+
+help() {
+   echo "Determine the size of a CHERIoT firmware image."
+   echo
+   echo "Syntax: $(basename "$0") [-h|c|f]"
+   echo "  -h                                         Print this help."
+   echo "  -f [firmware location]                     Print the size of the entire firmware image."
+   echo "  -c [firmware location] [compartment name]  Print the size of passed compartment."
```
davidchisnall, line 42: It would be good to have a more raw format (without the 'size of...' bits) as an option, so that we can have a CI script that records the size of each compartment and the total image in the test suite and lets us track code growth.

scripts/helper_find_llvm_install.sh
hlef: Good point, I will move it to an `includes` directory. I tried `.inc`, but it prevents vim from correctly detecting the file as a shell script for syntax highlighting.

scripts/output_code_size.sh
hlef: My bad! This is leftover from a previous version of the script...

scripts/output_code_size.sh
hlef: Agree, I will do that as a wrapper.

hlef:
@davidchisnall  Alright, updated with fixes to your comments :slightly_smiling_face:

davidchisnall (approved):

PR #173, opened 2024-02-14
Move `check_pointer_internal` to a library to optimize code size.

hlef:
Moving `check_pointer_internal` to a library allows us to keep a single copy of the function for the entire firmware. For the test suite, this reduces the overall firmware size by 128B.

- As part of this, rename `CHERI::detail::check_pointer_internal` to `::check_pointer` and expose it externally as a C-level API.

- Since we are moving the function to a library, modify its signature to pass the `CheckStack` boolean through arguments instead of templating.

- Also use that opportunity to clarify ambiguities in the check stack comment of the function.


sdk/include/cheri.h
```diff
@@ -0,0 +1,21 @@
...
+#include <compartment-macros.h>
+
+/**
+ * C API for `check_pointer`.
+ *
+ * This function behaves identically to `CHERI::check_pointer`, but is exposed
+ * without templating and C++ types.
+ *
+ * To reduce code size, this function is provided as part of the compartment
+ * helper library.
```
davidchisnall, line 16: Please can you document this without reference to the C++ version (or with just a note for C++ people that they should consider the C++ version instead)?  C programmers shouldn't have to read C++ code to understand what C APIs do.

sdk/lib/compartment_helpers/README.md
```diff
@@ -4,5 +4,4 @@ Compartment helpers library
 This library includes functions that help securing compartment boundaries.
 
 - [`claim_fast.cc`] contains the `heap_claim_fast` function.
-
-
+- [`check_pointer.cc`] contains the `::check_pointer` function.
```
davidchisnall, line 7:
```suggestion
- [`check_pointer.cc`] contains the `check_pointer` function.
```

We don't use :: in the document for anything else in the global namespace in these libraries.

hlef:
@davidchisnall Thanks for the comments, all addressed.

sdk/include/cheri.h
```diff
@@ -0,0 +1,25 @@
+// Copyright CHERIoT Contributors.
+// SPDX-License-Identifier: MIT
+
+#pragma once
+#include <compartment-macros.h>
+#include <stddef.h>
+#include <stdint.h>
+
+/**
+ * Checks that `ptr` is valid, unsealed, does not overlap the caller's stack,
```
davidchisnall, line 10: This contradicts the next paragraph.

sdk/include/cheri.h
```diff
@@ -0,0 +1,25 @@
...
+ * has at least `rawPermissions`, and has at least `space` bytes after the
+ * current offset.
+ *
+ * If the permissions do not include Global, then this will also check that the
+ * capability does not point to the current thread's stack.  This behaviour can
+ * be disabled (for example, for use in a shared library) by passing `false`
+ * for `checkStackNeeded`.
```
davidchisnall, line 17:
```suggestion
 * If the permissions do not include Global and `checkStackNeeded` is `false`, then this will also check that the
  * capability does not point to the current thread's stack.
```

sdk/xmake.lua
```diff
@@ -759,6 +759,7 @@ function firmware(name)
 		add_rules("firmware")
 		-- TODO: Make linking the allocator optional.
 		add_deps(name .. ".scheduler", "cheriot.loader", "cheriot.switcher", "cheriot.allocator")
+		add_deps("compartment_helpers")
```
davidchisnall, line 762: Don't unconditionally add an optional library to everything!

hlef:
@davidchisnall Thank you for the additional comments. Fixed!

davidchisnall (approved):

Issue #131, opened 2023-11-01
heap_free_all frees sealed objects


davidchisnall:
The heap_free_all function frees all objects allocated with an allocation capability.  This means that it can be used to attack compartments that have allocated memory for a given caller.

It shouldn't, and we should provide a heap_free_all_sealed that frees everything sealed with a particular type that can be used for cleanup.

This is somewhat complicated by the fact that we do want to drop claims on sealed objects, we just don't want to allow them to actually be freed. @nwf, what do you think the right behaviour is?

davidchisnall:
Thinking a bit more about this after doing a load of network stack things:

In the network stack, at least, sockets are reachable only by the sealed capability and they point to a load of other things that are owned by the network stack that are owned by the socket structure and so there is no way to gracefully clean them up.  Even a `heap_free_all` variant that requires an authorising sealing capability would not help here because the network stack needs to gracefully clean things up (which it actually does by posting a message to another thread to serialise all deallocation events).

The API that I think I want is actually more like this:

```
int heap_free_all_sealed(Timeout *timeout, SObj mallocCapability, __cheri_callback void (*callback)(SObj);
```

This would find all sealed objects allocated with (or claimed by) `mallocCapability` and call `callback` on them.  The callback could then gracefully free everything by dispatching to the correct cleanup function for all sealed objects that it knows about.  

For this to be useful, I think we need two additional APIs.  The first lives in the token library:

```
uint32_t token_type(SObj)
```

This returns the type.  When you're handed a sealed capability, you can use this to record the type so that you know where to dispatch the cleanup.

I think we also need:

```
int heap_claims_count(void*);
```

This returns the number of claims that exist on an object (or an error) and can be called wit sealed or unsealed objects.  In the network stack's socket close function, we require that the caller pass their own malloc capability so that we can drop the claim.  We don't free it if dropping the claim fails, so that you can separate the ability to allocate and deallocate sockets from the ability to send and receive over them trivially.  This means that we need a fallback path for 'oh, dropping the claim failed, but you've already dropped the claim via `heap_free_all` so that's fine' (as a side effect, anyone with the socket handle can close it if the compartment that owns it has reset itself, but I don't think that's a problem).

@nwf / @nwf-msr, what do you think?

nwf-msr:
I like this API better, to be sure, but I worry that
```
int heap_free_all_sealed(Timeout *timeout,
                         SObj mallocCapability,
[...]

PR #174, opened 2024-02-16
Exclude sealed objects from `heap_free_all`.

davidchisnall:
This is discussed in some detail in #131

### DIFF
--- a/sdk/core/allocator/alloc.h
+++ b/sdk/core/allocator/alloc.h
@@ -289,7 +289,13 @@ namespace displacement_proxy
 struct __packed __aligned(MallocAlignment)
 MChunkHeader
 {
-	static constexpr size_t OwnerIDWidth = 14;
+	/**
+	 * Each chunk has a 16-bit metadata field that is used to store a small
+	 * bitfield and the owner ID in the remaining bits.  This is the space not
+	 * consumed by the metadata.  It must be reduced if additional bits are
+	 * stolen for other fields.
+	 */
+	static constexpr size_t OwnerIDWidth = 13;
 	/**
 	 * Compressed size of the predecessor chunk.  See cell_prev().
 	 */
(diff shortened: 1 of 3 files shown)


davidchisnall:
@hlef

sdk/core/allocator/alloc.h
```diff
@@ -289,7 +289,13 @@ namespace displacement_proxy
...
 MChunkHeader
 {
-	static constexpr size_t OwnerIDWidth = 14;
+	/**
+	 * Each chunk has a 16-bit metadata field that is used to store a small
+	 * bitfield and the owner ID in the remaining bits.  This is the space not
+	 * consumed by the metadata.  It must be reduced if additional bits are
+	 * stolen for other fields.
+	 */
+	static constexpr size_t OwnerIDWidth = 13;
```
nwf-msr, line 298: For a follow-up PR, perhaps: do we ever actually check that the `uint16_t` owner IDs we're passing around fit in this width?  Maybe just in debug builds?

sdk/core/allocator/alloc.h
```diff
@@ -301,8 +307,15 @@ MChunkHeader
...
 	/// The unique identifier of the allocator.
 	uint16_t ownerID : OwnerIDWidth;
-	bool     isPrevInUse : 1;
-	bool     isCurrInUse : 1;
+	/**
+	 * Is this a sealed object?  If so, it should be exempted from free in
+	 * `heap_free_all` because deallocation requires consensus between the
+	 * holder of the allocator capability and the holder of the sealing
+	 * capability.
+	 */
```
nwf-msr, line 315: Probably also worth a quick comment that this, like `isCurrInUse`, effectively is a _type tag_ on the body of the chunk, indicating that it is a `struct SObjStruct`.

By that token, it'd be good explicitly zero the bit in `mspace_free` (or maybe `quarantine_pending_push`) and to teach the `ok_free_chunk` consistency test that `isSealedObject` must be `0` (because we can't have both a `struct SObjStruct` and a `MChunk` or `TChunk` in the chunk body).  (At the moment, the bit is ignored through the entire quarantine/free/coalesce part of heap bytes' lifetimes, and will be set or cleared when a header is allocated.)

sdk/core/allocator/alloc.h
davidchisnall: In fact, we do!

https://github.com/microsoft/cheriot-rtos/blob/8fc5d7da8fd15731580fc8031dc73b65ebc0e581/sdk/core/allocator/main.cc#L358

sdk/core/allocator/alloc.h
davidchisnall: That's probably a good idea when we're debugging.  I'd rather skip the extra instructions, but if we're already clearing the in-use bit then it can be folded into the same read-modify-write instruction.

davidchisnall: Done.

PR #184, opened 2024-03-12
Fix unlimited timeouts.

hlef:
Currently unlimited timeouts are not really unlimited, because we still do the substraction in `elapse`. Fix this by only doing the substraction when remaining is not `UnlimitedTimeout`.

While at it, improve the documentation, and add a test.


nwf (approved):
LGTM

davidchisnall (approved):

PR #190, opened 2024-03-22
Add `memchr` to the standard library.

hlef:
This is small and necessary to have `std::string_view::find`, which is useful, among others, for validation of input in the network stack.

Add a small test for the implementation.

### DIFF
--- a/sdk/lib/string/memchr.c
+++ b/sdk/lib/string/memchr.c
@@ -1,0 +1,23 @@
+// Copyright SCI Semiconductor and CHERIoT Contributors.
+// SPDX-License-Identifier: MIT
+
+#include <string.h>
+
+const void *__cheri_libcall memchr(const void *voidString,
+                                   int         intChar,
+                                   size_t      n)
+{
+	const unsigned char  c = (unsigned char)intChar;
+	const unsigned char *s = (const unsigned char *)voidString;
+
+	for (size_t i = 0; i != n; i++)
+	{
+		if (*s == c)
+		{
+			return (const void *)s;
+		}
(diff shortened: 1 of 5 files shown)


sdk/lib/string/memchr.c
```diff
@@ -0,0 +1,21 @@
+// Copyright SCI Semiconductor and CHERIoT Contributors.
+// SPDX-License-Identifier: MIT
+
+#include <string.h>
+
+const void *__cheri_libcall memchr(const void *voidString, int intChar, size_t n)
+{
+	const char  c = (char)intChar;
+	const char *s = (const char *)voidString;
```
davidchisnall, line 9: The specification for `memchr` requires that the comparison be interpreting the bytes as `unsigned char`, not `char`.

tests/misc-test.cc
```diff
@@ -47,7 +48,42 @@ void check_timeouts()
...
+ * - memchr returns NULL when the string does not contain the character (test
+ *   for non-NULL terminated string).
+ * - memchr does not stop at the first \0 character.
+ */
+void check_memchr()
+{
+	debug_log("Test memchr.");
+
+	char _string[] = {'C', 'H', 'E', 'R', 'R', 'I', 'E', 'S'};
+	auto string = std::string_view(_string, sizeof(_string));
```
davidchisnall, line 67: I don't think this use of a string view is adding anything.

hlef:
@davidchisnall Thanks! Comments addressed.

davidchisnall (approved):

PR #191, opened 2024-04-03
Incomplete drivers and board description for sonata

HU90m:
I'm opening this Draft PR, so you can see what I have as I work on putting together a board description and drivers for the [sonata system](https://github.com/lowRISC/sonata-system)


HU90m:
FYI: At the moment, I'm trying to get the simple hello world example running; I'm currently getting [here](https://github.com/microsoft/cheriot-rtos/blob/8b6ec4f8605c5716065f078a9d27af27c7c88bff/sdk/core/loader/boot.cc#L1208):

```
loader: Error handler for compartment is 24)
loader: First pass to find sealing key imports
../../sdk/core/loader/types.h:956 Invariant failure in size
Unused bits in sizeAndPermissions are not zero, field contains 0x66120000
```

sdk/core/loader/boot.S
```diff
@@ -13,6 +13,10 @@
     .p2align 2
     .type start,@function
 start:
+	.rept 32
+	.word 0x0000006f
+	.endr
+
```
HU90m, line 19: I've added this to fill the ibex interrupt vector. How should I set up interrupts properly for the CHERIoT RTOS?

davidchisnall: CHERIoT doesn't use the vectored interrupts (CHERIoT Ibex doesn't support the vectored mode).  Interrupts are configured by the scheduler.  If Sunburst is doing something different for the interrupt controller (not the CLINT + PLIC combination that we normally use with CHERIoT Ibex) then you'll need a custom driver.

sdk/boards/sonata.json
```diff
@@ -0,0 +1,32 @@
+{
+    "devices": {
+        "clint": {
+            "start": 0x2000000,
+            "length":  0x10000
+        },
+        "uart": {
+            "start": 0x81000000,
+            "length":    0x1000
```
davidchisnall, line 9: This doesn't seem to match the size of the UART device that you've added (should be 52 bytes?)

HU90m: Yep, you're spot on. I'll correct that

sdk/core/loader/boot.S
HU90m: Ah thanks! I'll have a peruse of the cheriot-safe system to see how the sonata system should be doing this

davidchisnall: On SAFE, we fill in a bunch of zeroes where the vector should be, but the interrupt vector SCRs are set up by the loader.  This code is shared between the Ibex and Sail targets currently, so hopefully Sonata will do the same thing and not need software changes.

davidchisnall:
@HU90m, do you want to have a call to discuss next steps soon?

HU90m:
> @HU90m, do you want to have a call to discuss next steps soon?

That would be brilliant, yes please! I'll send you an email now.

sdk/include/platform/sunburst/platform-uart.hh
```diff
@@ -0,0 +1,70 @@
...
+	uint32_t alertTest;
+	uint32_t ctrl;
+	uint32_t status;
+	uint32_t rData;
+	uint32_t wData;
+	uint32_t fifoCtrl;
+	uint32_t fifoStatus;
+	uint32_t ovrd;
+	uint32_t val;
+	uint32_t timeoutCtrl;
```
davidchisnall, line 26: These should all have doc comments explaining what they do (which are for reading, which for writing, which bits are valid in each and what they mean, and so on).

sdk/include/platform/sunburst/platform-uart.hh
```diff
@@ -0,0 +1,70 @@
...
+	};
+
+	bool can_write() volatile
+	{
+		return (fifoStatus & 0xff) < 32;
+	};
+
+	bool can_read() volatile
+	{
+		// Not efficient, but explanatory
```
davidchisnall, line 44: I don't understand this comment.

davidchisnall (approved):
Looks fine, but needs a few more comments explaining the interface.  Ideally we'd also be able to test Sonata in CI.

HU90m:
I've just added documentation to the minimal drivers.

PR #193, opened 2024-04-04
Remove redundant mask of stack high water mark in switcher.

rmn30:
Now that the stack high water mark is explicilty 16-byte aligned we no longer need to mask it in the switcher.

See https://github.com/microsoft/cheriot-sail/pull/49

This PR is likely to fail CI until the dev container is rebuilt with the above Sail changes.


davidchisnall (approved):
Looks good to me. Rebuilding the dev container now, should make CI happy. It would be nice to have a test that did a store at some unaligned displacement from the stack and checked that it had been zeroed.

PR #199, opened 2024-04-10
Fetch ethernet receive interrupt entropy at every reseed.

hlef:
Two reasons why we likely want this:
- If we don't fetch it every time, we will permute with the same number every time, which isn't great in terms of entropy. The runtime cost shouldn't matter because we're not calling reseed often at all.
- This makes us win 8 Bytes per entropy source.

### DIFF
--- a/sdk/include/platform/arty-a7/platform-entropy.hh
+++ b/sdk/include/platform/arty-a7/platform-entropy.hh
@@ -20,7 +20,6 @@ DECLARE_AND_DEFINE_INTERRUPT_CAPABILITY(RevokerInterruptEntropy,
 class EntropySource
 {
 	ds::xoroshiro::P128R64 prng;
-	const uint32_t        *ethernetInterruptCount;
 
 	public:
 	using ValueType = uint64_t;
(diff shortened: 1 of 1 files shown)


davidchisnall:
Only the second of these is true because we’re fetching a futex, each subsequent read will read the new value.

sdk/include/platform/arty-a7/platform-entropy.hh
```diff
@@ -9,6 +9,7 @@ DECLARE_AND_DEFINE_INTERRUPT_CAPABILITY(RevokerInterruptEntropy,
                                         InterruptName::EthernetReceiveInterrupt,
                                         true,
                                         false)
+#define REVOKER_INTERRUPT_ENTROPY STATIC_SEALED_VALUE(RevokerInterruptEntropy)
```
davidchisnall, line 12: Why a define for a thing that is used only once?

hlef: I found it more readable like that. I can remove.

davidchisnall (approved):

hlef:
Rebased on top of main and fixed the format issue.

PR #201, opened 2024-04-12
Make the scheduler tickless

davidchisnall:
Timers are now not set for a regular tick, they are set when a thread
may be preempted.  Specifically, the timer is set to the next timeout
that will trigger a scheduling operation.  This avoids timers
triggering a switch to the scheduler to do nothing (resume the currently
running thread).

This means that if a thread sleeps for ten ticks while another runs, we
will get one timer interrupt ten ticks in the future, rather than ten
interrupts one tick apart.

This means that ticks are now calculated retroactively based on elapsed
time, rather than counted on each context switch.

This, in turn, necessitates some small API changes.  We previously
conflated two things:

 - Sleep for N * (tick duration)
 - Yield and allow lower-priority threads to run for, at most, N * (tick
   duration)

These are now deconflated by adding a second parameter to thread_sleep.
Most sleeps are of the second form and so this is the default.

This reduces the time taken to run the test suite on Sonata by around 
30% and in the Ibex SAFE simulator by 13%.

### DIFF
--- a/sdk/core/scheduler/main.cc
+++ b/sdk/core/scheduler/main.cc
@@ -270,15 +270,20 @@ namespace sched
 
 		ExceptionGuard g{[=]() { sched_panic(mcause, mepc, mtval); }};
 
+		bool tick = false;
 		switch (mcause)
 		{
 			// Explicit yield call
 			case MCAUSE_ECALL_MACHINE:
-				schedNeeded = true;
+			{
+				schedNeeded           = true;
+				Thread *currentThread = Thread::current_get();
+				tick = currentThread && currentThread->is_ready();
 				break;
+			}
 			case MCAUSE_INTR | MCAUSE_MTIME:
-				Timer::do_interrupt();
 				schedNeeded = true;
+				tick        = true;
 				break;
 			case MCAUSE_INTR | MCAUSE_MEXTERN:
 				schedNeeded = false;
@@ -293,6 +298,7 @@ namespace sched
 					  std::tie(schedNeeded, std::ignore, std::ignore) =
 					    futex_wake(Capability{&word}.address());
 				  });
+				tick = schedNeeded;
 				break;
 			case MCAUSE_THREAD_EXIT:
 				// Make the current thread non-runnable.
@@ -305,13 +311,23 @@ namespace sched
 				// We cannot continue exiting this thread, make sure we will
 				// pick a new one.
 				schedNeeded  = true;
+				tick         = true;
 				sealedTStack = nullptr;
 				break;
 			default:
 				sched_panic(mcause, mepc, mtval);
 		}
+		if (tick || !Thread::any_ready())
+		{
+			Timer::expiretimers();
+		}
 		auto newContext =
 		  schedNeeded ? Thread::schedule(sealedTStack) : sealedTStack;
+#if 0
+		Debug::log("Thread: {}",
+		           Thread::current_get() ? Thread::current_get()->id_get() : 0);
+#endif
+		Timer::update();
 
 		if constexpr (Accounting)
 		{
@@ -351,7 +367,7 @@ namespace sched
 
 			if (shouldYield)
 			{
-				Thread::yield_interrupt_enabled();
+				yield();
 			}
 
 			return ret;
@@ -419,14 +435,17 @@ SystickReturn __cheri_compartment("sched") thread_systemtick_get()
 }
 
 __cheriot_minimum_stack(0x80) int __cheri_compartment("sched")
-  thread_sleep(Timeout *timeout)
+  thread_sleep(Timeout *timeout, uint32_t flags)
 {
 	STACK_CHECK(0x80);
 	if (!check_timeout_pointer(timeout))
 	{
 		return -EINVAL;
 	}
-	Thread::current_get()->suspend(timeout, nullptr, true);
+	// Debug::log("Thread {} sleeping for {} ticks",
+	//  Thread::current_get()->id_get(), timeout->remaining);
+	Thread *current = Thread::current_get();
+	current->suspend(timeout, nullptr, true, !(flags & ThreadSleepNoEarlyWake));
 	return 0;
 }
 
@@ -468,8 +487,10 @@ __cheriot_minimum_stack(0xa0) int futex_timed_wait(Timeout        *timeout,
 		// If we try to block ourself, that's a mistake.
 		if ((owningThread == currentThread) || (owningThread == nullptr))
 		{
-			Debug::log("futex_timed_wait: invalid owning thread {}",
-			           owningThread);
+			Debug::log("futex_timed_wait: thread {} acquiring PI futex with "
+			           "invalid owning thread {}",
+			           currentThread->id_get(),
+			           owningThreadID);
 			return -EINVAL;
 		}
 		Debug::log("Thread {} boosting priority of {} for futex {}",
@@ -550,7 +571,7 @@ __cheriot_minimum_stack(0x90) int futex_wake(uint32_t *address, uint32_t count)
 
 	if (shouldYield)
 	{
-		Thread::yield_interrupt_enabled();
+		yield();
 	}
 
 	return woke;
--- a/sdk/core/scheduler/thread.h
+++ b/sdk/core/scheduler/thread.h
@@ -18,6 +18,8 @@ namespace
 	// thread structures.
 	class MultiWaiterInternal;
 
+	uint64_t expiry_time_for_timeout(uint32_t timeout);
+
 	template<size_t NPrios>
 	class ThreadImpl final : private utils::NoCopyNoMove
 	{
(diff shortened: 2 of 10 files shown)


sdk/core/scheduler/timer.h
```diff
@@ -28,34 +28,72 @@ namespace
...
+			}
+		}
 
-			expiretimers();
-			setnext(TIMERCYCLES_PER_TICK);
+		static uint64_t update_tick()
+		{
+			uint64_t now     = time();
+			uint32_t elapsed = now - lastTickTime;
+			int32_t  error   = elapsed % TIMERCYCLES_PER_TICK;
```
rmn30, line 69: This is worthy of some comment. I think it is calculating error by assuming elapsed should be  a multiples of TIMERCYCLES_PER_TICK and, if less than a tick has elapsed, it rounds up by one?

rmn30 (commented):
Don't understand scheduler enough to fully comment but some suggestsions.

sdk/core/scheduler/timer.h
davidchisnall: Yes, this was left over from a previous attempt to properly account the slack.  It's all largely noise now, so I'll remove it entirely.

sdk/lib/debug/debug.cc
```diff
@@ -335,7 +336,13 @@ debug_log_message_write(const char          *context,
 	DebugPrinter printer;
 	printer.write("\x1b[35m");
 	printer.write(context);
+#if 0
```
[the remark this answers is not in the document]
davidchisnall: This probably wants wiring up in an option, it's mostly disabled for now so that I can do timing checks without it there (it slows things down quite a bit with a real UART).

sdk/core/scheduler/timer.h
```diff
@@ -72,6 +117,39 @@ namespace
 					break;
 				}
 			}
+			if (!Thread::any_ready())
+			{
+				for (Thread *iter = Thread::waitingList; iter;)
```
nwf-msr, line 122: I think I'd prefer two `waitingList`-s, one for time elapsing and one for yielding, so that we wouldn't have to do this kind of two-pass rummaging and discrimination.  I suppose with our thread counts it doesn't hugely matter, but still.

sdk/core/scheduler/timer.h
```diff
@@ -72,6 +117,39 @@ namespace
...
+						iter = iterNext;
+						if (Thread::waitingList == nullptr ||
+						    iter == Thread::waitingList)
+						{
+							break;
+						}
+					}
+					else
+					{
+						break;
```
nwf-msr, line 140: I don't think that's right; I think you want to keep scanning the queue for the first yielding thread, not bail if the head of the queue isn't yielding.

sdk/include/platform/generic-riscv/platform-timer.hh
```diff
@@ -38,6 +38,21 @@ class StandardClint : private utils::NoCopyNoMove
 		         2 * sizeof(uint32_t));
 	}
 
+	static uint64_t time()
+	{
+		/// the low 32 bits
```
nwf-msr, line 43: Eh?

sdk/include/FreeRTOS-Compat/task.h
```diff
@@ -55,7 +55,7 @@ static inline BaseType_t xTaskCheckForTimeOut(TimeOut_t  *pxTimeOut,
 static inline void vTaskDelay(const TickType_t xTicksToDelay)
 {
 	struct Timeout timeout = {0, xTicksToDelay};
-	thread_sleep(&timeout);
+	thread_sleep(&timeout, 0);
```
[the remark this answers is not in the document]
davidchisnall: Ooops, I did this before I inverted the default.

sdk/core/scheduler/timer.h
davidchisnall: We're only doing the second walk when we have no runnable threads, so I don't really mind that it's taking a bit more time.

sdk/core/scheduler/timer.h
davidchisnall: This probably needs a comment.  I actually do (and therefore probably don't want to write this as a loop).  There's a fun case of:

 - Low-priority B sleeps 3 ticks.
 - High-priority A wakes, runs, and yields 5 ticks.

I think I want B to wake when it finishes sleeping, but then A to wake when B has finished.  

I'm not 100% sure about this, but there are a lot of things similar to this (e.g. B ties to acquire a lock with a 1-tick timeout in the middle of A yielding) where I don't think I would want A to wake quite yet.

sdk/include/platform/generic-riscv/platform-timer.hh
davidchisnall: Well, indeed.

sdk/core/switcher/entry.S
```diff
@@ -970,7 +970,13 @@ __Z13thread_id_getv:
 	// Load the trusted stack pointer into a register that we will clobber in
 	// the next instruction when we load the thread ID.
 	cspecialr          ca0, mtdc
+	//cgettag            a1, ca0
+	// If this is a null pointer, don't try to dereference it and report that
+	// we are thread 0.  This permits the debug code to work even from things
+	// that are not real threads.
+	//beqz               a1, .Lend
```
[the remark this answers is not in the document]
davidchisnall: No, I was testing that Debug messages from the scheduler's fake thread were the only cases where this happened and forgot to uncomment it.

sdk/include/thread.h
```diff
@@ -23,6 +23,17 @@ typedef struct
 [[cheri::interrupt_state(disabled)]] SystickReturn __cheri_compartment("sched")
   thread_systemtick_get(void);
 
+enum ThreadSleepFlags : uint32_t
+{
+	/**
+	 * Sleep for up to the specified timeout, but wake early if there are no
+	 * other runnable threads.  This allows a high-priority thread to yield for
```
nwf-msr, line 30:
```suggestion
	 * other threads runnable or blocked with earlier timeouts.
	 * This allows a high-priority thread to yield for
```

I'm trying to capture the difference between "wake as soon as the system idles within the next N ticks" (which would be the "traverse the waiting queue until we found a yielding thread" behavior) and "wake as soon as the system looks like it might be idle N ticks from now" (the "only test the head of the waiting queue for yielding" behavior now implemented).

sdk/core/scheduler/main.cc
```diff
@@ -419,14 +435,17 @@ SystickReturn __cheri_compartment("sched") thread_systemtick_get()
...
+  thread_sleep(Timeout *timeout, uint32_t flags)
 {
 	STACK_CHECK(0x80);
 	if (!check_timeout_pointer(timeout))
 	{
 		return -EINVAL;
 	}
-	Thread::current_get()->suspend(timeout, nullptr, true);
+	////Debug::log("Thread {} sleeping for {} ticks",
+	/// Thread::current_get()->id_get(), timeout->remaining);
```
nwf-msr, line 446: Nit: could you `ifdef 0` this or cut down the number of slashes so that it doesn't look like a doxygen-style comment?

sdk/core/scheduler/thread.h
```diff
@@ -616,7 +647,12 @@ namespace
 		uint8_t priority;
 		/// The original priority level for this thread.  This never changes.
 		const uint8_t OriginalPriority;
-		ThreadState   state;
+		ThreadState   state : 2;
+		/**
+		 * If the thread is yielding, it may be scheduled before its timeout
+		 * expires, as long as no other threads are runnable.
```
nwf-msr, line 653: I might leave off the "as long as..." part here and just point at the docs for `ThreadSleepNoEarlyWake`?

davidchisnall: I think it's fine to have duplicated explanations, especially in doc comments.  If you mouse over a use of this field in VS Code (for example), you'll get this doc comment pop up.  Having it tell you to go and read other docs can be annoying.

sdk/core/scheduler/main.cc
davidchisnall: Not sure where the bonus slashes came from.

sdk/include/thread.h
davidchisnall: I've added some more explanation.

sdk/core/scheduler/timer.h
```diff
@@ -26,36 +26,96 @@ namespace
...
+				                       : Thread::waitingList->expiryTime;
+				setnext(nextTimer);
+			}
 		}
 
-		private:
+		/**
+		 * Wake any threads that were sleeping until a timeout before the
+		 * current time.  This also wakes yielded threads if there are no
+		 * runnable threads.
```
nwf-msr, line 103:
```suggestion
		 * threads runnable or sleeping with earlier timeouts.
```

sdk/core/scheduler/thread.h
nwf-msr: That's fair.  In which case...
```suggestion
		 * expires, as long as no other threads are runnable or
		 * sleeping with shorter timeouts.
```

nwf-msr (approved):
Two nits, but it LGTM.

sdk/core/scheduler/timer.h
```diff
@@ -26,36 +26,96 @@ namespace
...
+                                       (Thread::waitingList->expiryTime == -1));
+			bool  threadHasNoPeers =
+			  (thread == nullptr) || (!thread->has_priority_peers());
+			if (waitingListIsEmpty && threadHasNoPeers)
+			{
+				clear();
+			}
+			else
+			{
+				uint64_t nextTimer = waitingListIsEmpty
```
rmn30, line 93: @davidchisnall if the waiting list is non-empty *and* the current thread has peers I think this will set the timer for the next expiry time. Doesn't that mean the current thread could get more than one quantum?

davidchisnall: Fixed in #240, thanks!

PR #205, opened 2024-04-18
Add APIs for checking stack usage.

davidchisnall:
This makes it possible to dynamically check the maximum stack usage for a compartment entry point in testing.  This is intended to be used in conjunction with CHERIoT-Platform/llvm-project#22, which makes it possible to specify the minimum stack size (enforced by the switcher).

### DIFF
--- a/sdk/include/debug.hh
+++ b/sdk/include/debug.hh
@@ -2,13 +2,13 @@
 // SPDX-License-Identifier: MIT
 
 #pragma once
-#include "cdefs.h"
 #include <cheri.hh>
 #include <compartment.h>
 #include <concepts>
 #include <cstddef>
 #include <platform-uart.hh>
 #include <string.h>
+#include <switcher.h>
 
 #include <array>
 #include <string_view>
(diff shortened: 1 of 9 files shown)


sdk/include/switcher.h
```diff
@@ -49,3 +50,9 @@ __cheri_libcall _Bool switcher_interrupt_thread(void *);
  * next cross-compartment call or until they are explicitly overwritten.
  */
 __cheri_libcall void **switcher_thread_hazard_slots(void);
+
+/**
+ * Returns the lowest address that has been stored to on the stack in this
+ * compartment invocation.
```
rmn30, line 56: May be worth noting that this will be rounded down to 16-byte alignment?

sdk/include/debug.hh
```diff
@@ -741,4 +741,59 @@ namespace
...
+			if constexpr (Enabled)
+			{
+				ptraddr_t lowest = stack_lowest_used_address();
+				ptraddr_t highest =
+				  CHERI::Capability{__builtin_cheri_stack_get()}.top();
+				size_t used = highest - lowest;
+				if (used > stackUsage)
+				{
+					stackUsage = used;
+					ConditionalDebug<Enabled, Fn>::log("Stack used: {} bytes",
```
rmn30, line 792:
```suggestion
					ConditionalDebug<Enabled, Fn>::log("Stack use exceeded 
					{}: {} bytes", Expected,
```

rmn30 (approved):

sdk/include/debug.hh
davidchisnall: Why? This is printing the amount used.

rmn30: Just thought it might be useful to have both the expected and actual stack usage.

sdk/include/platform/arty-a7/platform-entropy.hh
```diff
@@ -5,7 +5,7 @@
 #include <platform/concepts/entropy.h>
 #include <riscvreg.h>
 
-DECLARE_AND_DEFINE_INTERRUPT_CAPABILITY(RevokerInterruptEntropy,
```
rmn30, line 8: This looks unrelated to other changes?

davidchisnall: Yup, accidentally committed and reverted.

sdk/include/debug.hh
```diff
@@ -741,4 +741,89 @@ namespace
...
+	 * callees.  This is intended to be used in compartment entry points to
+	 * check the stack size that is required for that entry point.  Use this
+	 * with some test vectors that explore different code paths to identify the
+	 * maximum stack usage.  This can then be used with the
+	 * [[cheriot::minimum_stack]] attribute to ensure that the switcher will
+	 * never invoke the entry point with insufficient stack.
+	 *
+	 * Note that this records only the stack usage for the current compartment
+	 * invocation (including any invoked shared libraries).  If this
+	 * compartment calls others, then smaller stacks may fail.  This failure is
```
nwf-msr, line 762: I'm not quite sure what this sentence is saying.  Do you mean "If this compartment calls others, then even stacks larger than this recorded threshold may fail"?

sdk/include/debug.hh
```diff
@@ -741,4 +741,89 @@ namespace
...
+	 * ```
+	 */
+	template<StackCheckMode Mode, size_t Expected, DebugContext Fn>
+	class StackUsageCheck
+	{
+		/**
+		 * The expected maximum.  This class is templated on the function name
+		 * and so there is one copy of this per function.
+		 */
+		static inline size_t stackUsage = Expected;
```
nwf-msr, line 796: Does this still get created even if the mode is `Disabled`?  I think maybe the kind of `std::conditional_t<>`... "stunts" we pull in `snmalloc`'s freelist guard objects could avoid that?

sdk/xmake.lua
```diff
@@ -39,6 +39,17 @@ debugOption("scheduler")
 debugOption("allocator")
 debugOption("token_library")
 
+function stackCheckOption(name)
+	option("stack-usage-check-" .. name)
+		set_default(false)
+		set_description("Enable dynamic stack usage checks in " .. name .. ". Do not enable this in debug builds!")
```
nwf-msr, line 45: Hm.  I'm not sure how I feel about the "do not enable in debug builds" bit.  I presume that's because the constants in the source are the release build stack usage and debug messages dramatically increase the stack usage.  Maybe debug builds should force any `Asserting` mode to be `Logging` instead?

tests/allocator-test.cc
```diff
@@ -542,7 +543,18 @@ void test_allocator()
...
+	t = 5;
+	Capability array{heap_allocate_array(&t, MALLOC_CAPABILITY, 0x80000004, 2)};
+	TEST(!array.is_valid(), "Allocating too large an array succeeded: {}", array);
+	array = heap_allocate_array(&t, MALLOC_CAPABILITY, 16, 2);
+	TEST(array.is_valid(), "Allocating array failed: {}", array);
+	TEST(array.length() == 32, "Allocating array returned incorrect length: {}", array);
+	ret = heap_free(MALLOC_CAPABILITY, array);
+	TEST(ret == 0, "Freeing array failed: {}", ret);
+
+
```
nwf-msr, line 555: This hunk looks unrelated, in that I don't see it fiddling with the stack bounds?

sdk/include/debug.hh
davidchisnall: It's `inline`, which means it's created in every compilation unit where it's referenced.  With `Disabled`, it isn't referenced and so doesn't exist.

sdk/xmake.lua
davidchisnall: There's no value to running it in debug builds because they will have larger stack usage and so you'll bump your requirements to something wrong.  The only exceptions to that rule are people that really understand what they're doing and know that they can ignore the advice and, importantly, why.

sdk/include/debug.hh
davidchisnall: Clarified.

PR #206, opened 2024-04-24
Make it possible to upgrade a lock for destruction without holding it.

hlef:
The ability to upgrade for destruction without holding the lock is useful for cleaning up from the error handler.

Add various tests to detect regressions in the `unlock` function.

### DIFF
--- a/tests/locks-test.cc
+++ b/tests/locks-test.cc
@@ -23,14 +23,12 @@ namespace
 	cheriot::atomic<int>  counter;
 
 	/**
-	 * Test that a lock meets the minimum requirements: it actually provides
-	 * mutual exclusion.
+	 * Test that a lock actually provides mutual exclusion.
 	 */
 	template<typename Lock>
 	void test_lock(Lock &lock)
 	{
 		modified = false;
-		debug_log("Acquiring lock in {}", __PRETTY_FUNCTION__);
 		{
 			LockGuard g{lock};
 			async([&]() {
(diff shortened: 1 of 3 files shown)


tests/locks-test.cc
```diff
@@ -211,6 +262,7 @@ namespace
 	 */
 	void test_ticket_lock_ordering()
 	{
+		counter = 0;
```
davidchisnall, line 265: What is this?

davidchisnall:
I think this looks good.  I wonder if it would be sensible to add a `force` parameter to the upgrade function.  Any graceful destruction should involve an upgrade from an owned lock, it's only destruction from an error handler that should require upgrading a non-owned lock.

hlef:
> I think this looks good. I wonder if it would be sensible to add a `force` parameter to the upgrade function. Any graceful destruction should involve an upgrade from an owned lock, it's only destruction from an error handler that should require upgrading a non-owned lock.

I agree that it's only destruction from an error handler that should require upgrading a non-owned lock. However I'm not sure what the `force` parameter would do?

Currently the upgrade function does not acquire the lock - it expects the lock to be already held. In this commit we are simply saying that the lock does not need to be held to call the function (and making sure that no races are possible with other functions of the API).

I do not think that it would make sense to acquire the lock *in* the upgrade function, as in the normal case you would only call upgrade on a lock that is already held.

davidchisnall:
Huh, I thought that priority-inherited mutexes checked that they were owned by the calling thread before upgrade.  It looks like they didn't, so I think this is fine.

davidchisnall (approved):

PR #211, opened 2024-05-01
Fix an assertion in the loader.

davidchisnall:
This assertion should have fired previously when we introduced the hazard pointer region for `heap_claim_fast` but it did not because we have devices before and after main SRAM.  On targets where this is not the case, it fires spuriously.

### DIFF
--- a/sdk/core/loader/boot.cc
+++ b/sdk/core/loader/boot.cc
@@ -464,13 +464,29 @@ namespace
 					}
 				}
 			}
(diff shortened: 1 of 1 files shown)


davidchisnall:
@HU90m, can you test this?

HU90m:
Brill! I'll test it now

HU90m:
This does the trick. I'm able to get to the scheduler. Thank you

PR #212, opened 2024-05-01
Add a `eventgroup_destroy_force` API.

hlef:
This new API destroys an event group without tacking the lock.

This API is inherently racy. Its main purpose is to cleanup the event group in an error handler context, when taking lock may be impossible.

While we are at it, all `eventgroup_destroy` operations, forced or not, should upgrade the lock for destruction before freeing it.

We could also add a `force` parameter to the existing `eventgroup_destroy` API, however we would need to go and edit all callers since this is a C API.

### DIFF
--- a/sdk/include/event.h
+++ b/sdk/include/event.h
@@ -108,3 +108,12 @@ int __cheri_libcall eventgroup_get(struct EventGroup *group, uint32_t *outBits);
  */
 int __cheri_libcall eventgroup_destroy(struct SObjStruct *heapCapability,
                                        struct EventGroup *group);
+
+/**
+ * Destroy an event group without tacking the lock.
+ *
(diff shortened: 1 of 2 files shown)


sdk/include/event.h
```diff
@@ -108,3 +108,12 @@ int __cheri_libcall eventgroup_get(struct EventGroup *group, uint32_t *outBits);
  */
 int __cheri_libcall eventgroup_destroy(struct SObjStruct *heapCapability,
                                        struct EventGroup *group);
+
+/**
+ * Destroy an event group without tacking the lock.
+ *
+ * This API is inherently racy. Its main purpose is to cleanup the event group
```
davidchisnall, line 115: Cleanup is a noun, you mean clean up.

davidchisnall (approved):

PR #220, opened 2024-05-13
Added user switches and joystick to Sonata GPIO driver

HU90m:


davidchisnall (commented):
Sorry, I reviewed this a week ago and then failed to hit the 'submit review' button!

davidchisnall (approved):

PR #222, opened 2024-05-15
Various fixes related to memory allocator error handling

hlef:
We have recently introduced strong stack size requirements in the allocator (https://github.com/microsoft/cheriot-rtos/pull/205). Following this, `-ECOMPARTMENTFAIL` is returned not only when the allocator compartment crashes, but also at compartment entry when the stack is insufficient to proceed safely.

Entering a state where the stack is not sufficient to call the allocator is not unrealistic given that we are running with very small stacks. We need to handle it well.

Unfortunately this behavior has not yet been documented, and the code has not been adapted to it. Nearly the entire core repository still does `NULL` pointer checks to check the return value of a memory allocation.

This commit documents this new behavior at appropriate places and fixes allocator call sites to ensure that the valid bit is checked instead of a comparison with a `NULL` pointer.

### DIFF
--- a/sdk/core/scheduler/common.h
+++ b/sdk/core/scheduler/common.h
@@ -247,6 +247,10 @@ namespace
 		HeapObject(struct SObjStruct *heapCapability, T *allocatedObject)
 		  : pointer(allocatedObject, heapCapability)
 		{
+			if (!__builtin_cheri_tag_get(allocatedObject))
+			{
+				pointer = nullptr;
+			}
 		}
 
 		/**
@@ -264,9 +268,13 @@ namespace
 		              heap_allocate(timeout, heapCapability, sizeof(T))),
 		            {heapCapability})
 		{
-			if (pointer)
+			if (__builtin_cheri_tag_get(pointer.raw()))
 			{
 				new (pointer.get()) T(std::forward<Args>(args)...);
+			}
+			else
+			{
+				pointer = nullptr;
 			}
 		}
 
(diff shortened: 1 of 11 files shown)


hlef:
@davidchisnall one thing which does not satisfy me here is that we make heavy use of  `__builtin_cheri_tag_get`. It does not feel like this built in should be used like that. Unfortunately this seems to be the only C-only API option right now.

Should we introduce a C wrapper for this builtin?

hlef:
PR updated to reflect changes of https://github.com/microsoft/cheriot-rtos/commit/3e0eaf0926c779d0b6ce47fac05786281fa4c01a.

sdk/core/scheduler/common.h
```diff
@@ -247,6 +247,10 @@ namespace
 		HeapObject(struct SObjStruct *heapCapability, T *allocatedObject)
 		  : pointer(allocatedObject, heapCapability)
 		{
+			if (__builtin_cheri_tag_get(allocatedObject) == 0)
```
davidchisnall, line 250: This built in should return a bool, so comparing it against zero is confusing.

hlef: Sounds good. I just replicated what I saw there: https://github.com/microsoft/cheriot-rtos/blob/main/tests/allocator-test.cc#L87

Should I fix this as well?

davidchisnall: Yes please!

hlef:
PR updated to check boolean return value for `__builtin_cheri_tag_get`.

sdk/core/scheduler/common.h
```diff
@@ -247,6 +247,10 @@ namespace
 		HeapObject(struct SObjStruct *heapCapability, T *allocatedObject)
 		  : pointer(allocatedObject, heapCapability)
 		{
+			if (__builtin_cheri_tag_get(allocatedObject) == false)
```
davidchisnall, line 250: Why not just ! __builtin_cheri_tag_get(allocatedObject)?

hlef: No particular reason, I can update. I do not have a strong preference.

PR #224, opened 2024-05-16
Add missing stack check.

davidchisnall:
The ABI has included space for the required stack space from the start, but we only recently started putting anything sensible there.  It turns out that the switcher was missing this check.

This introduces a new errno code to let the caller differentiate between a call that wasn't allowed to proceed and one that crashed.  We may want to use this code for other stack badness.

### DIFF
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -3,6 +3,7 @@
 
 #include "export-table-assembly.h"
 #include "trusted-stack-assembly.h"
+#include <errno.h>
 
 .include "assembly-helpers.s"
 
@@ -246,6 +247,17 @@ after_zero:
 	// table pointer.  Nothing between this point and transition to the callee
 	// should fault.
 	csc                ct1, TrustedStackFrame_offset_calleeExportTable(ctp)
+
+	// Load the minimum stack size required by the callee.
+	clbu               tp, ExportEntry_offset_minimumStackSize(ct1)
+	// The stack size is in 8-byte units, so multiply by 8.
+	slli               tp, tp, 3
+	// Check that the stack is large enough for the callee.
+	// At this point, we have already truncated the stack and so the length of
+	// the stack is the length that the callee can use.
+	cgetlen            t2, csp
+	bgtu               tp, t2, .Lstack_too_small
+
 	// Get the flags field into tp
 	clbu               tp, ExportEntry_offset_flags(ct1)
 	cgetbase           s1, ct1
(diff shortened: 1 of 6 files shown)


nwf (approved):
LGTM if it passes testing (not on work machine, so that doesn't count as a review, bah)

sdk/core/switcher/entry.S
```diff
@@ -302,6 +314,16 @@ after_zero:
...
 	zeroAllRegistersExcept ra, sp, gp, s0, s1, a0, a1
 	cret
+
+	// If the stack is too small, we don't do the call, but to avoid leaking
+	// any other state we still go through the same return path as normal.  We
+	// set the return registers to -ENOTENOUGHSTACK and 0, so users can see
+	// that this is the failure reason.
+.Lstack_too_small:
+	// Return register is -ENOTENOUGHSTACK
+	li                 a0, -140
```
rmn30, line 324: Why can't we use the macro here?

tests/test-runner.cc
```diff
@@ -104,6 +107,7 @@ void __cheri_compartment("test_runner") run_tests()
 	debug_log("Trying to print string: {}", std::string_view{testString, 13});
 
 	run_timed("All tests", []() {
+		run_timed("Stacks exhaustion in the switcher", test_stack);
```
rmn30, line 110: I guess this was left in from testing as it also appears below?
```suggestion
```

tests/stack-test.cc
```diff
@@ -72,8 +72,42 @@ namespace
...
+
+asm(".section .text\n"
+    ".global test_with_small_stack\n"
+    "test_with_small_stack:\n"
+    // Preserve the old stack
+    "  cmove ct0, csp\n"
+    // Add space to the requested size for the spill slots and the size of the
+    // stack that `test_small_stack` will use, plus the four capabilities that
+    // the switcher will spill for us.
+    "  add a0, a0, 64\n"
```
rmn30, line 92: I guess this is 2 x spills here, 2 x spills in test_small_stack and 4 x switcher spills? Would be good to be explicit. Are the spills in `test_small_stack` dependent on anything like debug level? Concerned the magic number may be a bit brittle.

rmn30 (approved):

sdk/core/switcher/entry.S
davidchisnall: We probably could.  I wasn't sure about including errno.h, but I think it should be fine in hindsight.

tests/stack-test.cc
davidchisnall: The magic number *is* brittle, but the stack must be 16-byte aligned and we check 16 bytes either side of the line, so we'll at least get a failure that someone can debug if one of the assumptions changes.  I'll improve the comment.

PR #227, opened 2024-05-21
Fix Sonata's memory layout.

davidchisnall:
Sonata starts the revokable memory range (the range that the shadow memory covers) 0x80 bytes before the start of code memory, but the board description file did not mention this.


marnovandermaas (approved):
Yes, this aligns with the current Sonata RTL. The base of revocable memory starts from the base of SRAM.

PR #229, opened 2024-05-24
Fix epoch calculation in the allocator.

davidchisnall:
The allocator tracks complete (even) and partial (odd) epochs and puts things in a free list depending on the complete epoch that will allow them to be popped.  When we poll the revoker, this means that the current epoch may be a partial epoch one *before* the current epoch.

Unfortunately, we were doing this check with unsigned arithmetic and so it was possible to remove things from quarantine too early.


nwf-msr (approved):
LGTM; sorry for the bug!

PR #233, opened 2024-05-27
Changed the Sonata instruction memory start

HU90m:
There is now a bootloader on that sonata platform that loads software to 0x00101000.


davidchisnall:
We can also lose the revocable memory start entry.

Do I need a new bitfile to test this?

davidchisnall (approved):

PR #234, opened 2024-05-27
Added Sonata simple SPI driver.

HU90m:


sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,92 @@
...
+
+	// Waits for the SPI device to become idle
+	void wait_idle() volatile
+	{
+		// Wait whilst IDLE field in STATUS is low
+		while ((status & 0x40000) == 0) {}
+	}
+
+	// Sends `len` bytes from the given `data` buffer.
+	void tx(const uint8_t data[], uint32_t len) volatile
```
HU90m, line 52: It would be nice to use `std::span<uint8_t` but we don't have that in the CHERIoT standard library, so I resorted to simple C arrays.

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,92 @@
...
+ * Documentation source can be found at:
+ * https://github.com/lowRISC/sonata-system/blob/1a59633d2515d4fe186a07d53e49ff95c18d9bbf/doc/ip/spi.md
+ *
+ * Rendered documentation is served from:
+ * https://lowrisc.org/sonata-system/doc/ip/spi.html
+ */
+struct SonataSpi
+{
+	// Interrupt State Register
+	uint32_t intrState;
```
davidchisnall, line 17: Please use full words for variable and function names. Our coding guidelines to avoid abbreviations. You know what intr means but the next person looking at this file or seeing it pop up in autocomplete may not.

Same for most of the other fields and methods here.

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * Documentation source can be found at:
+ * https://github.com/lowRISC/sonata-system/blob/1a59633d2515d4fe186a07d53e49ff95c18d9bbf/doc/ip/spi.md
+ *
+ * Rendered documentation is served from:
+ * https://lowrisc.org/sonata-system/doc/ip/spi.html
+ */
+struct SonataSpi
+{
+	// Interrupt State Register
```
davidchisnall, line 16: Please make these doc comments (three slashes).

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,92 @@
...
+	uint32_t rxFifo;
+	// Bytes written here are pushed to the transmit FIFO. If the FIFO
+	uint32_t txFifo;
+
+
+	// Initialises the SPI block
+	void
+	init(bool cpol, bool cpha, bool msbFirst, uint16_t halfClkPeriod) volatile
+	{
+		cfg = (cpol ? 1 << 31 : 0) | (cpha ? 1 << 30 : 0) |
```
davidchisnall, line 40: What are all of these magic numbers?

sdk/include/platform/sunburst/platform-spi.hh
davidchisnall: That’s fine (though please name these ‘transmit’ and ‘receive’ or similar). We probably should also have a concept covering this interface.

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,92 @@
...
+				}
+			}
+
+			txFifo = data[i];
+			txAvail--;
+		}
+	}
+
+	// Receives `len` bytes and puts them in the `data` buffer.
+	void rx(uint8_t data[], uint32_t len) volatile
```
davidchisnall, line 78: What is the behaviour of this if there are not the required number of bytes?

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,92 @@
...
+	{
+		wait_idle();
+		// Set RX_ENABLE
+		control = 0x8;
+		start   = len;
+
+		for (int i = 0; i < len; ++i)
+		{
+			// Wait for at least one byte to be available in the RX FIFO
+			while (((status >> 8) & 0xff) == 0) {}
```
davidchisnall, line 88: I don’t really like this spinning without providing an opportunity to yield. Is there not an interrupt that this can block on?

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,92 @@
+#pragma once
+#include <cdefs.h>
+#include <stdint.h>
+
+/**
+ * A Simple Driver for the Sonata's GPIO.
```
nwf, line 6: Copy-paste-o
```suggestion
 * A Simple Driver for the Sonata's SPI.
```

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,171 @@
...
+
+	/// Start Register Fields
+	enum : uint32_t
+	{
+		/// Number of bytes to receive/transmit in the SPI operation
+		StartByteCountMask = 0x7ffu,
+	};
+
+
+	/// Initialises the SPI block
```
HU90m, line 109: Should I repeat or more the explanation of the Configuration Register Fields in `init()`'s doc comment?

sdk/include/platform/sunburst/platform-spi.hh
HU90m: I had followed the naming of [the hardware documentation](https://lowrisc.org/sonata-system/doc/ip/spi.html), but I agree renaming to be more explanatory would probably be better.

sdk/include/platform/sunburst/platform-spi.hh
HU90m: Ah yes, sorry. This wasn't in a good state to PR.

sdk/include/platform/sunburst/platform-spi.hh
HU90m: There aren't interrupts for the SPI peripheral yet I'm afraid

sdk/include/platform/sunburst/platform-spi.hh
davidchisnall: If the code below can use the symbolic constants then that's fine.

I would like the doc comment t explain what the arguments mean.

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,175 @@
...
+	{
+		// Wait whilst IDLE field in STATUS is low
+		while ((status & StatusIdle) == 0) {}
+	}
+
+	/// Sends `Len` bytes from the given `data` buffer,
+	/// where `Len` is at most `0x7ff`.
+	void transmit(const uint8_t data[], const uint16_t Len) volatile
+	{
+		const uint32_t LenTruncated = Len & StartByteCountMask;
```
davidchisnall, line 132: Maybe a `Debug::Assert` that `Len <= 0x7ff`?

sdk/include/platform/sunburst/platform-spi.hh
davidchisnall: In that case, can we name this method something like `blocking_receive` (or `blocking_read`, for consistency with the UART interface)?  And maybe have a separate method to query whether there's data pending so that people can avoid blocking?

If we can't wait on an interrupt, it would be nice to put a one-tick `thread_sleep` in the wait loop (possibly after a small number of iterations?) so that we're not blocking the entire core until the timer interrupt fires if there's nothing to do.

sdk/include/platform/sunburst/platform-spi.hh
HU90m: Ah yes, good call

sdk/include/platform/sunburst/platform-spi.hh
HU90m: Yes, that sounds a lot nicer

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,175 @@
...
+ */
+struct SonataSpi
+{
+	/// Interrupt State Register
+	uint32_t interruptState;
+	/// Interrupt Enable Register
+	uint32_t interruptEnable;
+	/// Interrupt Test Register
+	uint32_t interruptTest;
+	/// Configuration register. Controls how the SPI block transmits and
```
davidchisnall, line 22: Minor nit, but we normally use `/** */` blocks for multi-line doc comments.

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,175 @@
...
+	/// receives data. This register can only be modified whilst the SPI block
+	/// is idle.
+	uint32_t configuration;
+	/// Controls the operation of the SPI block. This register can
+	/// only be modified whilst the SPI block is idle.
```
davidchisnall, line 27:
```suggestion
	/// be modified only whilst the SPI block is idle.
```

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,175 @@
...
+	uint32_t control;
+	/// Status information about the SPI block
+	uint32_t status;
+	/// When written begins an SPI operation.
```
davidchisnall, line 31: This took a few attempts for me to parse.  Maybe 'Writes to this begin an SPI operation'?

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,175 @@
...
+
+	/// Configuration Register Fields
+	enum : uint32_t
+	{
+		/// The length of a half period (i.e. positive edge to negative edge) of
+		/// the SPI clock, measured in system clock cycles reduced by 1. At the
+		/// standard Sonata 50 MHz system clock a value of 0 gives a 25 MHz SPI
+		/// clock, a value of 1 gives a 12.5 MHz SPI clock, a value of 2 gives
+		/// a 8.33 MHz SPI clock and so on.
+		ConfigurationHalfClkPeriodMask = 0xffu << 0,
```
davidchisnall, line 49:
```suggestion
		ConfigurationHalfClockPeriodMask = 0xffu << 0,
```

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,175 @@
...
+		/// When set the most significant bit (MSB) is the first bit sent and
+		/// received with each byte
+		ConfigurationMSBFirst = 1u << 29,
+		/// The phase of the spi_clk signal. When CPHA is 0 data is sampled on
```
davidchisnall, line 53: What is CPHA?

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,175 @@
...
+		/// the leading edge and changes on the trailing edge. The first data
+		/// bit is immediately available before the first leading edge of the
+		/// clock when transmission begins. When CPHA is 1 data is sampled on
+		/// the trailing edge and change on the leading edge.
+		ConfigurationCPHA = 1u << 30,
+		/// The polarity of the spi_clk signal. When CPOL is 0 clock is low when
```
davidchisnall, line 59: What is CPOL?

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,175 @@
...
+		/// idle and the leading edge is positive. When CPOL is 1 clock is high
+		/// when idle and the leading edge is negative
+		ConfigurationCPOL = 1u << 31,
+	};
+
+	/// Control Register Fields
+	enum : uint32_t
+	{
+		/// Write 1 to clear the transmit FIFO.
+		ControlTxClear = 1 << 0,
```
davidchisnall, line 69:
```suggestion
		ControlTransmitClear = 1 << 0,
```

The abbreviations tx and rx are ubiquitous in communication literature, but not everyone reading this code will have that background.

Same comment for the next ones down.

HU90m:
^ Just a rebase

HU90m:
^ I think I've addressed all the suggestions

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,227 @@
...
+ *
+ * Rendered documentation is served from:
+ * https://lowrisc.org/sonata-system/doc/ip/spi.html
+ */
+struct SonataSpi
+{
+	/// Interrupt State Register
+	uint32_t interruptState;
+	/// Interrupt Enable Register
+	uint32_t interruptEnable;
```
davidchisnall, line 20: It looks as if we're missing the interrupt definitions in the board JSON?

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,227 @@
...
+	/// Interrupt Test Register
+	uint32_t interruptTest;
+	/**
+	 * Configuration register. Controls how the SPI block transmits and
+	 * receives data. This register can only be modified whilst the SPI block
```
davidchisnall, line 25:
```suggestion
	 * receives data. This register can be modified only whilst the SPI block
```

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,227 @@
...
+	 */
+	uint32_t receiveFifo;
+	/**
+	 * Bytes written here are pushed to the transmit FIFO. If the FIFO is full
+	 * writes are ignored.
+	 */
+	uint32_t transmitFifo;
+
+	/// Configuration Register Fields
+	enum : uint32_t
```
davidchisnall, line 53: It would be nice to add a `clang::flag_enum` attribute here.

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,227 @@
...
+		/*
+		 * The polarity of the spi_clk signal. When ClockPolarity is 0 clock is
+		 * low when idle and the leading edge is positive. When ClkPolarity is 1
+		 * clock is high when idle and the leading edge is negative
+		 */
+		ConfigurationClockPolarity = 1u << 31,
+	};
+
+	/// Control Register Fields
+	enum : uint32_t
```
davidchisnall, line 85: Same here (`clang::flag_enum`)

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,227 @@
...
+	 */
+	using Debug = ConditionalDebug<DebugSonataSpi, "Sonata SPI">;
+
+	/**
+	 * Initialises the SPI block
+	 */
+	void init(const bool     ClockPolarity,
+	          const bool     ClockPhase,
+	          const bool     MsbFirst,
+	          const uint16_t HalfClkPeriod) volatile
```
davidchisnall, line 158:
```suggestion
	          const uint16_t HalfClockPeriod) volatile
```

Can the doc comment explain the parameters too?  This is the thing that pops up when you type this in an IDE, so it's nice if it can be self-contained.

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,227 @@
...
+	{
+		// Wait whilst IDLE field in STATUS is low
+		while ((status & StatusIdle) == 0) {}
+	}
+
+	/**
+	 * Sends `Len` bytes from the given `data` buffer,
+	 * where `Len` is at most `0x7ff`.
+	 */
+	void transmit(const uint8_t data[], const uint16_t Len) volatile
```
davidchisnall, line 177: I'm not really a fan of `const` parameters that are value types.  `const` on a parameter is a contract not to mutate the object, but when mutation is not observable in the caller then it's not helpful, and making this `const` means that we then have a new `LenTruncated` rather than just having a `length` parameter that we mask.

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	/*
+	 * Receives `Len` bytes and puts them in the `data` buffer,
+	 * where `Len` is at most `0x7ff`.
+	 *
+	 * This method will block until the requested number of bytes
+	 * has been seen. There is currently no timeout.
+	 */
+	void receive(uint8_t data[], const uint16_t Len) volatile
```
davidchisnall, line 211: Same comment here.

davidchisnall (approved):

sdk/include/platform/sunburst/platform-spi.hh
HU90m: I checked whether the interrupts are actually hooked up in the RTL, and I haven't tested them if they are. I'm apprehensive about adding them to the board description file until I have.

sdk/include/platform/sunburst/platform-spi.hh
HU90m: `clang::flag_enum` only appears to allow single bit fields, so it doesn't like some of the register fields being bit masks.

sdk/include/platform/sunburst/platform-spi.hh
davidchisnall: Might be nice to have a comment explaining this as a FIXME?

HU90m: Just checked the RTL; the SPI block currently doesn't have interrupts.

I have added a comment saying that these registers are reserved for future use.

HU90m:
I think that's everything

davidchisnall:
Rebased on main, auto-merge set.

PR #239, opened 2024-06-06
Added an OpenTitan I2C driver

HU90m:
This has been tested with the examples in https://github.com/lowRISC/sonata-software

### DIFF
--- a/sdk/include/platform/sunburst/platform-i2c.hh
+++ b/sdk/include/platform/sunburst/platform-i2c.hh
@@ -1,0 +1,421 @@
+#pragma once
+#include <cdefs.h>
+#include <debug.hh>
+#include <stdint.h>
+
+/**
+ * The interrupts of the OpenTitan's I2C block.
+ *
+ * Documentation source can be found at:
+ * https://github.com/lowRISC/opentitan/blob/9ddf276c64e2974ed8e528e8b2feb00b977861de/hw/ip/i2c/doc/interfaces.md
+ */
+enum class OpenTitanI2cInterrupt
+{
+	/**
+	 * A host mode interrupt. This is asserted whilst the Format FIFO level is
+	 * below the low threshold. This is a level status interrupt.
+	 */
+	FormatThreshold,
+	/**
+	 * A host mode interrupt. This is asserted whilst the Receive FIFO level is
+	 * above the high threshold. This is a level status interrupt.
+	 */
+	ReceiveThreshold,
+	/**
+	 * A target mode interrupt. This is asserted whilst the Aquired FIFO level
+	 * is above the high threshold. This is a level status interrupt.
+	 */
+	AcquiredThreshold,
+	/**
+	 * A host mode interrupt. This is raised if the Receive FIFO has overflowed.
+	 */
+	ReceiveOverflow,
+	/**
+	 * A host mode interrupt. This is raised if there is no ACK in response to
+	 * an address or data.
+	 */
+	Nak,
+	/**
+	 * A host mode interrupt. This is raised if the SCL line drops early (not
+	 * supported without clock synchronization).
+	 */
+	SclInterference,
+	/**
+	 * A host mode interrupt. This is raised if the SDA line goes low when host
+	 * is trying to assert high.
+	 */
+	SdaInterference,
+	/**
+	 * A host mode interrupt. This is raised if target stretches the clock
+	 * beyond the allowed timeout period.
+	 */
+	StretchTimeout,
+	/**
+	 * A host mode interrupt. This is raised if the target does not assert a
+	 * constant value of SDA during transmission.
+	 */
+	SdaUnstable,
+	/**
+	 * A host and target mode interrupt. In host mode, raised if the host issues
+	 * a repeated START or terminates the transaction by issuing STOP. In target
+	 * mode, raised if the external host issues a STOP or repeated START.
+	 */
+	CommandComplete,
+	/**
+	 * A target mode interrupt. This is raised if the target is stretching
+	 * clocks for a read command. This is a level status interrupt.
+	 */
+	TransmitStretch,
+	/**
+	 * A target mode interrupt. This is asserted whilst the Transmit FIFO level
+	 * is below the low threshold. This is a level status interrupt.
+	 */
+	TransmitThreshold,
+	/**
+	 * A target mode interrupt. This is raised if the target is stretching
+	 * clocks due to full Aquired FIFO or zero count in targetAckControl.NBYTES
+	 * (if enabled). This is a level status interrupt.
+	 */
+	AcquiredFull,
+	/**
+	 * A target mode interrupt. This is raised if STOP is received without a
+	 * preceding NACK during an external host read.
+	 */
+	UnexpectedStop,
+	/**
+	 * A target mode interrupt. This is raised if the host stops sending the
+	 * clock during an ongoing transaction.
+	 */
+	HostTimeout,
+};
+
+static constexpr uint32_t interrupt_bit(const OpenTitanI2cInterrupt Interrupt)
+{
+	return 1 << static_cast<uint32_t>(Interrupt);
+};
+
+/**
+ * Driver for the OpenTitan's I2C block.
+ *
+ * Documentation source can be found at:
+ * https://github.com/lowRISC/opentitan/tree/9ddf276c64e2974ed8e528e8b2feb00b977861de/hw/ip/i2c
+ */
+struct OpenTitanI2c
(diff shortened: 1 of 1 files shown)


sdk/include/platform/sunburst/platform-i2c.hh
```diff
@@ -0,0 +1,385 @@
...
+
+	/// Reset all of the fifos
+	void reset_fifos() volatile
+	{
+		fifoCtrl = (FifoControlReceiveReset | FifoControlFormatReset |
+		            FifoControlAcquiredReset | FifoControlTransmitReset);
+	}
+
+	/// Configure the I2C block to be in host mode
+	void set_host_mode() volatile
```
davidchisnall, line 244: We normally name things noun-verb, so that similar things are sorted together when autocompleting or reading generated docs.  Ideally avoid starting things with `set` because it ends up with a load of unrelated `set_` things in autocomplete.

My preferred style here (which we haven't done uniformly) is to have two overloads with the name of the thing, one of which takes an argument and updates it, the other that takes no arguments and does not update it.

sdk/include/platform/sunburst/platform-i2c.hh
```diff
@@ -0,0 +1,385 @@
...
+
+	/**
+	 * Set the I2C timing parameters appropriately for the given bit rate.
+	 * Distilled from:
+	 * https://github.com/lowRISC/opentitan/blob/9ddf276c64e2974ed8e528e8b2feb00b977861de/hw/ip/i2c/doc/programmers_guide.md
+	 */
+	void set_speed(const uint32_t SpeedKhz) volatile
+	{
+		// We must round up the system clock frequency to lengthen intervals.
+		constexpr uint32_t SysclkKhz = (I2cDevClockHz + 999) / 1000;
```
davidchisnall, line 257: Is this doing the same round up thing as your helper?

sdk/include/platform/sunburst/platform-i2c.hh
```diff
@@ -0,0 +1,385 @@
...
+		uint32_t mode = (SpeedKhz > 100u) + (SpeedKhz > 400u);
+
+		// Calculation of timing parameters
+		uint16_t tHigh =
+		  round_up_divide(SpcThigh[mode], clkPeriod);              // Spec. min.
+		uint16_t tLow = round_up_divide(SpcTlow[mode], clkPeriod); // Spec. min.
+		uint16_t tFall =
+		  round_up_divide(20 * 3 / 5, clkPeriod); // Spec. min. 3.3V
+		uint16_t tRise = round_up_divide(120, clkPeriod);
+		// Setup and Hold times for Start
```
davidchisnall, line 271: I can't quite parse this comment.

sdk/include/platform/sunburst/platform-i2c.hh
```diff
@@ -0,0 +1,385 @@
...
+		uint16_t thdSta = round_up_divide(SpcThdSta[mode], clkPeriod);
+		uint16_t tsuSta = round_up_divide(SpcTsuSta[mode], clkPeriod);
+		// Setup and Hold times for Data
+		uint16_t thdDat = round_up_divide(SpcThdDat[mode], clkPeriod);
+		uint16_t tsuDat = round_up_divide(SpcTsuDat[mode], clkPeriod);
+		uint16_t tBuf   = round_up_divide(SpcTBuf[mode], clkPeriod);
+		uint16_t tsuSto = round_up_divide(SpcTsuSto[mode], clkPeriod);
```
davidchisnall, line 278: There's a lot of code duplication here.  Would it be cleaner if you created an array of these and iterated over them?  Similarly, if `timing0` to `timing4` were exposed as an array, you'd be able to set them all in the same loop.

sdk/include/platform/sunburst/platform-i2c.hh
```diff
@@ -0,0 +1,385 @@
...
+	}
+
+	void blocking_write_byte(const uint32_t Fmt) volatile
+	{
+		while (0 != (StatusFormatFull & status)) {}
+		formatData = Fmt;
+	}
+
+	/// Returns true when the format fifo is empty
+	[[nodiscard]] bool format_empty() volatile
```
davidchisnall, line 300: Predicates should normally include `is` or `are` in their names.  This one is particularly bad.  We have an explicit style rule to avoid words that are both verbs and nouns and, in this case, format and empty are *both* verbs and nouns and I don't know if this is a thing that formats an empty thing, empties a format, or some combination of the two.

sdk/include/platform/sunburst/platform-i2c.hh
```diff
@@ -0,0 +1,385 @@
...
+		return 0 != (StatusFormatEmpty & status);
+	}
+
+	void blocking_write(const uint8_t  Addr7,
+	                    const uint8_t  data[],
+	                    const uint32_t NumBytes,
+	                    const bool     SkipStop) volatile
+	{
+		if (NumBytes == 0)
+			return;
```
davidchisnall, line 311: Missing braces.

sdk/include/platform/sunburst/platform-i2c.hh
```diff
@@ -0,0 +1,385 @@
...
+			{
+				buf[idx + chunkIdx] = readData;
+			}
+		}
+		return true;
+	}
+
+	/// Returns true if the given interrupt is asserted.
+	[[nodiscard]] bool
+	interrupt_asserted(const OpenTitanI2cInterrupt Interrupt) volatile
```
davidchisnall, line 353: Consider `interrupt_is_asserted`.

Avoid `const` for value parameters.

This could also do with some explanation of how these relate to interrupts in the interrupt controller.

sdk/include/platform/sunburst/platform-i2c.hh
```diff
@@ -0,0 +1,385 @@
...
+	{
+		return 0 != (interruptState & interrupt_bit(Interrupt));
+	}
+
+	/// Clears the given interrupt.
+	void clear_interrupt(const OpenTitanI2cInterrupt Interrupt) volatile
```
davidchisnall, line 359: Verb-noun again.

sdk/include/platform/sunburst/platform-i2c.hh
HU90m: I tend to use `const` by default even for value parameters, to ensure I don't modify variables I don't intend to. What are the disadvantages of having the extra `const` there?

(I have changed these to remove the `const` .)

sdk/include/platform/sunburst/platform-i2c.hh
HU90m: This is a brilliant idea

sdk/include/platform/sunburst/platform-i2c.hh
HU90m: `round_up_divide` wasn't originally used because it truncates the result to a 16 bit value. This isn't a problem for the sonata's clock frequency, so I made the suggested change.

However, CI is now failing with the following due to this and the lower default clock in `compile_flags.txt`.

```
implicit conversion from ''int'' to ''const uint16_t'' (aka ''const unsigned short'') changes value from 500000 to 41248
```

sdk/include/platform/sunburst/platform-i2c.hh
davidchisnall: Const in a parameter is an API contract where you promise that you won’t mutate it. Const for a value type is not a meaningful contract because it is not observable to the caller whether you mutate a non-aliased object. It increases cognitive load on the reader because you have written something that looks like an API contract but is not.  If you then do need to mutate it, it either complicates the code by making you have two live values that a reader needs to track (even though one is not changed) or requires you to change the function signature (which can have further knock-on effects on overload resolution, particularly in the presence of subclasses).

sdk/include/platform/sunburst/platform-i2c.hh
davidchisnall: I was a bit confused about why the `round_up_divide` function returned a 16-bit result, since there’s no guarantee that this is adequate to represent the result. I’d be more comfortable with it returning a `uint32_t`. Or, at least, having an assert that the result is representable as 16 bits.

davidchisnall: We can probably bump the clock speed in compile flags. The combination of flags doesn’t have to match any real target. It just has to not break things that it checks.

HU90m: I've added an assert to the `round_up_divide` function

davidchisnall (approved):

PR #240, opened 2024-06-07
Fix corner case in tickless scheduling.

davidchisnall:
There are two reasons we request a timer interrupt:

 - There is a peer at the same scheduler priority.
 - There is a timer that will expire.

If *both* of these were true, we were delaying the scheduling until the second condition.

### DIFF
--- a/sdk/core/scheduler/timer.h
+++ b/sdk/core/scheduler/timer.h
@@ -90,10 +90,15 @@ namespace
 			}
 			else
 			{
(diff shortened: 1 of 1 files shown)


rmn30 (approved):
I think this works. I was thinking you could make the conditional depend on `threadHasNoPeers` instead but I think this is better as it handles the case where the next timer expiry is sooner than one tick. I guess the 64-bit arithmetic is potentially expensive but it's probably OK.

PR #241, opened 2024-06-11
Add an API to control the blocking behavior of the allocator at a finer grain

hlef:
The current `timeout` argument of `heap_allocate*` functions determines whether or not and how long the allocator should block for an allocation to be fulfiled. However it does not control on what events the allocator should block.

Currently the allocator would block for three reasons: 1) there is no heap space available anymore, 2) the quota is exhausted, and 3) the quarantine is holding back memory which could be used for the allocation.

For `malloc` and `calloc`, we only want to wait on the quarantine. Thus, we need a finer way to control this blocking behavior.

This PR implements an API to do this, and uses it for `malloc` and `calloc`.

While we are at it, it does a little bit of cleanup in futex which has similar code with `flags`, but got it wrong.

### DIFF
--- a/sdk/include/stdlib.h
+++ b/sdk/include/stdlib.h
@@ -85,6 +85,17 @@ DEFINE_ALLOCATOR_CAPABILITY(__default_malloc_capability, MALLOC_QUOTA)
  */
 #define MALLOC_CAPABILITY STATIC_SEALED_VALUE(__default_malloc_capability)
 
+#ifndef MALLOC_WAIT_TICKS
+/**
+ * Define how long a call to `malloc` and `calloc` can block to fulfil an
+ * allocation. Regardless of this value, `malloc` and `calloc` will only ever
+ * block to wait for the quarantine to be processed. This means that, even with
+ * a non-zero value of `MALLOC_WAIT_TICKS`, `malloc` would immediately return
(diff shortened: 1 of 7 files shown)


sdk/include/stdlib.h
```diff
@@ -85,6 +85,15 @@ DEFINE_ALLOCATOR_CAPABILITY(__default_malloc_capability, MALLOC_QUOTA)
...
 #define MALLOC_CAPABILITY STATIC_SEALED_VALUE(__default_malloc_capability)
 
+/**
+ * Define how long a call to `malloc` and `calloc` can block to fulfil an
+ * allocation. Regardless of this value, `malloc` and `calloc` will only ever
+ * block to wait for the quarantine to be processed. This means that, even with
+ * a non-zero value of `MALLOC_WAIT_TICKS`, `malloc` would immediately return
+ * if the heap or the quota is exhausted.
+ */
+#define MALLOC_WAIT_TICKS 30
```
davidchisnall, line 95: Maybe wrap this in an `#ifndef` so that users can override it?  Since our malloc is always-inline, it's safe to make this a per-compilation-unit thing.

hlef:
I have updated the PR to address your comment and fix a few additional things.

davidchisnall (approved):

PR #244, opened 2024-06-12
Use @jrtc27's approach to enable large sealed objects.

### LINKED ISSUE #242: Constraints on sealed sizes can be reduced
In the RISC-V standardisation call, @jrtc27 pointed out that we can avoid complexity for the sealing arbitrarily large objects by placing the padding *before* the header such that the layout is:

 - Padding
 - Otype header
 - Unsealed object
 - Padding (within the unsealed object)

The bounds of the sealed object would then be the full allocation.  The address would be the start of the object header.  The allocate would ensure that the range from the header to the end is representable.  The unseal would do a csetbounds with the size computed from address to top.  This should keep the `token_unseal` simple and cost only a small amount of complexity in `token_seal`.

davidchisnall:
Allocating large sealed objects is difficult because sealed objects have a header that contains the virtual sealing type and both the full (sealed) object including the header and the (unsealed) part with the header excluded must be representable.

@jrtc27 pointed out that there's no real need for the header to be at the start of the allocation.  This neat trick lets us lay out the objects as:

```
|Padding|Header|Unsealed Object|
        ^
        Address of sealed capability is here.
```

As long as we first round up the size of the unsealed object to be representable, we can always then create a larger allocation that contains padding at the start to preserve representability of the sealed object.

This removes the restriction that we can't have sealed objects larger than 4072 bytes, which will eliminate some annoying redirection in a few places.

Fixes #242


sdk/core/allocator/main.cc
```diff
@@ -1093,13 +1090,20 @@ namespace
 			Debug::log("Underlying allocation failed for sealed object");
 			return {nullptr, nullptr};
 		}
+		obj.address() = obj.top() - (sz + ObjHdrSize);
+		// Round down the base to the heap alignment size.
+		// This ensures that the header is aligned and gives the same alignment
+		// as a normal allocation.  We will already be this aligned for most
+		// requested sizes.
+		obj.align_down(MallocAlignment);
```
nwf-msr, line 1098: This probably deserves more commentary, something like...

we're guaranteed that `obj` is still in-bounds because it is the result of `malloc_internal` for `sz + ObjHdrSize` bytes and so it is at least that big and its base must already be `MallocAlignment`-aligned, so `obj`'s address is at least its base.  Moreover, because `ObjHdrSize` is a multiple of `MallocAlignment`, the payload at `pointer_offset(obj, ObjHdrSize)` is still `MallocAlignment`-aligned.

nwf-msr (commented):
LGTM but I would like more commentary. :)

I think I'd like if the ASCII-art from the commit message ended up in a comment in the allocator, perhaps on `SObjStruct` in `token.h`.

sdk/core/allocator/main.cc
rmn30: Thanks, this was also bothering me!

rmn30: We also need an argument that the payload is still representable as it is no longer necessarily just the last `crrl(sz)` bytes of the allocation. It probably is given the limited situations in which the `align_down` will do anything (I think `crrl(sz)` would have to not be a multiple of `MallocAlignment`).

rmn30: Also what happens if `sz+ObjHdrSize` overflows?

rmn30: I think, if the original `sz` is very large `crrl` will return zero so we will allocate `ObjHdrSize` (8) bytes. The sealed capability would then by 0 bytes at the top of the allocation.

davidchisnall: Oh, size overflow is possible, good catch. I think we’ll then allocate something too small. You may be able to get something out of bounds. Worth checking for overflow explicitly.

nwf-msr: > We also need an argument that the payload is still representable as it is no longer necessarily just the last `crrl(sz)` bytes of the allocation. It probably is given the limited situations in which the `align_down` will do anything (I think `crrl(sz)` would have to not be a multiple of `MallocAlignment`).

So... more generally: we're allocating and returning a representable span of at least `crrl(8 + crrl(sz))` bytes (the malloc internal header's size is used for accounting internally, and is placed such that its adjacent body is representable, but it does not otherwise impact the representability of the returned capability), and so the question is "what are the representable at-least-`crrl(sz)`-sized and 8-aligned spans inside that allocation" and, in particular, "is the last `crrl(sz)` bytes such a span" and I think the answer to that is "not necessarily".

nwf-msr: Oh, wait, no, everything's fine.
1. the allocator guarantees that both endpoints of the `malloc`-d chunk are both `-cram(8 + crrl(sz))`-aligned and 8-aligned (really, whichever is more restrictive).
2. `-cram()`-alignment is monotone, so the last endpoint in particular is also `-cram(crrl(sz))`-aligned (and also still 8-aligned).
3. So there's some point strictly after the beginning of the allocation that's also all of
   1. at least `crrl(sz)` away from the end
   2. `-cram(crrl(sz))`-aligned and
   3. 8-aligned
4. We can put our header to the left of this point, since 8-alignment of separate things on either side of it guarantees space.

The only case where this is unusual is exactly as Robert said: if `crrl(sz)` isn't 0 mod 8, then we're effectively additionally padding the payload up to 8 bytes and moving it away from the actual end of the allocation.  But that's exactly what we would have done if we'd allocated a `sz`-byte object anyway.

davidchisnall: > Oh, wait, no, everything's fine.

And this is why I want everyone and their ~dog~ SMT solver to read this PR before I merge it!

rmn30: I've not written the SMT query yet but I wrote a quick test of some interesting values which it would be good to add to the test-suite:

```c++
auto key = token_key_new();
size_t sizes[] = {0, 0xffffffff, 511, 511 << 1, 511 << 2, 511 << 3};
void *unsealed;
for (size_t sz : sizes) {
  SObj sealed = blocking_forever<token_sealed_unsealed_alloc>(MALLOC_CAPABILITY, key, sz, &unsealed);
  Debug::log("size {}\n  sealed: {}\nunsealed: {}", sz, sealed, unsealed);
  token_obj_destroy(MALLOC_CAPABILITY, key, sealed);
}
```

Output looks OK (assuming we are OK with the zero length caps for first two) and shows padding both before and after the unsealed object:

```
Caller compartment: size 0x0
  sealed: 0x80007d10 (v:1 0x80007d10-0x80007d18 l:0x8 o:0xc p: G RWcgm- -- ---)
unsealed: 0x80007d18 (v:0 0x80007d18-0x80007d18 l:0x0 o:0x0 p: G RWcgm- -- ---)
Caller compartment: size 0xffffffff
  sealed: 0x80007d20 (v:1 0x80007d20-0x80007d28 l:0x8 o:0xc p: G RWcgm- -- ---)
unsealed: 0x80007d28 (v:0 0x80007d28-0x80007d28 l:0x0 o:0x0 p: G RWcgm- -- ---)
Caller compartment: size 0x1ff
  sealed: 0x80007d30 (v:1 0x80007d30-0x80007f38 l:0x208 o:0xc p: G RWcgm- -- ---)
unsealed: 0x80007d38 (v:1 0x80007d38-0x80007f37 l:0x1ff o:0x0 p: G RWcgm- -- ---)
Caller compartment: size 0x3fe
  sealed: 0x80007f40 (v:1 0x80007f40-0x80008348 l:0x408 o:0xc p: G RWcgm- -- ---)
unsealed: 0x80007f48 (v:1 0x80007f48-0x80008346 l:0x3fe o:0x0 p: G RWcgm- -- ---)
Caller compartment: size 0x7fc
  sealed: 0x80008350 (v:1 0x80008350-0x80008b58 l:0x808 o:0xc p: G RWcgm- -- ---)
unsealed: 0x80008358 (v:1 0x80008358-0x80008b54 l:0x7fc o:0x0 p: G RWcgm- -- ---)
Caller compartment: size 0xff8
  sealed: 0x80008b60 (v:1 0x80008b60-0x80009b60 l:0x1000 o:0xc p: G RWcgm- -- ---)
unsealed: 0x80008b68 (v:1 0x80008b68-0x80009b60 l:0xff8 o:0x0 p: G RWcgm- -- ---)
```

rmn30: Incidentally when I run with a too small quota and allocator debugging enabled I see what looks like a crash during a debug print:

```
Allocator: Quota exhausted trying to allocate 0xCaller compartment: size 0xff8
  sealed: 0xffffffff (v:0 0xfffffe00-0xfffffe00 l:0x0 o:0x0 p: - ------ -- ---)
unsealed: 0x8000b358 (v:0 0x8000b358-0x8000bb54 l:0x7fc o:0x0 p: G RWcgm- -- ---)
```

Note sure if that is expected?

nwf-msr (approved):
LGTM.  Thank you for the additional commentary. <3

rmn30:
For completeness, I proved the following property using Sail SMT backend. It checks that sealed object will be representable and large enough given some assumptions about the allocator behaviour, namely rounding up to a multiple of 8 and a representable size. This only checks the size and not the alignment, but we can rely on the fact that the top boundary is aligned by the allocator coupled with the monotonicity of `cram` mentioned by @nwf-msr.

```js
function prop_alloc_sealed_rep(reqSz : CapAddrBits) -> bool = {
    /* alloc_sealed_unsealed first rounds up requested size to a representable size */
    let unsealedSz = getRepresentableLength(reqSz);
    /* then adds 8 bytes for header */
    let sealedReqSz = unsealedSz + 8;
    /* The allocator rounds this up to multiple of 8 */
    let sealedReq8  = [sealedReqSz + 7 with 2..0=0b000]; 
    /* and to a representable size. */
    let allocSz = getRepresentableLength(sealedReq8); 
    /* alloc_sealed_unsealed computes headerPtr relative to top of the allocation rounded down to an 8 byte boundary */
    let headerPtr = [allocSz - sealedReqSz with 2..0=0b000];
    /* the sealed object is 8 bytes above headerPtr */
    let sealedPtr = headerPtr + 8; 
    /* therefore the size of the sealed object can be computed from the size of the allocation minus the sealedPtr */
    let sealedObjSz = allocSz - sealedPtr; 
    /* We want the sealedObjSz to be exactly representable */
    let representable = sealedObjSz == getRepresentableLength(sealedObjSz);
    /* ...and large enough for the requeseted size */
    let bigEnough = sealedObjSz >=_u reqSz;
    /* For the final property exclude sizes that are zero or result in crrl overflow */
    (unsealedSz != zeros()) --> (representable & bigEnough)
}
```

PR #246, opened 2024-06-13
Added a driver for the Sonata RGB LED Controller

HU90m:


sdk/include/platform/sunburst/platform-rgbctrl.hh
```diff
@@ -0,0 +1,83 @@
...
+ * A driver for the Sonata's RGB LED Controller
+ */
+struct SonataRgbLedController
+{
+	/**
+	 * Registers for setting the 8-bit red, green, and blue values
+	 * for the two RGB Leds.
+	 */
+	uint32_t rgbLed0;
+	uint32_t rgbLed1;
```
davidchisnall, line 24: Should this be `uint32_t ledColours[2]` or similar?  You can then use the value of `SonataRgbLed` (0 or 1) to index into them.

sdk/include/platform/sunburst/platform-rgbctrl.hh
```diff
@@ -0,0 +1,83 @@
...
+	/// Control Register
```
davidchisnall, line 25: It would be nice to document the bits here.  You've done that properly with the enums below, so just a forward ref saying that these are set with the values from the control enum below would be nice.

sdk/include/platform/sunburst/platform-rgbctrl.hh
```diff
@@ -0,0 +1,83 @@
...
+	/// Status Register Fields
+	enum [[clang::flag_enum]] : uint32_t{
+	  /**
+	   * When asserted controller is idle and new colours can be set,
+	   * otherwise writes to regLed0, regLed1, and control are ignored.
+	   */
+	  StatusIdle = 1 << 0,
+	};
+
+	void wait_for_idle() volatile
```
davidchisnall, line 50: What does idle mean for an LED?  I vaguely remember you or Greg saying that these are surprisingly stateful devices and you're sending them update commands, but a comment to that effect would be nice.

sdk/include/platform/sunburst/platform-rgbctrl.hh
```diff
@@ -0,0 +1,83 @@
...
+		               static_cast<uint32_t>(red);
+		wait_for_idle();
+
+		switch (led)
+		{
+			case SonataRgbLed::Led0:
+				rgbLed0 = rgb;
+			case SonataRgbLed::Led1:
+				rgbLed1 = rgb;
+		};
```
davidchisnall, line 69:
```suggestion
		wait_for_idle();
		ledColours[led] = static_cast<uint32_t>(blue) << 16) |
		 		               (static_cast<uint32_t>(green) << 8) |
		 		               static_cast<uint32_t>(red);
```

sdk/include/platform/sunburst/platform-rgbctrl.hh
```diff
@@ -0,0 +1,83 @@
...
+	  StatusIdle = 1 << 0,
+	};
+
+	void wait_for_idle() volatile
+	{
+		while ((status & StatusIdle) == 0) {}
+	}
+
+	void
+	rgb(uint8_t red, uint8_t green, uint8_t blue, SonataRgbLed led) volatile
```
davidchisnall, line 56: From the existence of the `update` method, I presume that this does not actually change the colour?  A comment to that effect would be nice.

davidchisnall, line 56: This is quite subjective, but I generally prefer APIs like this to have the object that they're operating on as the first parameter (we have a few exceptions to this in the style guide, but none of them apply here), which would suggest having the `led` parameter first.

sdk/include/platform/sunburst/platform-rgbctrl.hh
```diff
@@ -0,0 +1,83 @@
...
+		switch (led)
+		{
+			case SonataRgbLed::Led0:
+				rgbLed0 = rgb;
+			case SonataRgbLed::Led1:
+				rgbLed1 = rgb;
+		};
+	}
+
+	void update() volatile
```
davidchisnall, line 72: Does this need to be called after `rgb`?  What happens if you call it first (LEDs come on in some undefined colour?)

sdk/include/platform/sunburst/platform-rgbctrl.hh
```diff
@@ -0,0 +1,83 @@
...
+	{
+		wait_for_idle();
+		control = ControlSet;
+	}
+
+	void clear() volatile
```
davidchisnall, line 78: Does this turn the LEDs off?  Both of them?

sdk/include/platform/sunburst/platform-rgbctrl.hh
HU90m: The reset values of the `ledColour` registers is 0x0, if they haven't been set they will default to black (off).

sdk/include/platform/sunburst/platform-rgbctrl.hh
```diff
@@ -0,0 +1,89 @@
...
+	enum [[clang::flag_enum]] : uint32_t{
+	  /// Write 1 to set RGB LEDs to specified colours.
+	  ControlSet = 1 << 0,
+	  /**
+	   * Write 1 to turn off RGB LEDs.
+	   * Write to ControlSet to turn on again.
+	   */
+	  ControlOff = 1 << 1,
+	};
+	/// Control Register (fields documented above)
```
davidchisnall, line 35: 'Above' isn't great because the 'above' bit isn't visible in the mouse-over pop-up in IDEs (which is the main thing that we write doc comments for).  If you point at another enum by name, people can refer to that.

sdk/include/platform/sunburst/platform-rgbctrl.hh
davidchisnall: Does it ever become busy as a result of something external (and so is this racy)?  I'm assuming not and this is similar to the colour LED strips where you just write a load of 24-bit values in a waveform down the wire to update each one and so the controller here is just a little queue that stores the two words and writes them down the wire?

sdk/include/platform/sunburst/platform-rgbctrl.hh
davidchisnall: Is there a way of turning one of the LEDs off after they've both been turned on?

sdk/include/platform/sunburst/platform-rgbctrl.hh
HU90m: Yes, I don't think it can be a result of something external.

(Idle is assigned [here](https://github.com/lowRISC/sonata-system/blob/191f5a796e4fcc032c6325d1eb1122150e23d0db/rtl/ip/rgbled_ctrl/rtl/ws281x_drv.sv#L134).)

davidchisnall (approved):

PR #248, opened 2024-06-13
Add APIs to retrieve the thread ID of the owner of a lock.

hlef:
It is commonly needed in error handlers to have to check if the thread on which the error handler was invoked was holding a particular lock when it crashed. Typically, we would want to do this to be able to forcefully release the lock. This is currently difficult to do, as the thread ID which is stored in priority inheriting flag locks is not exposed through the C/C++ APIs.

This commit adds news APIs `get_owner_thread_id` (C++) and `flaglock_priority_inheriting_get_owner_thread_id` (C) to retrieve the thread ID of the owner of a lock.

Following this commit, error handlers can call `get_owner_thread_id` (or `flaglock_priority_inheriting_get_owner_thread_id`) on a lock and compare the result with `thread_id_get` to know if the thread which crashed owned the lock.

### DIFF
--- a/sdk/include/locks.h
+++ b/sdk/include/locks.h
@@ -144,6 +144,33 @@ void __cheri_libcall
 flaglock_upgrade_for_destruction(struct FlagLockState *lock);
 
 /**
+ * Return the thread ID of the owner of the lock.
+ *
+ * This is only available for priority inherited locks, as this is the only
+ * case where we store the thread ID of the owner.
+ *
+ * The return value is 0 if the lock is not owned or if called on a
+ * non-priority inherited flag lock. The return value is undefined if called on
+ * an uninitialized lock.
+ *
+ * This *will* race with succesful `lock` and `unlock` operations on other
+ * threads, and should thus not be used to check if the lock is owned.
+ *
+ * The main use case for this function is in the error handler to check whether
+ * or not the lock is owned by the thread on which the error handler was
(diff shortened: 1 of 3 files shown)


davidchisnall (approved):

davidchisnall:
This currently does an atomic load (libcall), it might be a good idea to cast the flag word to a `uint32_t*` first so that we do an unsequenced load.

sdk/include/locks.h
```diff
@@ -143,6 +143,30 @@ void __cheri_libcall flaglock_unlock(struct FlagLockState *lock);
...
+ *
+ * The main use case for this function is in the error handler to check whether
+ * or not the lock is owned by the thread on which the error handler was
+ * invoked.  In this case we can call this function and compare the result with
+ * `thread_id_get` to know if the current thread owns the lock.
+ */
+__always_inline static inline uint16_t
+flaglock_priority_inheriting_get_owner_thread_id(struct FlagLockState *lock)
+{
+	return (lock->lockWord & 0x0000ffff);
```
davidchisnall, line 167:
```suggestion
	// The lock must be held at this point for the value to be stable so do a non-atomic read.
	return ((*(uint32_t*)&(lock->lockWord)) & 0x0000ffff);
```

nwf-msr (approved):
I'm happy with this, but if there's no rush it might be better to wait for David to get back for another glance before merging.

davidchisnall (approved):

PR #249, opened 2024-06-14
Fix invalid dereference in the `Pointer` proxy.

hlef:
This code does not compile, we are running into this issue because we just started using it. Add a test to reproduce the issue and check for basic behavior.


nwf-msr (approved):
Whoops!  We stopped using `Pointer` proxies in the allocator and I guess I broke it after it stopped getting tested.

Thanks for the fix and for the expanded test.

PR #250, opened 2024-06-14
Update Sonata getting started docs.

davidchisnall:


docs/GettingStarted.md
```diff
@@ -352,3 +376,99 @@ issions are:
...
+On macOS, the device name includes the serial number of the device and so will be something like `/dev/tty.usbserial-LN28292` (the last digit is the serial port number).
+On Windows, it will be a COM port, but unfortunately Windows is not always consistent about the order in which the ports appear.
+
+If you install a serial terminal (for example, `minicom` on *NIX or PuTTY on Windows), you can point it at the relevant serial port.
+
+The serial setup is the same as for the Arty A7 [above](#installing-and-running-the-firmware), aside from the different device name.
+
+Once a serial terminal is connected, running `xmake run` should show the output from the test suite.
+
+If you are unsure which of Sonata's serial ports is the correct one, you can connect your serial console to all four and see which produces output from the test suite.
```
rmn30, line 474: I don't think it's a good idea to connect to all of them at the same time. I think one might be the debugger and actually halts CPU or something.
```suggestion
If you are unsure which of Sonata's serial ports is the correct one, you can connect your serial console to each of them in turn and see which produces output from the test suite.
```

rmn30 (approved):

docs/GettingStarted.md
```diff
@@ -223,6 +224,29 @@ For Ubuntu, you can do:
 # apt install xmake
 ```
 
+### Installing u2futils
+
+If you are working with Sonata, you will need to convert the ELF files that the linker produces to [USB Flashing Format (UF2)](https://github.com/microsoft/uf2) format.
```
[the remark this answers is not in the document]
davidchisnall: USB Flashing Format format?

docs/GettingStarted.md
davidchisnall: > I don't think it's a good idea to connect to all of them at the same time. I think one might be the debugger and actually halts CPU or something.

I connected to all four on my Mac to figure out the right one and it was fine.

PR #255, opened 2024-06-17
Add std:string overload for Debug

PhilDay-CT:
Add an overload in debug for std:string based on the pattern for std::string_view

### DIFF
--- a/sdk/include/__debug.h
+++ b/sdk/include/__debug.h
@@ -29,6 +29,8 @@ enum DebugFormatArgumentKind : ptraddr_t
 	DebugFormatArgumentCString,
 	/// String view, printed as-is.
 	DebugFormatArgumentStringView,
+	/// String, printed as-is.
+	DebugFormatArgumentString,
 };
 
 struct DebugFormatArgument
--- a/sdk/include/__debug.h
+++ b/sdk/include/__debug.h
@@ -29,8 +29,6 @@ enum DebugFormatArgumentKind : ptraddr_t
 	DebugFormatArgumentCString,
 	/// String view, printed as-is.
 	DebugFormatArgumentStringView,
-	/// String, printed as-is.
-	DebugFormatArgumentString,
 };
 
 struct DebugFormatArgument
--- a/sdk/include/debug.hh
+++ b/sdk/include/debug.hh
@@ -65,10 +65,6 @@ struct DebugWriter
 	 */
 	virtual void write(std::string_view) = 0;
 	/**
-	 * Write a string.
-	 */
-	virtual void write(std::string) = 0;
-	/**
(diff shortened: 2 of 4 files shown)


davidchisnall:
Hmm, I’d hoped that you’d be able to just forward the string as a string view here, but that requires constructing an on-stack string view. I don’t really like adding another custom type directly.

It’s probably better to have a class that handles std::string as we do for enums and just calls the C string overload on the writer.

rmn30:
Could we just resort to the underlying string like this?

```
template<>
struct DebugFormatArgumentAdaptor<std::string>
{
	__always_inline static DebugFormatArgument
	construct(const std::string &value)
	{
		return DebugFormatArgumentAdaptor<const char *>::construct(value.c_str());
	}
};
```

davidchisnall (approved):
Thanks, I forgot these were internally null terminated. Can you add a comment here reminding me to add a test for this when the compiler bug is fixed?

https://github.com/microsoft/cheriot-rtos/blob/58e3aee92a22604b9c75715fad45822e80229f46/tests/test-runner.cc#L96

davidchisnall:
(Or add a test if there’s an easy way to do it that doesn’t trigger the compiler bug)

PhilDay-CT:
>> (Or add a test if there’s an easy way to do it that doesn’t trigger the compiler bug)

I couldn't see a way to create a std::string to test with without hitting the compiler bug - could just hold of merging this PR until that's fixed, then we could add the test in there.   (There's no real rush from my side on this - I have a bunch of other stuff to get on with)

rmn30:
It is possible to bypass the problematic constructor using C++14 string literals:
```C++
using namespace std::string_literals;
std::string s = "hello"s;
```
I think this is even very slightly more efficient as it doesn't result in a runtime call to `strlen`!

davidchisnall:
#260 should make `__builtin_strlen` work, after the dev container has finished rebuilding.  Once that's merged, this can be rebased and merged (you can replace the test that uses `.c_str()` for printing, that was just to make sure that the `std::string` was correctly constructed).

davidchisnall:
This is the dev container rebuild: https://cirrus-ci.com/task/5612295751991296

davidchisnall:
The dev container fix is in and #260 is merged, so if you can rebase this (happy on the merge conflict to just throw away my test and add yours) then we can merge it.

davidchisnall:
Did you rebase on main?

davidchisnall:
I fixed the merge conflict, it just needs the indenting fixed.

PhilDay-CT:
Sorry,  realised I'd pulled main from origin rather that upstream and sorted that out before I saw your last comment

PR #258, opened 2024-06-20
Test and document the linked list from `ds::linked_list`.

hlef:
Although our linked list implementation does feature comments, they are more about the underlying technical details and do not really help newcomers understand how to use it. This matters because we do not have `std::list`.

Address this by adding a test for the linked list which also explains how to use it for a basic use-case.

### DIFF
--- a/tests/list-test.cc
+++ b/tests/list-test.cc
@@ -1,0 +1,241 @@
+// Copyright Microsoft and CHERIoT Contributors.
+// SPDX-License-Identifier: MIT
+
+#define TEST_NAME "List"
+#include "tests.hh"
+#include <ds/linked_list.h>
+
+using CHERI::Capability;
+
+namespace
+{
+	/**
+	 * Example class we want to link into a doubly linked list.
+	 *
+	 * The class contains a single integer for the purposes of the test.
+	 *
+	 * `ds::linked_list` is an intrusive list: we embed the list node into
+	 * the class we want to link.  There are various implementations of the
+	 * list nodes. Here we use the most simple one
+	 * (`ds::linked_list::cell::Pointer`) which relies on two pointers
+	 * `next` and `prev`.
+	 */
+	struct LinkedObject
+	{
+		using ObjectRing = ds::linked_list::cell::Pointer;
+
+		int data;
+		/**
+		 * List node: links objects into the doubly-linked list.
+		 */
+		ObjectRing ring __attribute__((__cheri_no_subobject_bounds__)) = {};
+		/**
+		 * Container-of for the above field. This is used to retrieve the
+		 * corresponding object from a list element.
+		 */
+		__always_inline static struct LinkedObject *from_ring(ObjectRing *c)
+		{
+			return reinterpret_cast<struct LinkedObject *>(
+			  reinterpret_cast<uintptr_t>(c) -
+			  offsetof(struct LinkedObject, ring));
+		}
+	};
+} // namespace
+
+/**
+ * `ds::linked_list`s are circular, doubly-linked collections. While they can
+ * stand on their own as rings of objects, it is sometimes convenient to create
+ * a designated 'sentinel' node that participates in the collection without
+ * being part of the collection:
+ *
+ * - a sentinel node provides pointers to the effective head and tail of the
+ *   collection (the successor and predecessor of the sentinel, respectively)
+ *
+ * - a sentinel allows not having to special-case 'the collection is empty' in
+ *   as many places as some other representations (that is, collections with
+ *   sentinels need fewer NULL pointer checks)
+ *
+ * - a sentinel provides many handy functions to operate on the list
+ *
+ * Note: do not allocate the sentinel (or any list cell) on the stack, because
+ * it would lead some list nodes to hold a pointer to a stack value, i.e., to
+ * an invalid capability. This would manifest as a crash while using the list.
+ */
+ds::linked_list::Sentinel<LinkedObject::ObjectRing> objects = {};
+
+void test_list()
+{
+	debug_log("Testing the list implementation.");
+
+	// Number of elements we will add to the list in the test. Must be
+	// divisible by two.
+	static constexpr int NumberOfListElements = 30;
+
+	auto heapAtStart = heap_quota_remaining(MALLOC_CAPABILITY);
+
+	TEST(objects.is_empty(), "Newly created list is not empty");
+
+	// Create heap-allocated objects, and link them into the linked list.
+	for (int i = 0; i < NumberOfListElements; i++)
+	{
+		Timeout       t{UnlimitedTimeout};
+		LinkedObject *o = static_cast<LinkedObject *>(
+		  heap_allocate(&t, MALLOC_CAPABILITY, sizeof(LinkedObject)));
+		TEST(Capability{o}.is_valid(), "Cannot allocate linked object");
+
+		// Use the object integer as an index.
+		o->data = i;
+		// The list node has not yet been initialized.
+		o->ring.cell_reset();
+
+		// Test that we can retrieve the object from the link node and
+		// that this results in a capability which is identical to what
+		// we got from the allocator.
+		TEST(Capability{o} == Capability{LinkedObject::from_ring(&(o->ring))},
+		     "The container of method does not return the right object");
+		TEST(Capability{LinkedObject::from_ring(&(o->ring))}.is_valid(),
+		     "Capability retrieved from `from_ring` is invalid");
+
+		// Add the new object to the list through the sentinel node.
+		objects.append(&(o->ring));
+	}
+
+	TEST(!objects.is_empty(), "The list is empty after adding objects");
+
+	// Test that the sentinel can be used to retrieve the first and last
+	// elements of the list as expected.
+	TEST(LinkedObject::from_ring(objects.last())->data ==
+	       NumberOfListElements - 1,
+	     "Last element of the list is incorrect, expected {}, got {}",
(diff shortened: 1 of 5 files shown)


tests/list-test.cc
```diff
@@ -0,0 +1,211 @@
...
+	/**
+	 * Example class we want to link into a doubly linked list.
+	 *
+	 * The class contains a single integer for the purposes of the test.
+	 *
+	 * `ds::linked_list` is an intrusive list: we embed the list node into
+	 * the class we want to link.  There are various implementations of the
+	 * list nodes. Here we use the most simple one
+	 * (`ds::linked_list::cell::Pointer`) which relies on two pointers
+	 * `next` and `prev`.
```
nwf-msr, line 21: This bit of the comment might be more naturally associated with the `using ObjectRing` above?  That `using` could, I think, move into the `LinkedObject` class were that to be a more natural place for it (the `Sentinel<ObjectRing>` below would have to become `Sentinel<LinkedObject::ObjectRing>`, but otherwise nothing changes, I think).

tests/list-test.cc
```diff
@@ -0,0 +1,211 @@
...
+			  reinterpret_cast<uintptr_t>(c) -
+			  offsetof(struct LinkedObject, ring));
+		}
+	};
+} // namespace
+
+/**
+ * The sentinel is a special class encapsulating a list node. It provides many
+ * handy functions to operate on the list. The sentinel is thus part of the
+ * list, without being one of the objects linked.
```
nwf-msr, line 46: This is all true, but it might skip over some steps.  Perhaps something like...

`ds::linked_list`-s are circular, doubly-linked collections.  While they can stand on their own as rings of objects, it is sometimes convenient to create a designated "sentinel" node that participates in the collection without being part of the collection: it naturally provides pointers to the effective head and tail of the collection (the successor and predecessor of the sentinel, respectively) while not having to special-case "the collection is empty" in as many places as some other representations (that is, collections with sentinels need fewer "null pointer" checks).  Towards that end, the `Sentinel` class captures the notion of a designated not-an-element node and provides many handy functions to operate on the list.

tests/list-test.cc
```diff
@@ -0,0 +1,211 @@
...
+ *
+ * Note: do not attempt to define the sentinel on the stack, because it would
+ * lead some list nodes holding a pointer to a stack value, i.e., to an invalid
+ * capability. This would lead to a crash while using the list.
```
nwf-msr, line 50: This is a really good point and is more general than just the sentinel: don't mix heap and stack allocated objects in a `ds::linked_list` in general!  All of our `cell` implementations will crash in one way or another if you try.

tests/list-test.cc
```diff
@@ -0,0 +1,211 @@
...
+		TEST(CHERI::Capability{o}.is_valid(), "Cannot allocate linked object");
+
+		// Use the object integer as an index.
+		o->data = i;
+		// The list node has not yet been initialized.
+		o->ring.cell_reset();
+
+		// Test that we can retrieve the object from the link node and
+		// that this results in a valid capability.
+		TEST(o == LinkedObject::from_ring(&(o->ring)),
```
nwf-msr, line 81: `==` on pointers in CHERI C/C++ presently defaults to address-only equality.  You might use `__builtin_cheri_equal_exact`, or wrap with `Capability` first (since its `operator==` uses that primitive), to be more stringent.

tests/list-test.cc
```diff
@@ -0,0 +1,211 @@
...
+		     "The container of method does not return the right object");
+		TEST(CHERI::Capability{LinkedObject::from_ring(&(o->ring))}.is_valid(),
+		     "Capability retrieved from `from_ring` is invalid");
+
+		// Add the new object to the list through the sentinel node.
+		objects.append(&(o->ring));
+	}
+
+	TEST(!objects.is_empty(), "The list is not empty.");
```
rmn30, line 90:
```suggestion
	TEST(!objects.is_empty(), "The list is empty after adding objects.");
```

tests/list-test.cc
```diff
@@ -0,0 +1,211 @@
...
+
+	// Test that we can go through the list by following `cell_next`
+	// pointers as expected.
+	int counter = 0;
+	// While at it, retrieve a pointer to the middle element which we will
+	// use to cleave the list later.
+	ObjectRing *middle = nullptr;
+	// We reach the sentinel when we have gone through all elements of the
+	// list.
+	for (auto *cell = objects.first(); cell != &objects.sentinel;
```
nwf-msr, line 120: More idiomatically, there's `.search()` on the `Sentinel`, though it's probably not bad to demonstrate the raw form as well.

tests/list-test.cc
```diff
@@ -0,0 +1,211 @@
...
+
+		// Add the new object to the list through the sentinel node.
+		objects.append(&(o->ring));
+	}
+
+	TEST(!objects.is_empty(), "The list is not empty.");
+
+	// Test that the sentinel can be used to retrieve the first and last
+	// elements of the list as expected.
+	TEST(LinkedObject::from_ring(objects.last())->data ==
```
rmn30, line 94: Aside: would be nice to have a `TEST_EQUAL` to make this pattern less repetitive.

tests/list-test.cc
```diff
@@ -0,0 +1,211 @@
...
+		{
+			middle = cell;
+		}
+		counter++;
+	}
+
+	TEST(middle != nullptr, "Could not find middle element of the list");
+
+	// Cut the list in the middle.
+	ds::linked_list::remove(middle, objects.last());
```
nwf-msr, line 139: Speaking of `Sentinel.search`, `cell::remove` makes the claim that it can be used within if one is careful.  It'd be good to both validate that that isn't a stale claim and, if all's good, demonstrate that somewhere in here, too.

tests/list-test.cc
```diff
@@ -0,0 +1,211 @@
...
+		if (counter == NumberOfListElements / 2)
+		{
+			middle = cell;
+		}
+		counter++;
+	}
+
+	TEST(middle != nullptr, "Could not find middle element of the list");
+
+	// Cut the list in the middle.
```
nwf-msr, line 138: It might be worth pointing out that `middle` is, after this, still a valid handle to a valid collection of objects that have become detached from the sentinel.

tests/list-test.cc
nwf-msr, line 139: It might be worth testing that the return from `remove` is a node whose successor is the sentinel.

tests/list-test.cc
```diff
@@ -0,0 +1,211 @@
...
+	}
+	TEST(counter == NumberOfListElements / 2,
+	     "Cleaving didn't leave a list with the right number of elements");
+
+	// Now remove (and free) a single element from the list.
+	TEST(LinkedObject::from_ring(objects.first())->data == 0,
+	     "First element of the list is incorrect, expected {}, got {}",
+	     0,
+	     LinkedObject::from_ring(objects.first())->data);
+	// We must keep a reference to the removed object to free it.
```
nwf-msr, line 156: It may be worth expanding on this comment.  It's typical that `T *remove(T*)` methods return the thing being removed, but `ds::linked_list::cell::remove` breaks that expectation (for reasons that I think are good reasons, but YMMV), which might be surprising to users.

nwf-msr (commented):
Thank you for doing what I should have done in the first place.  This will be a very useful bit of machine-checked documentation to point new people towards. ❤️

Some suggestions here and there, but it's off to a great start.

tests/list-test.cc
```diff
@@ -0,0 +1,211 @@
...
+
+	// We must also free the span of the list which we removed earlier.
+	cell = middle;
+	do
+	{
+		struct LinkedObject *o = LinkedObject::from_ring(cell);
+		cell                   = cell->cell_next();
+		heap_free(MALLOC_CAPABILITY, o);
+		counter++;
+	} while (cell != middle);
```
nwf-msr, line 195: This might be a good place to demonstrate some of why the interface is as quirky as it is.  You could showcase in-place mutation and traversal without a sentinel with something like (not compile tested, but hopefully close):

```
ds::linked_list::search(middle, [](ObjectRing &cell) {
  auto l = ds::linked_list::unsafe_remove(cell);
  free(LinkedObject::from_ring(cell);
  cell = l;
 });
free(LinkedObject::from_ring(middle));

tests/list-test.cc
hlef: Agree, makes sense!

tests/list-test.cc
hlef: That's great! I'm adding it :slightly_smiling_face:

tests/list-test.cc
hlef: Oh, great point! I will apply this to other comparisons in the test.

tests/list-test.cc
hlef: Ah, good point! This actually bit me earlier because I wrongly assumed it would return the removed object.

tests/list-test.cc
hlef: Agree! Though I would rather avoid introducing it in this PR if possible (if we do introduce such a helper, it would be nice to go through other tests to use it)

tests/list-test.cc
hlef: Awesome, I added this with some small fixes.

hlef:
@nwf-msr @rmn30  All comments addressed (hopefully) :slightly_smiling_face:

nwf-msr (approved):
This looks great.  Thank you again.

PR #260, opened 2024-06-25
Fix `__builtin_strlen`

davidchisnall:
This is the other half of CHERIoT-Platform/llvm-project#29

The compiler will now use the correct calling convention for `__builtin_strlen`, but does not know how to generate it as a mangled name.  We do this by hand to avoid adding more complexity to the compiler for a special case.


nwf-msr (approved):
LGTM

davidchisnall:
Unfortunately, rebuilding the dev container is currently held up by the fact that SAFE is broken.

PR #269, opened 2024-07-02
Weird Tag violation issue - not to be merged

PhilDay-CT:
Hi Folks,

In reworking the config broker change I've created a weird issue that I could do with some other eyes on please - it seemed easiest to create a cut down reproducer and supply it as a dummy PR.

The code is in examples/10a.odd_error.  As it stand it compiles and runs fine, but if I comment out line 37 in subscriber1.cc and recompile then at run time it generates a tag violation.   The weird bit (to me at least) is that line is in a function that I'm not using anymore - it's a hangover from the previous callback based approach.   However I've clearly created done something to I guess upset the linker ?? 

If I also comment out line 55 (which is the other call to print_config()   ) then that also seems to keep things happy, so I guess its something to do with the global defined in line 25  - but the tag violation is caught in the config_broker compartment, and this value is only even used in those two local calls ?

### DIFF
--- a/examples/10a.odd_error/config_broker.cc
+++ b/examples/10a.odd_error/config_broker.cc
@@ -1,0 +1,284 @@
+// Copyright Microsoft and CHERIoT Contributors.
+// SPDX-License-Identifier: MIT
+
+// Contributed by Configured Things Ltd
+
+#include "cdefs.h"
+#include "cheri.hh"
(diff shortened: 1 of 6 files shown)


PhilDay-CT:
Seems to be stack related - was still weird behaviour though that commenting out a line of code in an used function could change the bevavour

PR #270, opened 2024-07-12
Fix the memory safety example for the non-trapping behaviour.

davidchisnall:
Ages ago, we made store-local violations clear the tag rather than trap.

### DIFF
--- a/examples/08.memory_safety/memory_safety_inner.cc
+++ b/examples/08.memory_safety/memory_safety_inner.cc
@@ -12,7 +12,8 @@ using Debug = ConditionalDebug<true, "Memory safety compartment">;
 
 using namespace CHERI;
 
-char *allocation = NULL;
+char *allocation;
+static char *volatile volatilePointer;
 
 extern "C" ErrorRecoveryBehaviour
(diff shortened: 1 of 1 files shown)


PhilDay-CT:
Would be good to also update the ReadMe to explain that preventing them from being captured means that they they are invalidated if they are stored, rather than that they can't be stored.

davidchisnall:
Can you file an issue pointing to the places in the docs where we talk about these?

examples/08.memory_safety/memory_safety_inner.cc
```diff
@@ -136,11 +137,13 @@ int memory_safety_inner_entry(MemorySafetyBugClass operation)
...
 			Debug::log("Trigger storing a stack pointer {} into global",
 			           Capability{buf});
-			allocation = buf;
-
-			Debug::Assert(false,
+			volatilePointer = buf;
+			Capability tmp  = volatilePointer;
+			Debug::log("tmp: {}", tmp);
+			Debug::Assert(!tmp.is_valid(),
 			              "Code after storing stack pointer into global should "
```
rmn30, line 144: This message needs updating.

PR #272, opened 2024-07-12
update ReadMe for example 08

PhilDay-CT:
Update the ReadMe for example 08 to better reflect the impact of storing non global pointers in the heap/global

### DIFF
--- a/examples/08.memory_safety/memory_safety_inner.cc
+++ b/examples/08.memory_safety/memory_safety_inner.cc
@@ -140,9 +140,10 @@ int memory_safety_inner_entry(MemorySafetyBugClass operation)
 			volatilePointer = buf;
 			Capability tmp  = volatilePointer;
 			Debug::log("tmp: {}", tmp);
-			Debug::Assert(!tmp.is_valid(),
-			              "Code after storing stack pointer into global should "
-			              "be unreachable");
+			Debug::Assert(
+			  !tmp.is_valid(),
+			  "Code after accessing a stack pointer stored into global should "
+			  "be unreachable");
 			return tmp[0];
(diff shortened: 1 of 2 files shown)


davidchisnall (approved):

examples/08.memory_safety/memory_safety_inner.cc
```diff
@@ -140,9 +140,10 @@ int memory_safety_inner_entry(MemorySafetyBugClass operation)
 			volatilePointer = buf;
 			Capability tmp  = volatilePointer;
 			Debug::log("tmp: {}", tmp);
-			Debug::Assert(!tmp.is_valid(),
-			              "Code after storing stack pointer into global should "
-			              "be unreachable");
+			Debug::Assert(
+			  !tmp.is_valid(),
+			  "Code after accessing a stack pointer stored into global should "
```
rmn30, line 145: This message still needs fixing....

davidchisnall: Ooops.

PhilDay-CT: Doesn't my updated version capture what its now testing ?

davidchisnall: No, it isn't unreachable, which is why I added the dereference of the untagged pointer afterwards.  It should give back an untagged value.

PhilDay-CT: Ah, I was thinking that accessing tmp in the Debug at line 142 would trigger a tag violation, but of course you have to actually have to deference it  to get that

PR #277, opened 2024-07-26
Adding features to the OpenTitanUart driver

HU90m:
I fleshed out the OpenTitan UART driver a little while working on https://github.com/lowRISC/sonata-software/pull/26. I don't think this is quite ready to be merged yet; opening this purely for visibility.


davidchisnall:
There's a bunch of stuff in here that's just improving the existing code, rather than adding new things.  It would be great to have those merged soon (I don't anticipate problems with the other bits either).

HU90m:
> There's a bunch of stuff in here that's just improving the existing code, rather than adding new things. It would be great to have those merged soon (I don't anticipate problems with the other bits either).

:+1: Rebased onto https://github.com/microsoft/cheriot-rtos/pull/278#event-13694639924 and marked as ready for review.

davidchisnall (approved):

davidchisnall:
We probably need to document somewhere that this changes the behaviour of `stderr`, for anyone who is using `stdio.h`.

HU90m:
> We probably need to document somewhere that this changes the behaviour of `stderr`, for anyone who is using `stdio.h`.

Where would be a good place to document this change?

davidchisnall:
Good question.  A blog post maybe, and possibly in the Sonata bit of the [Getting Started](https://github.com/microsoft/cheriot-rtos/blob/main/docs/GettingStarted.md) doc?

PR #279, opened 2024-08-01
Add more examples of temporal safety

PhilDay-CT:
Following the recent thread with Murali on sub objects I though it would be useful to add some more use cases to the temporal_safety example to illustrate sub objects and also claims

### DIFF
--- a/examples/04.temporal_safety/README.md
+++ b/examples/04.temporal_safety/README.md
@@ -1,12 +1,14 @@
 Temporal safety example
 =======================
 
-This example shows a trivial use-after-free bug.
-The code simply allocates an object and frees it, printing the value of the pointer both before and after deallocation.
+This example shows an number of use-after-free cases.
+
+The first case simply allocates an object and frees it, printing the value of the pointer both before and after deallocation.
 
 The output from this should be something roughly like this:
 
 ```
+Allocating compartment: -- Simple Case --
 Allocating compartment: Allocated: 0x80004ce0 (v:1 0x80004ce0-0x80004d0a l:0x2a o:0x0 p: G RWcgm- -- ---)
 Allocating compartment: Use after free: 0x80004ce0 (v:0 0x80004ce0-0x80004d0a l:0x2a o:0x0 p: G RWcgm- -- ---)
 ```
@@ -43,3 +45,24 @@ Any of these can be cleared from a copy of the pointer before passing it to othe
 For example, if you remove `W` and `M` permissions from a pointer that you pass as a parameter then you have a guarantee that nothing reachable from the pointer will be mutated.
 Similarly, if you remove `G` and `L` then you have the guarantee that nothing reachable from the pointer will be captured.
 If you remove `G` but not `L` then you have the weaker guarantee that the pointer that you passed will not be captured but pointers reachable from it might be.
+
+The next three use cases show the handling of sub object, a capability that references a sub-range of the allocation.
+
+In the first of these the sub object is passed to free(), but this has no impact until the original allocation is also freed (i.e an allocation does not become fragmented if you try to free the middle of it).
+
+In the second use case a claim is made on the sub object, which counts against the quota but now means that the sub object remains valid even if the enclosing allocation is freed.
+Both the sub object and the enclosing allocation become invalid when the claim is releases. 
+
+The third use case shows the difference between a claim and a fast claim.
+Claims are persistent, count against the compartment's quota, and last until they are explicitly released, but they are expensive as they require a cross compartment call to the allocator.
+Fast claims are ephemeral and belong to the thread rather than the compartment.
+Each thread may only hold one fast claim (on up to two objects).
+They do not count against a quota, but they only last until the thread makes a cross compartment call or another fast claim.
+In the example the claim on the sub object is made with a fast claim, but when the enclosing object is now freed the fast claim is dropped (as free() is a cross compartment call) so both the enclosing and sub objects become invalid.
+_This is a poor use of a fast claim used to illustrate the behaviour; The normal use case is to establish a claim early in the entry to a compartment to prevent an object becoming invalid while the compartment processes it, which may include making a persistent claim._    
+
+The final use case shows how an object initially allocated in one compartment be claimed by (add counted against the quota) of a second compartment.
+This allow, for example, a zero copy data buffer pattern.
+Even when the allocation is freed by (and removed from the quota of) the original compartment it remain valid because it is now claimed by the second compartment.
+Note that the quota is reduced by more when making a claim than an allocation because a small amount of additional heap is required for the claim headers.   
+
--- a/examples/04.temporal_safety/README.md
+++ b/examples/04.temporal_safety/README.md
@@ -46,23 +46,27 @@ For example, if you remove `W` and `M` permissions from a pointer that you pass 
 Similarly, if you remove `G` and `L` then you have the guarantee that nothing reachable from the pointer will be captured.
 If you remove `G` but not `L` then you have the weaker guarantee that the pointer that you passed will not be captured but pointers reachable from it might be.
 
-The next three use cases show the handling of sub object, a capability that references a sub-range of the allocation.
+The next three use cases show the handling of a sub-object, a capability that references a sub-range of the allocation.
 
-In the first of these the sub object is passed to free(), but this has no impact until the original allocation is also freed (i.e an allocation does not become fragmented if you try to free the middle of it).
+In the first of these the sub object is passed to free().
+As no claims have been made, free() would take action only if a cap to the entire object were passed in.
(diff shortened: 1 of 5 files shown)


examples/04.temporal_safety/README.md
```diff
@@ -43,3 +45,24 @@ Any of these can be cleared from a copy of the pointer before passing it to othe
 For example, if you remove `W` and `M` permissions from a pointer that you pass as a parameter then you have a guarantee that nothing reachable from the pointer will be mutated.
 Similarly, if you remove `G` and `L` then you have the guarantee that nothing reachable from the pointer will be captured.
 If you remove `G` but not `L` then you have the weaker guarantee that the pointer that you passed will not be captured but pointers reachable from it might be.
+
+The next three use cases show the handling of sub object, a capability that references a sub-range of the allocation.
```
nwf-msr, line 49: Perhaps hyphenate, here and throughout?

```suggestion
The next three use cases show the handling of a sub-object, a capability that references a sub-range of the allocation.
```

examples/04.temporal_safety/README.md
```diff
@@ -43,3 +45,24 @@ Any of these can be cleared from a copy of the pointer before passing it to othe
...
+
+In the first of these the sub object is passed to free(), but this has no impact until the original allocation is also freed (i.e an allocation does not become fragmented if you try to free the middle of it).
```
nwf-msr, line 51: Hm.  I might expand that a bit.  Something like...

```suggestion
In the first of these the sub object is passed to free().
As no claims have been made, free() would take action only if a cap to the entire object were passed in.
Thus, this call to free() has no effect on the heap or on the pointers held by the client.
(In particular, unlike many historical implementations of malloc, freeing a sub-object will not erroneously return this sub-object's memory to the free pool.)
```

examples/04.temporal_safety/README.md
```diff
@@ -43,3 +45,24 @@ Any of these can be cleared from a copy of the pointer before passing it to othe
...
+
+In the second use case a claim is made on the sub object, which counts against the quota but now means that the sub object remains valid even if the enclosing allocation is freed.
+Both the sub object and the enclosing allocation become invalid when the claim is releases.
```
nwf-msr, line 54: "is freed" might mean several things.  Perhaps...

```suggestion
In the second use case a claim is made on the sub object.
This charges the claimant's quota and ensures that the sub-object (and, indeed, the entire object) remains allocated for the duration of the claim, even when the enclosing allocation is passed to free().
Thereafter, releasing the claim on the sub-object will cause the object to be freed, invalidating the pointers to the object and sub-object alike.
```

examples/04.temporal_safety/README.md
```diff
@@ -43,3 +45,24 @@ Any of these can be cleared from a copy of the pointer before passing it to othe
...
 If you remove `G` but not `L` then you have the weaker guarantee that the pointer that you passed will not be captured but pointers reachable from it might be.
+
+The next three use cases show the handling of sub object, a capability that references a sub-range of the allocation.
+
+In the first of these the sub object is passed to free(), but this has no impact until the original allocation is also freed (i.e an allocation does not become fragmented if you try to free the middle of it).
+
+In the second use case a claim is made on the sub object, which counts against the quota but now means that the sub object remains valid even if the enclosing allocation is freed.
+Both the sub object and the enclosing allocation become invalid when the claim is releases. 
+
+The third use case shows the difference between a claim and a fast claim.
```
nwf-msr, line 56: This might be good motivation to finally rename "fast" claims, perhaps before landing an updated version of this PR?  @davidchisnall?

examples/04.temporal_safety/README.md
```diff
@@ -43,3 +45,24 @@ Any of these can be cleared from a copy of the pointer before passing it to othe
...
+Claims are persistent, count against the compartment's quota, and last until they are explicitly released, but they are expensive as they require a cross compartment call to the allocator.
+Fast claims are ephemeral and belong to the thread rather than the compartment.
+Each thread may only hold one fast claim (on up to two objects).
```
nwf-msr, line 59:
```suggestion
Each thread may hold only at most one fast claim (on up to two objects).
```

examples/04.temporal_safety/README.md
```diff
@@ -43,3 +45,24 @@ Any of these can be cleared from a copy of the pointer before passing it to othe
...
+They do not count against a quota, but they only last until the thread makes a cross compartment call or another fast claim.
+In the example the claim on the sub object is made with a fast claim, but when the enclosing object is now freed the fast claim is dropped (as free() is a cross compartment call) so both the enclosing and sub objects become invalid.
+_This is a poor use of a fast claim used to illustrate the behaviour; The normal use case is to establish a claim early in the entry to a compartment to prevent an object becoming invalid while the compartment processes it, which may include making a persistent claim._    
+
+The final use case shows how an object initially allocated in one compartment be claimed by (add counted against the quota) of a second compartment.
```
nwf-msr, line 64:
```suggestion
The final use case shows how an object initially allocated in one compartment may be claimed by (add counted against the quota) of a second compartment.
```

examples/04.temporal_safety/README.md
```diff
@@ -43,3 +45,24 @@ Any of these can be cleared from a copy of the pointer before passing it to othe
...
+This allow, for example, a zero copy data buffer pattern.
```
nwf-msr, line 65:
```suggestion
This allows, for example, a zero-copy data buffer pattern.
```

examples/04.temporal_safety/README.md
```diff
@@ -43,3 +45,24 @@ Any of these can be cleared from a copy of the pointer before passing it to othe
...
+Even when the allocation is freed by (and removed from the quota of) the original compartment it remain valid because it is now claimed by the second compartment.
```
nwf-msr, line 66:
```suggestion
Even when the allocation is freed by (and removed from the quota of) the original compartment, it remains valid, because it is now claimed by the second compartment.
```

examples/04.temporal_safety/README.md
```diff
@@ -43,3 +45,24 @@ Any of these can be cleared from a copy of the pointer before passing it to othe
...
+Note that the quota is reduced by more when making a claim than an allocation because a small amount of additional heap is required for the claim headers.
```
nwf-msr, line 67:
```suggestion
Note that the quota charge for a claim on an object is slightly larger than the size of the object itself, because a small amount of additional heap is required for the claim headers.
```

examples/04.temporal_safety/allocate.cc
```diff
@@ -5,17 +5,141 @@
...
+	// Simple case
+	{
+		Debug::log("----- Simple Case -----");
+		void *x = malloc(42);
+		// Print the allocated value:
+		Debug::log("Allocated: {}", x);
+		free(x);
+		// Print the dangling pointer, note that it is no longer a valid pointer
+		// (v:0)
+		Debug::log("Use after free: {}", x);
```
nwf-msr, line 25: I know the original didn't, but it might be nice to sprinkle in some executable assertions about what the user should see, in addition to the commentary.

nwf-msr (commented):
Thanks!  This looks great and really helps fill in some gaps in our documentation.  Mostly comments on the prose; the code itself looks fine.

examples/04.temporal_safety/allocate.cc
```diff
@@ -5,17 +5,141 @@
...
+	// Using a claim in another compartment
+	{
+		Debug::log("----- Claim in another compartment -----");
+		void *x = malloc(10);
+
+		Debug::log("Allocated : {}", x);
+		Debug::log("heap quota: {}", heap_quota_remaining(MALLOC_CAPABILITY));
+
+		// Get the claimant compartment to make a fast claim
+		make_claim(x);
```
nwf-msr, line 121: Belatedly... Strictly speaking, FWIW, you don't need to have a separate compartment here; it would suffice to have a 2nd malloc capability.  I think maybe an aside in a comment to that effect would be useful, but also having it be an actual compartment is more reflective of how users will actually see this used in practice.

examples/04.temporal_safety/allocate.cc
PhilDay-CT: I could, but that's where I hit issue https://github.com/CHERIoT-Platform/llvm-project/issues/38  and most of the meaningful asserts would be on is_valid()

Trying to work around that with "volatile" only partly worked, and I think it distracts from the example

examples/04.temporal_safety/allocate.cc
PhilDay-CT: Yeah I thought making a second claim on the full object in the same compartment would just look odd.  I'll update the comment.

nwf-msr (approved):

nwf-msr (approved):

davidchisnall:
Thanks!

PR #283, opened 2024-08-13
Add initial support for pre-shared globals.

### LINKED ISSUE #282: Allow pre-shared objects [design proposal]
There are some use cases where it's convenient to pre-share a global.  We have a handful of places where it's useful to allow sharing objects where the (initial) sharing relationship is defined statically.  For example:

 - The restartable network stack has an epoch counter that various compartments use to determine the [re]initialisation status.
 - We expose various statistic counters in different compartments.
 - The config broker exposes a structure for each config item that contains an epoch counter and a pointer to the data.

Currently, these all share the same pattern: One compartment owns the global (or heap object, in a few cases) and hands out restricted capabilities to it.  This involves adding some extra function entry points that can hand out the capabilities and then dynamically setting the permissions.

In the loader, we have a generic mechanism for doing this currently, but it's restricted to memory-mapped I/O regions.  We explicitly check that these are not pointing to memory (with an explicit exemption for the allocator to let it access the hazards array).  I propose that we relax this and allow a set of shared objects to be defined in a shared-objects region of the program.

A shared object is not necessarily owned by a single compartment and so the proposed model separates out defining them and importing them.  Importing will use a slight modification of the existing MMIO macros (to eventually be replaced with compiler annotations).  

I propose that shared objects are created in xmake in a similar manner to threads (though collected from dependent builds, so we can define them in the build for a compartment).  Objects with the same name can be declared more than once, but must be declared with the same size.

The linker report should be extended to inspect the shared-object region and report the specific objects and sizes that are imported, rather than just addresses.

Pre-shared objects have the following restrictions:

 - As with MMIO objects, pointers to them cannot be used to initialise other globals.
 - They are always zero-initialised.
 - They must have globally-unique names within a firmware image.  The name does not have to be a valid C identifier.

davidchisnall:
This provides a mechanism to statically define globals that can be shared between compartments statically.  As with MMIO imports, these can be imported with restricted permissions.

They currently show up in the audit log as MMIO imports.  This is not ideal, we should plumb their names through into the audit log so that cheriot-audit can check that all references to a pre-shared object refer to exactly one object.

Fixes #282

### DIFF
--- a/sdk/include/compartment-macros.h
+++ b/sdk/include/compartment-macros.h
@@ -6,26 +6,27 @@
 #include <stdbool.h>
 
 /**
- * Helper macro, should not be used directly.
+ * Helper macro for MMIO and pre-shared object imports, should not be used
+ * directly.
  */
-#define MMIO_CAPABILITY_WITH_PERMISSIONS_HELPER(type,                          \
-                                                name,                          \
-                                                mangledName,                   \
-                                                permitLoad,                    \
-                                                permitStore,                   \
-                                                permitLoadStoreCapabilities,   \
-                                                permitLoadMutable)             \
+#define IMPORT_CAPABILITY_WITH_PERMISSIONS_HELPER(type,                        \
+                                                  name,                        \
+                                                  prefix,                      \
+                                                  mangledName,                 \
+                                                  permitLoad,                  \
+                                                  permitStore,                 \
+                                                  permitLoadStoreCapabilities, \
+                                                  permitLoadMutable)           \
 	({                                                                         \
-		volatile type *ret; /* NOLINT(bugprone-macro-parentheses) */           \
+		type *ret; /* NOLINT(bugprone-macro-parentheses) */                    \
 		__asm(".ifndef " mangledName "\n"                                      \
 		      "  .type     " mangledName ",@object\n"                          \
 		      "  .section  .compartment_imports." #name                        \
 		      ",\"awG\",@progbits," #name ",comdat\n"                          \
 		      "  .globl    " mangledName "\n"                                  \
 		      "  .p2align  3\n" mangledName ":\n"                              \
-		      "  .word __export_mem_" #name "\n"                               \
-		      "  .word __export_mem_" #name "_end - __export_mem_" #name       \
-		      " + %c1\n"                                                       \
+		      "  .word " #prefix #name "\n"                                    \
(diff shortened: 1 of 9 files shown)


sdk/xmake.lua
```diff
@@ -683,6 +683,36 @@ rule("firmware")
...
+		-- TODO: We should sort pre-shared globals by size to minimise padding.
+		-- Each global is emitted as a separate section so that we can use
+		-- CAPALIGN and let the linker insert the required padding.
+		local shared_objects_template =
+			"\n\t\t__cheriot_shared_object_section_${global} : CAPALIGN" ..
+			"\n\t\t{" ..
+			"\n\t\t\t__cheriot_shared_object_${global} = .;" ..
+			"\n\t\t\t. += ${size};" ..
+			"\n\t\t\t__cheriot_shared_object_${global}_end = .;" ..
+			"\n\t\t}\n"
```
nwf, line 709: Could we generalize the template to permit `__export_mem_hazard_pointers` and `__export_mem_allocator_epoch` to also be included in the `shared_objects` table, so that they get sorted for size along with the user defined ones?  Or, could we make those objects use this generic mechanism instead?

sdk/include/compartment-macros.h
```diff
@@ -78,6 +98,44 @@
...
 	MMIO_CAPABILITY_WITH_PERMISSIONS(type, name, true, true, false, false)
 
+/**
+ * Provide a capability of the type `type *` referring to the pre-shared object
+ * with `name` as its name.  This macro can be used only in code (it cannot be
+ * used to initialise a global).
+ *
+ * The last arguments specify the set of permissions that this capability
+ * holds.  Pre-shared objects are always global and without store local.  They
+ * may optionally omit additional permissions.
```
nwf, line 108: I don't see any commentary about the necessary corresponding `xmake` bits.  There's a worked example down in `tests/` but I think some prose might be in order?

sdk/xmake.lua
davidchisnall: Yes, next step is to make these use the new mechanism.

sdk/include/compartment-macros.h
davidchisnall: Yes, once this is merged I'll update the programmers' guide.

nwf (approved):

sdk/xmake.lua
davidchisnall: Added in the next commit.

PR #284, opened 2024-08-15
Reduce the permissions on hazard pointers.

davidchisnall:
The allocator is happy with just load + capability permissions.  Without store, it can't accidentally damage hazard lists.  Without load-mutable, it can't accidentally load a writeable capability.

This is purely defence in depth.  The allocator can still write to any heap memory, so not being able to load mutable capabilities doesn't restrict what it can do.  The allocator can still ignore hazard pointers, so not being able to write them is also no reduction in ability.  It simply constrains where bugs must be.

### DIFF
--- a/sdk/core/allocator/alloc.h
+++ b/sdk/core/allocator/alloc.h
@@ -1281,7 +1281,7 @@ class MState
 		// It is now safe to walk the hazard list.
 		Capability<void *> hazards =
 		  const_cast<void **>(SHARED_OBJECT_WITH_PERMISSIONS(
-		    void *, allocator_hazard_pointers, true, true, true, false));
+		    void *, allocator_hazard_pointers, true, false, true, false));
 		size_t pointers = hazards.length() / sizeof(void *);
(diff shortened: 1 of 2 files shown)


rmn30:
Nice! It doesn't even need to load readable capabilities but unfortunately we don't have a LoadReadable permission... Only comment is that since `SHARED_OBJECT_WITH_PERMISSIONS` is used with the same arguments in two locations it might be nice to define a macro for that.

rmn30:
I also wish that macro had more readable arguments that four booleans... We could have `enum LoadPerm {NoLoad, Load}` etc.

nwf-msr (approved):
LGTM, though I agree with Robert about auditability of "true, false, true, false".

davidchisnall:
I agree, but sadly macros don't understand C types.

davidchisnall:
(If you look at the implementation of the macro, it needs things that are understandable as booleans in both C preprocessor and GNU assembly, via string substitution, so it's pretty hard to use anything other than true and false)

PR #285, opened 2024-08-20
Fix a bug in the Sonata Ethernet driver.

davidchisnall:
The returned Ethernet frame was precisely bounded, but Ethernet frames can be all sorts of odd sizes and may not be representable with the capability bounds encoding.  This resulted in the tag being cleared, which caused a trap in the firewall.  The firewall is designed to fail closed, so this then prevented any further packets from arriving.

The bounding is purely defence in depth, so it's fine to use imprecise bounds here.

### DIFF
--- a/sdk/include/platform/sunburst/platform-ethernet.hh
+++ b/sdk/include/platform/sunburst/platform-ethernet.hh
@@ -711,7 +711,7 @@ class Ksz8851Ethernet
 			framesToProcess -= 1;
 
 			Capability<uint8_t> boundedBuffer{receiveBuffer.get()};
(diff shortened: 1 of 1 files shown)


marnovandermaas (approved):
This looks like a reasonable change to me. It would be good to get someone else to approve this as well though. Maybe @nbdd0121?

davidchisnall:
Probably not.  There isn't a security barrier here, it's purely defence in depth to set the bounds.  

I'm also seeing an issue on Sonata where IPv6 doesn't seem to work at all.  If I enable IPv6, I don't see a single incoming IPv6 packet, not even the reply to the router advertisement.

PR #288, opened 2024-08-29
Implement PWM Driver for Sonata Board

AlexJones0:
This PR implements a driver for the Pulse-Width Modulation (PWM) used on the Sonata board.

Note that the linked documentation source is currently out of date - I believe that @HU90m intends to update that to reflect the actual hardware, which is what this driver is based on.

I've confirmed that this is working for the 1 PWM that we have on Sonata by implementing the following in a simple piece of firmware with a single compartment: 
```cpp
#include <compartment.h>
#include <platform-pwm.hh>
#include <thread.h>

void __cheri_compartment("pwm_test") pwm_test()
{	
	auto pwm = MMIO_CAPABILITY(SonataPulseWidthModulation, pwm);
	uint8_t counter = 0;
	while (true) 
        {
		counter = (counter + 1) % 255;
		pwm->output_set(0, 255, counter);
		thread_millisecond_wait(50);
	}
}
```

Checking with a Multimeter I can then see analogue variation between 0V and 3.3V on the mikroBUS PWM output. 

Sonata currently only supports 1 PWM but this could potentially be changed, so I've designed the driver to be extensible for this case.

### DIFF
--- a/sdk/boards/sonata-prerelease.json
+++ b/sdk/boards/sonata-prerelease.json
@@ -63,6 +63,10 @@
         "plic": {
             "start" : 0x88000000,
             "end"   : 0x88400000
+        },
+        "pwm": {
+            "start" : 0x80001000,
+            "length": 0x00001000
         }
     },
     "instruction_memory": {
--- a/sdk/include/platform/sunburst/platform-pwm.hh
+++ b/sdk/include/platform/sunburst/platform-pwm.hh
@@ -1,0 +1,68 @@
+#pragma once
+#include <debug.hh>
+#include <stdint.h>
+
+/**
+ * A driver for Sonata's Pulse-Width Modulation (PWM).
+ *
+ * Documentation source can be found at:
+ * https://github.com/lowRISC/sonata-system/blob/97a525c48f7bf051b999d0178dba04859819bc5e/doc/ip/pwm.md
+ *
+ * Rendered documentation is served from:
+ * https://lowrisc.github.io/sonata-system/doc/ip/pwm.html
+ */
+struct SonataPulseWidthModulation
+{
+	/**
+	 * Flag to set when debugging the driver for UART log messages.
+	 */
+	static constexpr bool DebugDriver = false;
+
+	/**
+	 * Helper for conditional debug logs and assertions.
+	 */
+	using Debug = ConditionalDebug<DebugDriver, "PWM">;
+
+	/**
+	 * The number of pulse-width modulated outputs that are available.
+	 */
+	static constexpr size_t OutputCount = 1;
+
+	/**
+	 * The pulse-width modulation outputs available on Sonata.
+	 */
+	struct OutputRegisters
+	{
+		/**
+		 * The duty cycle of the wave, represented as a width counter. That
+		 * is, the number of clock cycles for which the signal will be on. The
+		 * duty cycle as a percentage is (duty cycle / period) * 100.
+		 */
+		uint32_t dutyCycle;
+
+		/**
+		 * The period (width) of the output block wave, set with the number of
+		 * clock cycles that one period should last. The maximum period is 255
+		 * as only an 8 bit counter is being used.
+		 */
+		uint32_t period;
+	} outputs[OutputCount];
+
+	/*
+	 * Sets the output of a specified pulse-width modulated output.
+	 *
+	 * The first argument is the index of the output. The second argument is
+	 * the period (length) of the output wave represented as a counter of
+	 * system clock cycles. The third argument is the number of clock cycles
+	 * for which a high pulse is sent within that period.
+	 *
+	 * So for example `output_set(0, 200, 31)` should set a 15.5% output.
+	 */
+	void output_set(uint32_t index, uint8_t period, uint8_t dutyCycle) volatile
+	{
+		Debug::Assert(index < OutputCount, "Specified PWM is out of range");
+		Debug::Assert(dutyCycle <= period, "Duty cycle cannot exceed 100%");
+		outputs[index].period    = period;
+		outputs[index].dutyCycle = dutyCycle;
+	}
+};


sdk/include/platform/sunburst/platform-pwm.hh
```diff
@@ -0,0 +1,66 @@
...
+		 * The period (width) of the block wave.
+		 */
+		uint32_t period;
+	} pwm[PwmCount];
+
+	/*
+	 * Sets the output of the specified PWM.
+	 */
+	void
+	set_pwm_output(uint32_t index, uint32_t period, uint32_t dutyCycle) volatile
```
HU90m, line 58: Since 255 is the maximum PWM period, I think it would be nicer to have `period` and `dutyCycle` be `uint8_t` and remove the `MaxPwmPeriod`. We don't need this to be parameterised.

sdk/include/platform/sunburst/platform-pwm.hh
```diff
@@ -0,0 +1,66 @@
...
+	 */
+	static constexpr bool DebugPwm = false;
+
+	/**
+	 * Helper for conditional debug logs and assertions.
+	 */
+	using Debug = ConditionalDebug<DebugPwm, "PWM">;
+
+	/**
+	 * The number of PWMs that are initialised.
```
HU90m, line 27:
```suggestion
	 * The number of PWM outputs available.
```

sdk/include/platform/sunburst/platform-pwm.hh
```diff
@@ -0,0 +1,66 @@
...
+	{
+		/**
+		 * The duty cycle of the PWM wave, represented as a width counter. That
+		 * is, the number of clock cycles for which the signal will be on. The
+		 * duty cycle as a percentage is (duty cycle / period) * 100.
+		 */
+		uint32_t dutyCycle;
+
+		/**
+		 * The period (width) of the block wave.
```
HU90m, line 49: A little more here would be nice. Something like: "The period of the output, set with the number of clock cycles a period should last."

sdk/include/platform/sunburst/platform-pwm.hh
```diff
@@ -0,0 +1,66 @@
...
+		 */
+		uint32_t period;
+	} pwm[PwmCount];
+
+	/*
+	 * Sets the output of the specified PWM.
```
HU90m, line 55: Please could you document the parameters of the function.

HU90m (changes requested):

sdk/include/platform/sunburst/platform-pwm.hh
AlexJones0: Ok, in that case since the Max PWM period is unlikely to change I'll remove the static variable and make the `period` and `dutyCycle` a `uint8_t`, and I'll just document the counter size somewhere so that it is clear when using the interface why a `uint8_t` is supplied if the register is a `uint32_t`.

sdk/include/platform/sunburst/platform-pwm.hh
AlexJones0: Agreed - also changing the name `pwmCount` to `pwmOutputCount` so this is more explicit that we are talking about possible PWM outputs and not individual devices.

sdk/include/platform/sunburst/platform-pwm.hh
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * Rendered documentation is served from:
+ * https://lowrisc.github.io/sonata-system/doc/ip/pwm.html
+ */
+struct SonataPwm
+{
+	/**
+	 * Flag to set when debugging the driver for UART log messages.
+	 */
+	static constexpr bool DebugPwm = false;
```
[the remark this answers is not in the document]
AlexJones0: Sure - in this case I was making the debug static variables follow the naming convention of the other Sonata drivers, e.g. `debugEthernet` or `debugOpenTitanI2c` or `debugSonataSpi`, but if you think that it's not worth it it here then I'm happy to change this.

sdk/include/platform/sunburst/platform-pwm.hh
```diff
@@ -0,0 +1,66 @@
+#pragma once
+#include <debug.hh>
+#include <stdint.h>
+
+/**
+ * A driver for Sonata's PWM.
```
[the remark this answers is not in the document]
AlexJones0: Understood - I intend to maintain the `Pwm` in the struct name so that we only have to use `SonataPwm` instead of `SonataPulseWidthModulation` every time we want to instantiate PWM - be sure to let me know if you want this changed as well. Instead, I've updated the high-level documentation to expand the abbreviation so that it is clear what it is referring to, and have mostly scrubbed references to the abbreviation from other parts of the code.

sdk/include/platform/sunburst/platform-pwm.hh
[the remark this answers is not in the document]
davidchisnall: Probably fine for the debug value, but not so much for the public APIs.

sdk/include/platform/sunburst/platform-pwm.hh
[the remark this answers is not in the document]
davidchisnall: No, please use `SonataPulseWidthModulation`.  If someone wants to use a shorter local name, they can use a `using` directive.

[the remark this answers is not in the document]
AlexJones0: Understood, I'll make that change now.

AlexJones0:
Thanks for the comments - I've updated the driver to address all the comments as far as I can tell, and I've updated the test code to fit the new interface. I can confirm again that based on the simple test the driver is working by observing the analogue output vary via a multimeter.

sdk/boards/sonata-prerelease.json
```diff
@@ -63,6 +63,10 @@
         "plic": {
             "start" : 0x88000000,
             "end"   : 0x88400000
+        },
+        "pwm": {
```
davidchisnall, line 67: @HU90m Is this also present in the 0.2 release?

davidchisnall (commented):
LGTM, I'm happy if @HU90m is happy.

sdk/boards/sonata-prerelease.json
AlexJones0: > @HU90m Is this also present in the 0.2 release?

I've just checked for @HU90m, and while the PWM existed in the code at the point of the 0.2 release, it wasn't hooked up to any pins in the RTL and so there's no way to use it. I think in this case it doesn't make sense to add this to the 0.2 board file also.

HU90m (approved):
@HU90m is happy

HU90m:
@davidchisnall this is awaiting your approval before it will merge

davidchisnall (approved):

PR #289, opened 2024-09-02
Implement ADC Driver for Sonata Board

AlexJones0:
This PR implements an Analogue-to-Digital Converter (ADC) driver for the Sonata Board's Xilinx ADC (XADC). 

The relevant documentation can be found on the [Sonata System documentation](https://lowrisc.github.io/sonata-system/doc/ip/adc.html) and in the [XADC Specification](https://docs.amd.com/r/en-US/ug480_7Series_XADC), and is referenced in comments through the driver code.

Sonata's XADC is configured to be initialised with default values that start it in "independent ADC mode", which will by default monitor system internals and sample connected analogue channels. As a result, this driver aims to provide a very simple interface for reading these registers, as well as changing the XADC's clock speed or powering it down so that you can reduce power draw if not using the ADC. I've tried to design the driver so that it should be easily extensible to support more features in the future, so that we can iterate and improve the driver if/as more control is required.

**Outdated**: ~I've marked this PR as a draft as I'd like some feedback on the way that I'm currently handling the MMIO capability for the ADC. As per the [CHERIoT Programmer's Guide](https://cheriot.org/book/top-drivers-top.html#mmio_capabilities), because the complete XADC has a fairly large control structure and I want to avoid defining a C struct covering every single field (many of which are invalid!), I initialise an `MMIO_CAPABILITY` called `registers` covering the entire range, which I then access using the class' methods. I can see that in the `generic-riscv/platform-plic.hh` driver, a `setField` lambda is used, which might be a nice alternative but might become a bit verbose if setting a lot of fields, and it doesn't allow you to as easily define generic behaviour over the Dynamic Reconfiguration Port (DRP) registers. I'm curious as to others' opinions on what would be a best practice for this in terms of readability.~

The XADC is only present in the most recent 0.4 release (only added a week or so ago), and so I've only added this to the `sonata-prerelease.json` board description file for now. 

I've confirmed that this is working by implementing the following piece of firmware within a single simple compartment:
```c++
#include <platform-adc.hh>
#include <debug.hh>
#include <compartment.h>
#include <thread.h>

using Debug = ConditionalDebug<true, "ADC_Testing">;
using SonataAdc = SonataAnalogueDigitalConverter;

[[noreturn]] void __cheri_compartment("adc_test") adc_entry()
{
    SonataAdc::ClockDivider divider = (CPU_TIMER_HZ / SonataAdc::MinClockFrequencyHz) / 2;
    SonataAdc *adc = new SonataAdc(divider, SonataAdc::PowerDownMode::None);

    while (true)
    {
    	Debug::log(
            "ADC A0: {}\t, A1: {}\t, A2: {}\t, A3: {}\t, A4: {}\t, A5: {}\t Temp: {}",
            adc->read_last_measurement(SonataAdc::MeasurementRegister::ArduinoA0),
            adc->read_last_measurement(SonataAdc::MeasurementRegister::ArduinoA1),
            adc->read_last_measurement(SonataAdc::MeasurementRegister::ArduinoA2),
            adc->read_last_measurement(SonataAdc::MeasurementRegister::ArduinoA3),
            adc->read_last_measurement(SonataAdc::MeasurementRegister::ArduinoA4),
            adc->read_last_measurement(SonataAdc::MeasurementRegister::ArduinoA5),
            adc->read_last_measurement(SonataAdc::MeasurementRegister::Temperature)
        );
        thread_millisecond_wait(30);
    }
}

```
Then, plugging in a power supply to each of the pins A0 through A5 and varying the voltage from 0V to 3.3V I can see the measured voltage being read by the ADC correspondingly vary as expected. Temperature measurement also seems to be consistent with expectations, giving a measurement of around 2600, corresponding to around 46-47°C. Varying the divider to different valid values still measures as expected, and all power down options are behaving as described.

### DIFF
--- a/sdk/include/platform/sunburst/platform-adc.hh
+++ b/sdk/include/platform/sunburst/platform-adc.hh
@@ -1,0 +1,353 @@
+#pragma once
+#include <debug.hh>
+#include <stdint.h>
+#include <utils.hh>
+
+/**
+ * A simple driver for Sonata's XADC (Xilinx Analogue to Digital Converter).
+ *
+ * Documentation source can be found at:
+ * https://github.com/lowRISC/sonata-system/blob/97a525c48f7bf051b999d0178dba04859819bc5e/doc/ip/adc.md
+ *
+ * Rendered documentation is served from:
+ * https://lowrisc.github.io/sonata-system/doc/ip/adc.html
+ */
+class SonataAnalogueDigitalConverter : private utils::NoCopyNoMove
+{
+	/**
+	 * Flag to set when debugging the driver for UART log messages.
+	 */
+	static constexpr bool DebugDriver = true;
+
+	/**
+	 * Helper for conditional debug logs and assertions.
+	 */
+	using Debug = ConditionalDebug<DebugDriver, "ADC">;
+
+	/**
+	 * Results of measurements / analogue conversions are stored in Dynamic
+	 * Reconfiguration Port (DRP) status registers as most-significant-bit
+	 * justified 12 bit values, represented by this mask.
+	 */
+	static constexpr uint16_t MeasurementMask = 0xFFF0;
+
+	/**
+	 * The location (offset) of the Xilinx Analogue-to-Digital Converter's
+	 * Dynamic Reconfiguration Port (DRP) registers, which are 16-bit registers
+	 * that are sequentially mapped to memory in 4-byte (word) intervals, in the
+	 * lower 2 bytes of each word. This includes both status (read-only) and
+	 * control (read/write) registers.
+	 *
+	 * https://docs.amd.com/r/en-US/ug480_7Series_XADC/XADC-Register-Interface
+	 */
+	enum class RegisterOffset : uint8_t
+	{
+		Temperature                      = 0x00,
+		VoltageInternalSupply            = 0x01,
+		VoltageAuxiliarySupply           = 0x02,
+		VoltageDedicated                 = 0x03,
+		VoltageInternalPositiveReference = 0x04,
+		VoltageInternalNegativeReference = 0x05,
+		VoltageBlockRamSupply            = 0x06,
+
+		/* Offset 0x07 is undefined. */
+
+		/* ADC A's calibration coefficient status registers omitted. */
+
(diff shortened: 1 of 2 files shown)


sdk/include/platform/sunburst/platform-adc.hh
```diff
@@ -0,0 +1,350 @@
...
+	 * mode.
+	 *
+	 * Takes the clock divider and power down modes to initialise. The
+	 * clock divider should divide the system clock to create a signal
+	 * between 1 and 26 MHz, and should be at least 2.
+	 */
+	SonataAnalogueDigitalConverter(ClockDivider  divider,
+	                               PowerDownMode powerDown)
+	{
+		registers = MMIO_CAPABILITY(uint32_t, adc);
```
davidchisnall, line 265: I think the overall approach is good, but this is an anti pattern. You are loading from the import table and then storing it in the memory here. Accessing the field is about as expensive as the original initialisation was. The typical pattern here is to add an always-inline function that wraps the macro.

AlexJones0: I see, thank you for the feedback - I see now the note in the CHERIoT Programmer's Guide about this anti-pattern. Am I understanding correctly that the inlining `MMIO_CAPABILITY` approach is more efficient because it expands to a single load capability load, whereas by loading once and storing to memory I am performing (more expensive) memory accesses equivalent to the original initialisation each time?

AlexJones0:
I've rebased on main, made changes to the driver's MMIO access based on the given feedback, and now marked the PR as being ready for review. Running the same tests as in the original post, the ADC driver still seems to be working as intended.

sdk/include/platform/sunburst/platform-adc.hh
davidchisnall: Yes, the `MMIO_CAPABILITY` macro will expand to an auipcc (typically the compressed version since it will need a tiny offset) followed by a load-capability.  The other one may be a single load-capability relative to the base, but it now means that we have a stateful object and everything that entails.

davidchisnall (approved):

PR #290, opened 2024-09-02
Sleep better with queues.

davidchisnall:
If a queue is full / empty and the end that can fix that is locked, we were not sleeping in the push / pop calls.


nwf (approved):
LGTM; good catch.

PR #291, opened 2024-09-02
Fix atomicity of Sonata ethernet driver.

davidchisnall:
The network stack expects to be able to receive frames in one thread and send them from another.  The Sonata Ethernet has some MMIO state that is shared between the two.

This is not a complete fix because other SPI devices can also be interleaved.  We see this sometimes when packets end up written into the LCD display.

### DIFF
--- a/sdk/include/platform/sunburst/platform-ethernet.hh
+++ b/sdk/include/platform/sunburst/platform-ethernet.hh
@@ -435,6 +435,12 @@ class Ksz8851Ethernet
 	RecursiveMutex receiveBufferMutex;
 
 	/**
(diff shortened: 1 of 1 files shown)


nwf:
I suppose, unless we get a hardware fix (managed chip-selects), we'd want a shared lock object for the SPI GPIO space, or to move all the SPI accesses into a singular compartment with private lock object?  That's an unfortunate bit of cross-coupling, but it is what it is.

davidchisnall:
Yes, it breaks the security model. We need a hardware fix from lowRISC.

PR #293, opened 2024-09-03
Add stdaligned.h & stdnoreturn.h.

sleffler:
Added in C11, deprecated in C23.

Pulled verbatim from OpenTitan.


davidchisnall:
I’d rather not add Apache licensed code to an MIT codebase, especially not for something where it’s unclear that copyright even applies.

sleffler:
No problem; you asked for a PR and this is the code I'm using.

PR #294, opened 2024-09-03
Add a linux-style bogomips estimator.

sleffler:
Now you can definitively answer the question: how fast is this simulator? (just kidding)

### DIFF
--- a/examples/10.bogomips/bogomips.cc
+++ b/examples/10.bogomips/bogomips.cc
@@ -1,0 +1,79 @@
+// Copyright Google LLC and CHERIoT Contributors.
+// SPDX-License-Identifier: MIT
+
+/*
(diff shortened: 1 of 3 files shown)


davidchisnall:
Is it possible for the license on these to match the rest of the project?

rmn30:
This is fun but unfortunately it appears to be derived from GPL 2 code which means we can't include it in this repo. Arguably it is sufficiently trivial that it is not copyrightable but IANAL.

If we do merge please also add a README in `10.bogomips` and point out that `thread.h` provides `thread_microsecond_spin` and `thread_millisecond_wait`.

sleffler:
This code has nothing remaining of consequence from the cited code and it
went through multiple iterations before reaching its current form.
Regardless I will just keep it personal. FWIW using this our mpact
simulator measures at 20x the perf of sail.

PR #296, opened 2024-09-06
Add memrchr to the string library.

sleffler:


davidchisnall (approved):
Looks good to me, thanks!

PR #297, opened 2024-09-16
Add some primitive unwinding support

davidchisnall:
This is two commits.  The first reserves space on the stack on each compartment transition so that you have space for per-thread per-invocation state.

The second adds some exception-like macros that are similar to Win16 SEH and OpenStep exceptions.

### DIFF
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -227,6 +227,8 @@ __Z26compartment_switcher_entryz:
 #endif
 	zero_stack         t2, s0, gp
 after_zero:
+	// Reserve space for unwind state and so on.
+	cincoffset 	       csp, csp, -16
 #ifdef CONFIG_MSHWM
 	// store new stack top as stack high water mark
 	csrw               CSR_MSHWM, sp
@@ -256,6 +258,8 @@ after_zero:
 	// At this point, we have already truncated the stack and so the length of
 	// the stack is the length that the callee can use.
 	cgetlen            t2, csp
+	// Include the space we reserved for the unwind state.
+	addi               t2, t2, -16
 	bgtu               tp, t2, .Lstack_too_small
 
 	// Get the flags field into tp
--- a/sdk/include/setjmp.h
+++ b/sdk/include/setjmp.h
@@ -1,0 +1,64 @@
+// Copyright Microsoft and CHERIoT Contributors.
+// SPDX-License-Identifier: MIT
+#pragma once
+
+/**
+ * This is a minimal implementation of setjmp/longjmp.
+ *
+ * CHERIoT cannot store a `jmp_buf` anywhere other than the stack without
+ * clearing tags (which will then cause `longjmp` to fail).
+ */
+
+#include <stddef.h>
+#include <stdint.h>
+
+/**
+ * Jump buffer for setjmp/longjmp.
+ */
+struct __jmp_buf
+{
+	uintptr_t __cs0;
+	uintptr_t __cs1;
+	uintptr_t __csp;
+	uintptr_t __cra;
+};
+
+/**
+ * C requires that `setjmp` and `longjmp` take a `jmp_buf` by reference and so
+ * this ends up being defined as an array of one element, which allows it to
+ * both be allocated and passed by reference.
+ */
+typedef struct __jmp_buf jmp_buf[1];
+
+/**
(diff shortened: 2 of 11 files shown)


marnovandermaas (commented):
Some minor comments from my side. It looks good to me but I don't feel qualified to approve this PR due to limited knowledge of the RTOS.

sdk/core/switcher/entry.S
```diff
@@ -164,7 +164,7 @@ switcher_scheduler_entry_csp:
 	.p2align 2
 	.type __Z26compartment_switcher_entryz,@function
 __Z26compartment_switcher_entryz:
-	cincoffset        csp, csp, -SPILL_SLOT_SIZE
+	cincoffset        csp, csp, -(SPILL_SLOT_SIZE)
```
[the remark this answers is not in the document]
davidchisnall: No, this is accidental.  I previously moved the stack pointer here, but it was not the right place and I didn't fully revert the change.

sdk/include/unwind.h
```diff
@@ -0,0 +1,98 @@
...
+__always_inline static inline void cleanup_unwind(void)
+{
+	CleanupList **__head = cleanup_list_head();
+	CleanupList  *__top  = *__head;
+	*__head              = __top->next;
+	longjmp(&__top->env, 1);
+}
+
+/**
+ * Simple error hanlding macros.  These are modelled on the OpenStep exception
```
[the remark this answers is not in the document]
marnovandermaas: Resolved now

sdk/core/switcher/entry.S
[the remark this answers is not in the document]
marnovandermaas: Resolved now

sdk/include/setjmp.h
```diff
@@ -0,0 +1,64 @@
...
+        ".globl longjmp\n"
+        ".p2align 2\n"
+        ".type longjmp,@function\n"
+        "longjmp:\n"
+        "	clc	cs0, 0(ca0)\n"
+        "	clc	cs1, 8(ca0)\n"
+        "	clc	csp, 16(ca0)\n"
+        "	clc	cra, 24(ca0)\n"
+        "	mv	a0, a1\n"
+        "	cjr	cra\n");
```
rmn30, line 64: Very minor nit: you've spelled this  `cret` above and `cjr` here.  I think I can see why but it's a bit weird...

davidchisnall: Yes, I prefer having it expanded here because we're not returning to the caller (as `cret` would imply), we're jumping to the cra value that we've just loaded.

sdk/include/unwind.h
```diff
@@ -0,0 +1,97 @@
...
+	CleanupList *next;
+	/// Jump buffer to return to.
+	__jmp_buf env;
+};
+
+/**
+ * Head of the cleanup list.
+ *
+ * This is stored in the space that the switcher reserves at the top of the
+ * stack.
```
nwf, line 20: Add some commentary about this being guaranteed to be zero'd because it's managed by the switcher?  Otherwise it looks, from C, like the head is on stack and never initialized.

sdk/include/setjmp.h
```diff
@@ -0,0 +1,64 @@
+// Copyright Microsoft and CHERIoT Contributors.
+// SPDX-License-Identifier: MIT
+#pragma once
+
+/**
+ * This is a minimal implementation of setjmp/longjmp.
+ *
+ * CHERIoT cannot store a `jmp_buf` anywhere other than the stack without
+ * clearing tags (which will then cause `longjmp` to fail.
```
nwf, line 9: Missing close paren.

nwf (approved):

sdk/include/unwind.h
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * Head of the cleanup list.
+ *
+ * This is stored in the space that the switcher reserves at the top of the
+ * stack.  The stack is zeroed on entry to a compartment and so this will be
+ * null until explicitly written to.
+ */
+__always_inline static inline struct CleanupList **cleanup_list_head()
+{
+	void     *csp = __builtin_cheri_stack_get();
```
rmn30, line 25: Does the first invocation of a thread entry go through the switcher path that reserves the space on stack? If not this might overwrite stack pointer and link register stored by the entry function. Maybe no big deal but perhaps unexpected.

PR #300, opened 2024-09-17
Allow the error invocation handler count to be reset.

### LINKED ISSUE #299: Compartments should be able to reset their handler invocation count
Currently, the switcher has a fixed limit per compartment invocation of the number of times that a compartment may invoke a handler.  This is intended to address cases where recovery is not complete.  There are two problems with this:

 - Entry point compartments are invoked precisely once, so each error-handler invocation will consume a finite resource.
 - If, after #297 is merged, you unwind out of an error handler, the switcher doesn't record this and will force unwind the next time you fault.

The simplest fix is to expose an API from the switcher to allow the count to be reset.

davidchisnall:
This lets long-running compartment invocations (for example, thread entry points) determine that they are in a known-valid state.

Fixes #299

### DIFF
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -992,6 +992,25 @@ __Z25stack_lowest_used_addressv:
 	csrr               a0, CSR_MSHWM
 	cret
 
+	.section .text, "ax", @progbits
+	.p2align 2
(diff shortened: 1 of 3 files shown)


nwf (approved):
LGTM

PR #301, opened 2024-09-19
Add support for stackless error handlers.

### LINKED ISSUE #298: Better handling for stack overflow
Currently, most traps can be handled by an error handler, which is modelled on a UNIX signal handler (though used only for synchronous events).  This works well in a lot of cases, but there are two annoying cases:

 - We have not quite exhausted the stack, but don't have space for the register spill frame.
 - We have exhausted the stack.

With #297, we can now unwind out of an error handler and run cleanup higher up the stack in a controlled way, which means that a lot of error handlers will not need the context at all.

We could simply pass null as the context the argument if we're in the first case (install context is then not allowed), but there's an annoying corner case between the two: If the error handler spills enough registers to fall off the stack.

There are a few options here:

Option 1: Make the unwinding mechanism part of the error handling logic in the switcher.  In the switcher, we try to pop an error handler and invoke it, if either there is no error handler or if there is not enough stack space to invoke one.

Option 2: We add a second error handler entry point that runs either if the current one is not provided or if there is insufficient stack space it to run.  This can be implemented in assembly and use no stack.

Option 3: Something I haven't thought of.

Opinions?

davidchisnall:
These are given a stack bounded it was on entry into the compartment, with an undefined offset.  As such, they can be used even when the cause of the error is stack overflow.

Unlike the existing error handler, this cannot take a snapshot of the register file, because there is no space for it.

Both kinds of error handler can be present.  When both are present, the existing kind will be used if there is sufficient stack space to invoke it, otherwise the stackless one will be invoked.

This also adds a 'library' that provides an implementation of a stackless error handler that interoperates with the new setjmp-based unwind code.  When an error occurs, it pops the top jump buffer and invokes it.

A few other small cleanups while I was touching adjacent things:

 - There were some issues in the switcher API to recover the initial stack, the tests were passing largely by coincidence.  This code is now fixed and the tests improved.
 - A load of recently added libraries were not linked to in the README for libraries.  They are now.

Fixes #298


sdk/core/switcher/entry.S
```diff
@@ -880,8 +933,21 @@ __Z22switcher_recover_stackv:
 	// two instructions.
 	cspecialr          ca0, mtdc
 	clhu               a1, TrustedStack_offset_frameoffset(ca0)
+	addi               a1, a1, -TrustedStackFrame_size
 	cincoffset         ca0, ca0, a1
 	clc                ca0, TrustedStackFrame_offset_csp(ca0)
+	// If this is the first frame, then the recovered stack will be the stack
+	// on entry.  If this is not the first frame then then we need
```
rmn30, line 940: Please complete this comment. A bit more explanation of the lines below would be nice too. Why are these the correct bounds?

sdk/lib/unwind_error_handler/README.md
```diff
@@ -0,0 +1,9 @@
+Unwind error handler library
+----------------------------
+
+This library provides an error handler that unwinds using the APIs in [`unwind.h`](../../include/unwind.h).
+
+This error handler can work even in cases of stack overflow.
+
+Note: Unlike most libraries, this is *not* built as a shared library.
```
rmn30, line 8: Maybe mention why not?

rmn30 (commented):
Looks good but a bit scary. Really wish we had some formal methods for the switcher.

sdk/core/loader/boot.cc
```diff
@@ -855,6 +855,8 @@ namespace
 
 			// Stack pointer points to the top of the stack.
 			stack.address() += config.stack.size();
+			// Reserve space at the start of the stack for error handling and so on.
+			stack.address() -= 16;
```
nwf, line 859: This `16` needs to agree with the `16` at https://github.com/CHERIoT-Platform/cheriot-rtos/blob/e4ebccc60840de78a919c42e6dd7045c99b68245/sdk/core/switcher/entry.S#L231 .  Can we get a named constant (`#define` is fine)?

sdk/core/switcher/entry.S
davidchisnall: Done.

sdk/lib/unwind_error_handler/README.md
davidchisnall: Done.

sdk/core/loader/boot.cc
davidchisnall: Done.

sdk/include/fail-simulator-on-error.h
```diff
@@ -59,7 +59,7 @@ compartment_error_handler(ErrorState *frame, size_t mcause, size_t mtval)
 		if (registerNumber == CHERI::RegisterNumber::CRA &&
 		    returnCapability.address() == 0 &&
 		    exceptionCode == CHERI::CauseCode::TagViolation &&
-		    stackCapability.top() == stackCapability.address())
+		    (stackCapability.top() - 16) == stackCapability.address())
```
rmn30, line 62:
```suggestion
		    (stackCapability.top() - STACK_RESERVED_SPACE) == stackCapability.address())
```
or at least a comment explaining the magic 16.

rmn30: Thanks for the comment. I guess you don't want to include trusted stack headers here but by not using the symbolic name someone could change the value of `STACK_ENTRY_RESERVED_SPACE` and miss this. Consequences are not too grave, but you could at least use the symbol name in the comment so that it shows up in search?

davidchisnall: CI fails if you get this wrong.

sdk/compartment.ldscript
```diff
@@ -13,7 +13,9 @@ SECTIONS
 		# delta.  The final layout will be the compartment import table
 		# followed by the text segment.  There won't be any padding, because
 		# the compartment import table is more strongly aligned than text.
-		LONG(DEFINED(compartment_error_handler) ? compartment_error_handler - __compartment_code_start + SIZEOF(.compartment_import_table) : -1);
+		SHORT(DEFINED(compartment_error_handler) ? compartment_error_handler - __compartment_code_start + SIZEOF(.compartment_import_table) : 0xffff);
```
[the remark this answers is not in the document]
davidchisnall: I don't think so, but the linker script puts it at the start anyway.  It has to go before anything that has relaxations.

sdk/core/switcher/entry.S
```diff
@@ -597,36 +597,86 @@ exception_entry_asm:
...
-	cincoffset         ct0, ct0, -(16*8)
 	// See if we can find a handler:
 	clhu               tp, TrustedStack_offset_frameoffset(csp)
 	li                 t1, TrustedStack_offset_frames
 	beq                tp, t1, .Lend_of_stack
 	addi               tp, tp, -TrustedStackFrame_size
+
 	// ctp points to the current available trusted stack frame.
 	cincoffset         ctp, csp, tp
 	// ct1 now contains the export table for the callee
```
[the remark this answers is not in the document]
davidchisnall: Done

sdk/compartment.ldscript
```diff
@@ -34,6 +36,7 @@ SECTIONS
 		# If there is a compartment error handler, make sure that it is before
 		# anything that can have linker relaxations so that its displacement
 		# from __compartment_code_start is a constant.
+		*(.compartment_error_handler_stackless));
 		*(.compartment_error_handler);
```
rmn30, line 40: If the stackless error handler contains a linker relaxation won't this violate the assertion in the comment?

davidchisnall: Yes, but the stackless one needs to be written in assembly.

sdk/compartment.ldscript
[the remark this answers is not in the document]
rmn30: I think it could happen if someone wrote a very long stackless error handler. You could use an `ASSERT` directive to check for this.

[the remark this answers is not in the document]
davidchisnall: In theory, someone could write a 64 KiB stackless error handler *and* want a normal error handler.  I can't help feeling that anyone who does this deserves everything that they get.

sdk/core/switcher/entry.S
```diff
@@ -597,36 +597,85 @@ exception_entry_asm:
...
+	cgettag            t1, ct0
+	// If there isn't enough space on the stack, see if there's a stackless
+	// handler.
+	beqz               t1, .Ltry_stackless_handler
+
 	clc                ct1, TrustedStackFrame_offset_calleeExportTable(ctp)
-	// Reset the export table pointer to point to the *start* of the export
+	// Set the export table pointer to point to the *start* of the export
 	// table.  It will currently point to the entry point that was raised.
 	// TODO: We might want to pass this to the error handler, it might be
```
rmn30, line 631: Minor: I think you could lift this common code before the branch. If you need an extra register maybe you could use `a0` which only contains 1 or 0?

davidchisnall: a0 needs to be live out at the branch that may enter the second path.

rmn30 (approved):

vmurali:
> Looks good but a bit scary. Really wish we had some formal methods for the switcher.

I would post the Simpsons hiding meme again, but the reality is formal methods always take the backseat compared to building new things :( I am hoping to get back to it at some point in the near future.

PR #306, opened 2024-10-03
Don't invoke error handlers when the callee errors.

### LINKED ISSUE #302: Remove error handler invocation as a result of an unwind
Currently, we have the following flow:

1. Compartment A calls compartment B
2. A fault occurs in B.
3. If we force unwind out of B *and* A has an error handler, A's error handler is invoked.

This was useful in the test suite for checking that tests do not terminate unexpectedly.  So far, it has been annoying in 100% of all other cases, and every error handler that we've written has a special case to silently ignore it.

This might not be the worst idea I ever had, but it's definitely near the top of the list.  Removing this path would also simplify the switcher code.

davidchisnall:
I originally made a forced unwind out of a compartment invoke the caller's error handler (if it existed) with a special error code.  The idea was that you could use error handlers to jump to recovery paths, rather than checking return values.

This seemed like a good idea at the time, but turned out to be a mistake.  The only code that has ever used this behaviour was the test suite.  Other error handlers have special-case logic to check if this is the reason and ignore it if so.  It also added a lot of confusing control flow in the switcher, which is bad because the switcher is the core of the TCB.

Fixes #302


marnovandermaas (commented):
It might be good to split this commit into two. One changing the error handling and the other changing all the void function returns into integers.

Any particular reason that you prefer a function always returning zero to a function that returns void?

davidchisnall:
It is necessary for the caller to be able to detect failure.  With void return, we were relying on the error handler being invoked to determine whether things failed.  The error handler is no longer invoked when the callee force unwinds, this is signalled by returning an error value.  If the return type is `void`, there is no way for the caller to extract that register value.

These are not independent changes.

nwf (approved):
The diff to `entry.S` hard to read by itself, but the resulting file seems sensible. LGTM.

PR #311, opened 2024-10-07
Make the headers more bare-metal friendly.

davidchisnall:
This should allow the freestanding headers from CHERIoT RTOS to be used for bare-metal compilation.

### DIFF
--- a/sdk/include/cdefs.h
+++ b/sdk/include/cdefs.h
@@ -61,7 +61,6 @@ using _Bool = bool;
 #define __section(x) __attribute__((section(x)))
 #define __alloc_size(x) __attribute__((alloc_size(x)))
 #define __alloc_align(x) __attribute__((alloc_align(x)))
-#define __cheri_callback __attribute__((cheri_ccallback))
 #if __has_attribute(cheriot_minimum_stack)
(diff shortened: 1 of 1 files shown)


sdk/include/cdefs.h
```diff
@@ -61,23 +61,41 @@ using _Bool = bool;
...
 #	define __cheri_compartment(x)
 #else
 #	define __cheri_compartment(x) __attribute__((cheri_compartment(x)))
 #endif
-#define __cheri_libcall __attribute__((cheri_libcall))
+
+// Define the CHERIoT calling-convention attributes macros to nothing if we're
+// targeting bare metal and to the correct attributes if we're targeting the
+// RTOS.
+#ifdef __CHERIOT_BAREMETAL__
```
sleffler, line 84: "__CHERIOT_BAREMETAL__" or "_CHERIOT_BAREMETAL_"? (I originally used the latter)

davidchisnall: I think the right thing is to make the compiler emit it, in which case using `__` as the prefix that the GNU conventions reserve for the compiler, rather than `_C` (which the GNU conventions reserve for the standard library) makes more sense.

davidchisnall: See CHERIoT-Platform/llvm-project#43

PR #312, opened 2024-10-09
Capability.bounds().set_inexact_at_most() fixes

nwf:
- Don't take bounds by reference (rely on the caller to get the resulting length afterwards) for consistency with the other APIs here.

- We were not falling into the bit-twiddling path in the case that the length could not fit in a mantissa, and so could return an untagged result.  Instead of trying to do all the requisite logic to see if setting bounds would succeed, just try to set the bounds and see if it worked.

- Update tests to reflect new interface, and add another that would have caught the above bug.

- Move the bit math out to a separate function, giving the compiler some more flexibility in inlining.

### DIFF
--- a/sdk/include/cheri.hh
+++ b/sdk/include/cheri.hh
@@ -414,12 +414,12 @@ namespace CHERI
 		 */
 		class PropertyProxyBase
 		{
+			protected:
 			/**
 			 * The capability that this proxy refers to.
 			 */
 			Capability &cap;
 
-			protected:
 			/**
 			 * Replaces the underlying capability
 			 */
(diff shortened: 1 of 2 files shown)


sdk/include/cheri.hh
```diff
@@ -546,54 +546,67 @@ namespace CHERI
 				return *this;
 			}
 
+			private:
+			BoundsProxy &set_inexact_at_most_slow(size_t bounds)
+			{
+				ptraddr_t newBaseAddress =
+				  static_cast<ptraddr_t>(reinterpret_cast<uintptr_t>(ptr()));
```
davidchisnall, line 553: Can we use the address get intrinsic instead of these casts?

davidchisnall (approved):

sdk/include/cheri.hh
nwf: Sure.  Out of curiosity, why is `PropteryProxyBase` `private` and not `protected`?  `cap.address()` seems nicer than `__builtin...`?

davidchisnall: Cap.address is fine too.

It’s private because I didn’t see a use case for subclasses accessing it, it’s an implementation detail.

PR #314, opened 2024-10-16
More defence in the thread unwind path.

davidchisnall:
We were spilling the trusted stack pointer to the csp slot in the thread's register save area.  This shouldn't matter because you can't resume a thread that has exited, but it's good defence in depth to not do this and spill 0 instead.


rmn30 (approved):
Apologies for not spotting this before approving previous PR!

PR #316, opened 2024-10-17
Separate static and dynamic sealing types.

davidchisnall:
This separates the sealing hardware sealing types used for heap-allocated objects from the sealing type used for static sealed objects.  The main goal of this was to remove the allocator from the TCB for statically sealed objects but, as a side benefit, we also delete about 200 lines of code and make the binaries around 100 bytes smaller.

We had previously reserved a type for the scheduler's sealed types to remove the allocator from the scheduler's TCB for static objects.  This is no longer required.  The allocator is in the scheduler's TCB for dynamically sealed objects (but it is anyway, because they live on the heap), but not for statically sealed objects.  This removes a load of special-case code from the scheduler.

This also means that we didn't increase the number of hardware sealing types used, because the scheduler no longer needs its own one.

The existing APIs are preserved, but the token unsealing APIs now have variants that will work *only* on static or dynamically sealed things and so users can remove allocator from the TCB for other objects that should be only statically sealed (and auditable), not heap allocated.

### DIFF
--- a/sdk/core/scheduler/main.cc
+++ b/sdk/core/scheduler/main.cc
@@ -348,7 +348,7 @@ namespace sched
 	template<typename T>
 	int typed_op(void *sealed, auto &&fn)
 	{
-		auto *unsealed = Handle::unseal<T>(sealed);
+		auto *unsealed = T::template unseal<T>(sealed);
 		// If we can't unseal the sealed capability and have it be of the
 		// correct type then return an error.
 		if (!unsealed)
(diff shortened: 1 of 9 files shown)


sdk/core/scheduler/main.cc
```diff
@@ -668,29 +650,18 @@ __cheriot_minimum_stack(0xc0) int multiwaiter_wait(Timeout           *timeout,
...
-		/**
-		 * Type marker used by `Handle`, tells it to use the dynamic path.
-		 */
-		static constexpr auto TypeMarker = Handle::Type::Dynamic;
-
 		/**
 		 * Dynamic type marker used by `Handle`.
 		 */
-		static Capability<void> dynamic_type_marker()
+		static SKey dynamic_type_marker()
```
nwf, line 660: I understand (now) that this is meant as "(dynamic type) marker", but maybe just call it `sealing_type` or `sealing_key` by similarity with `STATIC_SEALING_TYPE` or `SKey`/`InterruptKey`/`MultiWaiterKey`.

nwf (commented):
Woo, that's a fair bit of deleted code.  :tada:

Just some nits; otherwise it looks quite good.

sdk/core/scheduler/main.cc
davidchisnall: Probably a good idea.  The 'dynamic' name is left over from the internal version where we had both static and dynamic types.

davidchisnall:
I believe all review comments are addressed.

PR #320, opened 2024-10-18
WIP: various tweaks to, and a pile of documentation for, the switcher and exception handler

nwf:
The switcher is a very small, and yet also big, chunk of carefully hand-written, security-critical assembler.  In absence of formal verification of correctness, add more commentary about what's going on.

At the time of writing, this just adds "Register Atlas"es to the `__Z26compartment_switcher_entryz` entry point (and very slightly moves some code around).

I would also like to have "Control Flow Atlas"es with backwards arcs from labels with interesting control flow, but I haven't gotten there yet.

Mostly posting for bikeshedding and commentary; commit early, commit often.

### DIFF
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -197,10 +197,6 @@ __Z26compartment_switcher_entryz:
 	// The caller should back up all callee saved registers.
 	// mtdc should always have an offset of 0.
 	cspecialr          ct2, mtdc
-#ifndef NDEBUG
-	// XXX: This line is useless, only for mtdc to show up in debugging.
-	cmove              ct2, ct2
-#endif
 	clear_hazard_slots ct2, ctp
 
 	// make sure the trusted stack is still in bounds
@@ -438,10 +434,6 @@ exception_entry_asm:
 	// interrupts off and precious few registers available to us, so swap it
 	// with the csp (we'll put it back, later).
 	cspecialrw         csp, mtdc, csp
-#ifndef NDEBUG
-	// XXX: This move is useless, but just for debugging in the simulator.
-	cmove              csp, csp
-#endif
 
 	// If we read out zero, we've reentered the exception and are about to
 	// trap.  Make sure that we end up in an architectural trap loop: clobber
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -349,15 +349,14 @@ switcher_after_compartment_call:
 	clear_hazard_slots ctp, ct2
 	// make sure there is a frame left in the trusted stack
 	clhu               t2, TrustedStack_offset_frameoffset(ctp)
-	li                 tp, TrustedStack_offset_frames
+	li                 t0, TrustedStack_offset_frames
 	// Move to the previous trusted stack frame.
 	addi               t2, t2, -TrustedStackFrame_size
 	// If this is the first trusted stack frame, then the csp that we would be
 	// loading is the csp on entry, which does not have a spilled area.  In
 	// this case, we would fault when loading, so would exit the thread, but we
 	// should instead gracefully exit the thread.
-	bgeu               tp, t2, .Lcommon_defer_irqs_and_thread_exit
-	cspecialr          ctp, mtdc
+	bgeu               t0, t2, .Lcommon_defer_irqs_and_thread_exit
 	cincoffset         ct1, ctp, t2
 	// Restore the stack pointer.  All other spilled values are spilled there.
 	clc                csp, TrustedStackFrame_offset_csp(ct1)
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -614,12 +614,6 @@ exception_entry_asm:
 	// We're now out of the exception path, so make sure that mtdc contains
 	// the trusted stack pointer.
 	cspecialw   mtdc, csp
-	// Store an error value in return registers, which will be passed to the
-	// caller on unwind.  They are currently undefined, if we leave this path
-	// for a forced unwind then we will return whatever is in ca0 and ca1 to
-	// the caller so must ensure that we don't leak anything.
-	li                 a0, -1
-	li                 a1, 0
 
 	// We want to make sure we can't leak any switcher state into error
 	// handlers, so if we're faulting in the switcher then we should force
@@ -790,17 +784,10 @@ exception_entry_asm:
 	 */
 	csrci              mstatus, 0x8
 
-	// Move the return value to a register that will be cleared in a forced
-	// unwind and zero the return registers.
-	move               s0, a0
-	// Store an error value in return registers, which will be passed to the
-	// caller on unwind.
-	li                 a0, -1
-	li                 a1, 0
 	// Return values are 0 for install context, 1 for forced unwind.  Anything
 	// that is not either of these is invalid and so we should do a forced
 	// unwind anyway.
-	bne                s0, zero, .Lforce_unwind
+	bnez               a0, .Lforce_unwind
 
 	// We have been asked to install the new register context and resume.
 	// We do this by copying the register frame over the save area and entering
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -121,28 +121,6 @@ switcher_scheduler_entry_csp:
 .endm
 
 /**
- * Verify the compartment stack is valid, with the expected permissions and
- * unsealed.
- * This macro assumes t2 and tp are available to use.
- */
-.macro check_compartment_stack_integrity reg
-	// Check that the caller's CSP is a tagged, unsealed capability (with at
-	// least load permission - we'll check the other permissions properly
-	// later) by loading a byte.  If this doesn't work, we'll fall off this
-	// path into the exception handler and force unwind.
-	clb                t2, 0(\reg)
-	// make sure the caller's CSP has the expected permissions
-	cgetperm           t2, \reg
-	li                 tp, COMPARTMENT_STACK_PERMISSIONS
-	bne                tp, t2, .Lforce_unwind
-	// Check that the top and base are 16-byte aligned
-	cgetbase           t2, csp
-	or                 t2, t2, sp
-	andi               t2, t2, 0xf
-	bnez               t2, .Lforce_unwind
-.endm
-
-/**
  * Zero the stack.  The three operands are the base address, the top address,
  * and a scratch register to use.  The base must be a capability but it must
  * be provided without the c prefix because it is used as both a capability
@@ -197,11 +175,25 @@ __Z26compartment_switcher_entryz:
 	csc               cgp, SPILL_SLOT_cgp(ct2)
 	csc               cra, SPILL_SLOT_pcc(ct2)
 	cmove             csp, ct2
-	// before we access any privileged state, we can verify the
-	// compartment's csp is valid. If not, force unwind.
-	// Note that this check is purely to protect the callee, not the switcher
-	// itself.
-	check_compartment_stack_integrity csp
+
+	/*
+	 * Before we access any privileged state, we can verify the
+	 * compartment's csp is valid. If not, force unwind.  Note that this
+	 * check is purely to protect the callee, not the switcher itself.
+         *
+	 * Make sure the caller's CSP has the expected permissions and that its
+	 * top and base are 16-byte aligned.  We have already checked that it is
+	 * tagged and unsealed and 8-byte aligned by virtue of surviving the
+	 * stores above.
+	 */
+	cgetperm           t2, csp
+	li                 tp, COMPARTMENT_STACK_PERMISSIONS
+	bne                tp, t2, .Lforce_unwind
+	cgetbase           t2, csp
+	or                 t2, t2, sp
+	andi               t2, t2, 0xf
+	bnez               t2, .Lforce_unwind
+
 	// The caller should back up all callee saved registers.
 	// mtdc should always have an offset of 0.
 	cspecialr          ct2, mtdc
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -298,10 +298,6 @@ __Z26compartment_switcher_entryz:
 	// Load the target PCC and point to the function.
 	clc                cra, ExportTable_offset_pcc(ct1)
 	cincoffset         cra, cra, s0
-	// Get the number of registers to zero in t2
-	andi               t2, tp, 0x7
-	// Get the interrupt-disable bit in t1
-	andi               t1, tp, 0x10
 	// Zero any unused argument registers
 	// The low 3 bits of the flags field contain the number of arguments to
 	// pass.  We create a small sled that zeroes them and jump into the middle
@@ -310,6 +306,7 @@ __Z26compartment_switcher_entryz:
 .Lload_zero_arguments_start:
 	auipcc             cs0, %cheriot_compartment_hi(.Lzero_arguments_start)
 	cincoffset         cs0, cs0, %cheriot_compartment_lo_i(.Lload_zero_arguments_start)
+	andi               t2, tp, 0x7 // loader/types.h's ExportEntry::flags
 	// Change from the number of registers to pass into the number of 2-byte
 	// instructions to skip.
 	sll                t2, t2, 1
@@ -321,6 +318,7 @@ __Z26compartment_switcher_entryz:
 .Lzero_arguments_start:
 	zeroRegisters      a0, a1, a2, a3, a4, a5, t0
 	// Enable interrupts of the interrupt-disable bit is not set in flags
+	andi               t1, tp, 0x10
 	bnez               t1, .Lskip_interrupt_disable
 	csrsi              mstatus, 0x8
 .Lskip_interrupt_disable:
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -450,9 +450,7 @@ exception_entry_asm:
 	// slightly preferable to clearing mtdc, which would also ensure that we
 	// looped, because the architectural loop is tighter and involves no
 	// program text, making it easier for microarchitecture to detect.
-	bnez               sp, .Lexception_entry_still_alive
-	cspecialw          mtcc, csp
-.Lexception_entry_still_alive:
+	beqz               sp, .Lexception_reentered
 
 	// csp now points to the save reg frame that we can use.
 	// The guest csp (c2) is now in mtdc. Will be spilled later, but we
@@ -874,6 +872,11 @@ exception_entry_asm:
 	csc                ca1, TrustedStack_offset_ca1(csp)
 	j                  .Linstall_context
 
+.Lexception_reentered:
+	cmove              csp, cnull
+	cspecialw          mtcc, csp
+	clc                csp, 0(csp)
+	j                  .Lexception_reentered
 
 .size exception_entry_asm, . - exception_entry_asm
 
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -541,10 +541,10 @@ exception_entry_asm:
 	// in that posture until the mret in install_context.
 
 	// Switch onto the new thread's trusted stack
-	LoadCapPCC         ct0, compartment_switcher_sealing_key
+	LoadCapPCC         csp, compartment_switcher_sealing_key
 	li                 gp, SEAL_TYPE_SealedTrustedStacks
-	csetaddr           ct0, ct0, gp
-	cunseal            csp, ca0, ct0
+	csetaddr           csp, csp, gp
+	cunseal            csp, ca0, csp
 	clw                t0, TrustedStack_offset_mcause(csp)
 
 	// Only now that we have done something that actually requires the tag of
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -414,11 +414,10 @@ switcher_after_compartment_call:
 	clc                cra, SPILL_SLOT_pcc(csp)
 	clc                cgp, SPILL_SLOT_cgp(csp)
 	cincoffset         csp, csp, SPILL_SLOT_SIZE
-	// Set the return registers
+	// Set the first return register (a0) and zero the other (a1) below
 	li                 a0, -ENOTENOUGHTRUSTEDSTACK
-	li                 a1, 0
 	// Zero everything else
-	zeroAllRegistersExcept ra, sp, gp, s0, s1, a0, a1
+	zeroAllRegistersExcept ra, sp, gp, s0, s1, a0
 	cret
 
 .size compartment_switcher_entry, . - compartment_switcher_entry
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -340,8 +340,8 @@ __Z26compartment_switcher_entryz:
 	zeroRegisters      tp, t1, t2, s0, s1
 	cjalr              cra
 
-	.globl switcher_skip_compartment_call
-switcher_skip_compartment_call:
+	.globl switcher_after_compartment_call
+switcher_after_compartment_call:
 	// If we are doing a forced unwind of the trusted stack then we do almost
 	// exactly the same as a normal unwind.  We will jump here from the
 	// exception path (.Lforce_unwind)
@@ -419,7 +419,7 @@ switcher_skip_compartment_call:
 .Lstack_too_small:
 	li                 a0, -ENOTENOUGHSTACK
 	li                 a1, 0
-	j                  switcher_skip_compartment_call
+	j                  switcher_after_compartment_call
 .size compartment_switcher_entry, . - compartment_switcher_entry
 
 	// the entry point of all exceptions and interrupts
@@ -588,7 +588,7 @@ exception_entry_asm:
 .Lforce_unwind:
 	li                 a0, -ECOMPARTMENTFAIL
 	li                 a1, 0
-	j                  switcher_skip_compartment_call
+	j                  switcher_after_compartment_call
 
 
 // If we have run out of trusted stack, then just restore the caller's state
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -205,10 +205,9 @@ __Z26compartment_switcher_entryz:
 
 	// make sure the trusted stack is still in bounds
 	clhu               tp, TrustedStack_offset_frameoffset(ct2)
-	cgetlen            t2, ct2
-	bgeu               tp, t2, .Lout_of_trusted_stack
-	// we are past the stacks checks. Reload ct2; tp is still as it was
-	cspecialr          ct2, mtdc
+	cgetlen            s0, ct2
+	bgeu               tp, s0, .Lout_of_trusted_stack
+	// we are past the stacks checks.
 	// ctp points to the current available trusted stack frame.
 	cincoffset         ctp, ct2, tp
 	csc                csp, TrustedStackFrame_offset_csp(ctp)
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -412,6 +412,23 @@ switcher_after_compartment_call:
 	li                 a0, -ENOTENOUGHSTACK
 	li                 a1, 0
 	j                  switcher_after_compartment_call
+
+// If we have run out of trusted stack, then just restore the caller's state
+// and return an error value.
+.Lout_of_trusted_stack:
+	// Restore the spilled values
+	clc                cs0, SPILL_SLOT_cs0(csp)
+	clc                cs1, SPILL_SLOT_cs1(csp)
+	clc                cra, SPILL_SLOT_pcc(csp)
+	clc                cgp, SPILL_SLOT_cgp(csp)
+	cincoffset         csp, csp, SPILL_SLOT_SIZE
+	// Set the return registers
+	li                 a0, -ENOTENOUGHTRUSTEDSTACK
+	li                 a1, 0
+	// Zero everything else
+	zeroAllRegistersExcept ra, sp, gp, s0, s1, a0, a1
+	cret
+
 .size compartment_switcher_entry, . - compartment_switcher_entry
 
 	// the entry point of all exceptions and interrupts
@@ -581,23 +598,6 @@ exception_entry_asm:
 	li                 a0, -ECOMPARTMENTFAIL
 	li                 a1, 0
 	j                  switcher_after_compartment_call
-
-
-// If we have run out of trusted stack, then just restore the caller's state
-// and return an error value.
-.Lout_of_trusted_stack:
-	// Restore the spilled values
-	clc                cs0, SPILL_SLOT_cs0(csp)
-	clc                cs1, SPILL_SLOT_cs1(csp)
-	clc                cra, SPILL_SLOT_pcc(csp)
-	clc                cgp, SPILL_SLOT_cgp(csp)
-	cincoffset         csp, csp, SPILL_SLOT_SIZE
-	// Set the return registers
-	li                 a0, -ENOTENOUGHTRUSTEDSTACK
-	li                 a1, 0
-	// Zero everything else
-	zeroAllRegistersExcept ra, sp, gp, s0, s1, a0, a1
-	cret
 
 // If we have a possibly recoverable error, see if we have a useful error
 // handler.  At this point, the register state will have been saved in the
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -45,6 +45,49 @@
  * purpose, &c), one should not read too much of the psABI calling convention
  * into the code here.  Within the switcher, the machine is a raw register
  * machine and C is a distant, high-level language.
+ *
+ * Since this is the part of the map labeled "here be dragons", we have added
+ * "Register Atlas" comments throughout.  Lines within an atlas consist of a
+ * comma-whitespace-separated list of registers, a colon, and descriptive text.
+ * In general, atlases should cover all (including dead) registers.  Point
+ * changes to the atlas are denoted with "Atlas update", to emphasize that
+ * registers not named are not dead but instead retain their meaning from the
+ * last full atlas.
+ *
+ * Labels associated with interesting control flow are annotated with
+ *
+ *  - "FROM:", which may be repeated once for each predecessor label or these:
+ *    - "above": the immediately prior block
+ *    - "cross-call": untrusted code making a cross-compartment call
+ *    - "malice": untrusted code outside the switcher
+ *    - "interrupt": an asynchronous external event
+ *    - "error": a trap from within the switcher
+ *
+ *  - "IFROM:", which indicates an indirect transfer of control (through cjalr
+ *    or mepcc/mret, for example).
+ *
+ *  - "ITO:", the other direction of "IFROM:"
+ *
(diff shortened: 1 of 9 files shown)


sdk/core/switcher/entry.S
```diff
@@ -183,159 +188,266 @@ switcher_scheduler_entry_csp:
 	.p2align 2
 	.type __Z26compartment_switcher_entryz,@function
 __Z26compartment_switcher_entryz:
+	/*
+	 * Atlas:
+	 *  mtdc: thread trusted stack pointer
+	 *  ra: caller return address (nominally; corruption is on them)
```
rmn30, line 194: > (nominally; corruption is on them)

I think that this is entered via an interrupt disabling sentry so we can be sure that `ra` is a return sentry to the caller. If we ever change that to interrupt inheriting we'd lose that guarantee but I agree it doesn't matter much.

nwf: That's a good point; will update.

sdk/core/switcher/entry.S
```diff
@@ -183,159 +188,266 @@ switcher_scheduler_entry_csp:
...
+	 *  gp: caller state, to be spilled, value unused in switcher
+	 *  s0, s1: caller state, to be spilled, value unused in switcher
+	 *  t1: import table entry for the target callee
+	 *      (see LLVM's RISCVExpandPseudo::expandCompartmentCall)
+	 *  a0, a1, a2, a3, a4, a5, t0: caller arguments to callee, passed or zered
+	 *  tp, t2: scratch
+	 *
+	 * Traps on an invalid csp or an overflow (even mid-way through this spill
+	 * sequence) will trigger invocation of the compartment's error handler, or
+	 * forcibly unwind if none. Nothing secret from the caller would be visible.
```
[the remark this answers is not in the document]
nwf: Oh, hah, I hadn't thought about that read.  I'd meant "nothing [that is held to be] secret from the caller", but I agree that "nothing secret [originating] from the caller" is another way to read that sentence.  Will rework.

nwf:
OK, I believe this is ready for review.  As so often happens with my PRs, this is somewhat separable, with many of the "bottom" commits suitable for landing even if the whole thing is not, if that'd be desirable.  In the same spirit, this is almost surely best reviewed commit by commit, rather than all at once.

sdk/core/switcher/entry.S
```diff
@@ -444,57 +722,106 @@ exception_entry_asm:
...
+	 */
+	li                 t1, MCAUSE_THREAD_INTERRUPT
+	// LIVE OUT: mtdc, sp
 	beq                t0, t1, .Lhandle_injected_error
 
-	// Environment call from M-mode is exception code 11.
-	// We need to skip the ecall instruction to avoid an infinite loop.
+	/*
+	 * Environment call from M-mode is exception code 11.
+	 * We need to skip the ecall instruction to avoid an infinite loop.
```
davidchisnall, line 800: We could also remove mcall (or treat it as a local error that goes to the error handler) and expose a yield entry point, if that would simplify the code.

sdk/core/switcher/entry.S
```diff
@@ -504,207 +831,348 @@ exception_entry_asm:
...
+	 * Some switcher instructions' traps are handled specially, by looking at
+	 * the offset of mepcc.  Otherwise, we're off to a force unwind.
+	 */
+.Lhandle_error_in_switcher:
+	auipcc             ctp, %cheriot_compartment_hi(.Lswitcher_entry_first_spill)
+	cincoffset         ctp, ctp, %cheriot_compartment_lo_i(.Lhandle_error_in_switcher)
+	// LIVE OUT: mtdc
+	bne                t1, tp, .Lforce_unwind
+	li                 a0, -ENOTENOUGHSTACK
+	li                 a1, 0
```
davidchisnall, line 931: We never fall through from here and `.Linstall_return_context` has no predecessors, so I think this can be simplified by:

 - Move `.Lhandle_error_in_switcher` above `.Linstall_return_context`
 - Remove the `j .Lhandle_error_in_switcher`.
 - Invert the condition on 918 and have the normal path be the fallthrough one.

sdk/core/switcher/entry.S
```diff
@@ -504,207 +831,348 @@ exception_entry_asm:
...
+	// LIVE OUT: sp, a0, a1
+	j                  .Linstall_return_context
 
+.Lhandle_error_not_switcher:
```
davidchisnall, line 935: This lacks an atlas / live ins, but the label goes away if you make the change above.

sdk/core/switcher/entry.S
```diff
@@ -504,207 +831,348 @@ exception_entry_asm:
...
 	// Load the interrupted thread's stack pointer into ct0
 	clc                ct0, TrustedStack_offset_csp(csp)
-	// See if we can find a handler:
+	// Atlas: t0: interrupted thread's stack pointer
+
+	/*
+	 * If we have already unwound so far that the TrustedStack::frameoffset is
+	 * pointing at TrustedStack::frames[0] -- that is, if the stack has no
+	 * active frames on it -- then just go back to the context we came from,
+	 * effectively parking this thread in a (slow) infinite loop.
```
davidchisnall, line 944: Don't we want to hit the thread exit path in that case?

sdk/core/switcher/entry.S
```diff
@@ -504,207 +831,348 @@ exception_entry_asm:
...
+	cgetaddr           a1, ct0
+	cgetbase           a2, ct0
 	sub                a1, a1, a2
-	csetaddr           ca0, ca0, a2
-	// The code that installs the context expects csp to be in ct0
-	csetboundsexact    ct0, ca0, a1
+	csetaddr           ct0, ct0, a2
+	// The code that installs the context expects the target stack to be in ct0
+	csetboundsexact    ct0, ct0, a1
 .Lrecovered_stack:
```
davidchisnall, line 1065: Missing atlas, live-ins, predecessors.

sdk/core/switcher/entry.S
```diff
@@ -504,207 +831,348 @@ exception_entry_asm:
...
-	csc                cs1, 0(ct0)
-	// Source for context copy.
-	cincoffset         ca2, csp, TrustedStack_offset_cra
-	// Destination for context copy
-	cincoffset         ca3, ct0, TrustedStack_offset_cra
-	copyContext        ca3, ca2, cs1, a4
+	csc                cs1, TrustedStack_offset_mepcc(ct0)
+	// Now copy the 15 GPRs from the trusted stack (sp)
+	cincoffset         ca2, csp, TrustedStack_offset_cra // source
+	cincoffset         ca3, ct0, TrustedStack_offset_cra // destination
```
davidchisnall, line 1174: Why did you change the comment style here?  We don't put comments on the end of lines anywhere else in the switcher, do we?

sdk/core/switcher/entry.S
```diff
@@ -713,77 +1181,147 @@ exception_entry_asm:
...
-
-	// We have been asked to install the new register context and resume.
-	// We do this by copying the register frame over the save area and entering
-	// the exception resume path.  This may fault, but if it does then we will
-	// detect it as a double fault and forcibly unwind.
-
-	// Load the trusted stack pointer to ct1
 	cspecialr          ct1, mtdc
+	// Atlas: t1: pointer to TrustedStack
 #ifdef CONFIG_MSHWM
```
davidchisnall, line 1254: Maybe not this PR, but I think we can assume MSHWM now.  We haven't tested without it for ages and I expect everyone to use it.

sdk/core/switcher/entry.S
```diff
@@ -183,222 +190,479 @@ switcher_scheduler_entry_csp:
...
+	 *  tp: scalar offset of the next available TrustedStack::frames[]
+	 */
+	// LIVE OUT: mtdc, sp
 	bgeu               tp, t2, .Lout_of_trusted_stack
-	// we are past the stacks checks. Reload ct2; tp is still as it was
+	// we are past the stacks checks.
 	cspecialr          ct2, mtdc
-	// ctp points to the current available trusted stack frame.
+	// Atlas: t2: pointer to this thread's TrustedStack (again)
+	// The register file is (again) unsafe to expose to the caller
```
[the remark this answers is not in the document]
nwf: I'd thought clobbering `t2` was deliberate the first time I read this.  Reworked in light of the above (use `s0` instead).

sdk/core/switcher/entry.S
```diff
@@ -183,222 +190,479 @@ switcher_scheduler_entry_csp:
...
-	zero_stack         t2, s0, gp
+	zero_stack         /* base = */ t2, /* top = */ s0, /* scratch = */ gp
 .Lafter_zero:
+	/*
+	 * LIVE IN: mtdc, sp, tp, t0, t1, a0, a1, a2, a3, a4, a5
+	 *
+	 * Atlas:
+	 *  t2, gp: caller stack capabilities (dead)
+	 *  s0: scratch scalar (dead)
+	 */
```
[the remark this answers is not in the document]
nwf: It's a fairly local branch target, and we just gave one, so I wasn't planning on a full one again, but can if you'd rather.

hlef:
Just mentioning here: I encountered the bug where a compartment's error handler isn't called if the thread is at the root of the trusted stack while testing my future PR that uses the new error handlers in the NW stack (https://github.com/CHERIoT-Platform/network-stack/tree/hlefeuvre/stack-overflow-resilient-error-handler). I applied this patch to my local tree and it fixed it.

sdk/core/switcher/entry.S
```diff
@@ -183,222 +190,479 @@ switcher_scheduler_entry_csp:
...
+	/*
+	 * FROM: __Z26compartment_switcher_entryz
+	 * LIVE IN: mtdc, sp
+	 *
+	 * Atlas:
+	 *  mtdc: TrustedStack pointer
+	 *  sp: Caller stack pointer, pointing at switcher spill frame
+	 */
+	// Restore the spilled values
+	clc                cs0, SPILL_SLOT_cs0(csp)
```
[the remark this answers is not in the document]
nwf: We would not have made it this far.  Commentary updated.

sdk/core/switcher/entry.S
```diff
@@ -183,222 +190,479 @@ switcher_scheduler_entry_csp:
...
+	 * Make sure the caller's CSP has the expected permissions and that its
+	 * top and base are 16-byte aligned.  We have already checked that it is
+	 * tagged and unsealed and 8-byte aligned by virtue of surviving the
+	 * stores above.
+	 *
+	 * Uses tp and t2 as scratch scalars.
+	 */
+	cgetperm           t2, csp
+	li                 tp, COMPARTMENT_STACK_PERMISSIONS
+	bne                tp, t2, .Lforce_unwind
```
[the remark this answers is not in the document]
nwf: Please check that you are happy with the expanded comment.

sdk/core/switcher/entry.S
```diff
@@ -183,222 +190,479 @@ switcher_scheduler_entry_csp:
...
+	 *  tp, s0, s1, t0, t1, t2, a2, a3, a4, a5: dead or callee state (to be 0ed)
+	 */
+
+	/*
+	 * The return sentry given to the callee as part of that cjalr could be
+	 * captured by the callee or passed back to the caller.  We cannot assume
+	 * well-bracketed control flow.  However, the requirements of the next block
+	 * of code are minimal: mtdc must be a TrustedStack pointer.  The contents
+	 * of a0 and a1 will be exposed to the compartment above the one currently
+	 * executing, or the thread will be terminated if there is no such.
```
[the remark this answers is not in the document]
nwf: That's fair; I've reworded the comment.  Please advise.

sdk/core/switcher/entry.S
nwf: I think we already *have* hit the thread exit path in this case, and the scheduler is continuing to try to run us.  But all the same, see https://github.com/CHERIoT-Platform/cheriot-rtos/pull/328 (not presently part of this PR stack, but if it lands, I'll rebase).

sdk/core/switcher/entry.S
```diff
@@ -183,222 +190,479 @@ switcher_scheduler_entry_csp:
...
-	// trap loop: the reentry will probably have put something non-zero into
-	// mtdc, so we wouldn't hit this, and wouldn't loop, when we take that
-	// third trap.  (Exactly what we'd do instead is hard to say; we'd try
-	// spilling registers to an attacker-controlled pointer, at the very
-	// least.) Therefore, clobber mtcc (!) to ensure that the certainly
-	// upcoming third trap puts us in an architectural trap loop.  This is
-	// slightly preferable to clearing mtdc, which would also ensure that we
-	// looped, because the architectural loop is tighter and involves no
-	// program text, making it easier for microarchitecture to detect.
+	/*
```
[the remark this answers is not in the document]
nwf: At a glance, the dump file shows the correct line numbers for the few instructions after `exception_entry_asm`, and there have been loads of block comments by that point of the file.  Is that what you meant for me to check, or something else?

sdk/core/switcher/entry.S
nwf: Not this PR, please.  But maybe.

sdk/core/switcher/entry.S
```diff
@@ -183,222 +190,479 @@ switcher_scheduler_entry_csp:
...
+	 *
+	 * Atlas:
+	 *  mtdc: either pointer to TrustedStack or zero
+	 */
+	/*
+	 * We do not trust the interruptee's context. We cannot use its stack in any
+	 * way.  The save reg frame we can use is fetched from the tStack.
+	 * In general, mtdc holds the trusted stack register.  We are here with
+	 * interrupts off and precious few registers available to us, so swap it
+	 * with the csp (we'll put it back, later).
```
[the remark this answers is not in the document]
nwf: It's one of {the TrustedStack pointer, null, the captured csp}.  Prose or annotation suggestions more than welcome.

I am tempted to take a stab at moving the latter out to mscratch and see what happens; I'm a little worried that it'd make the state machine nicer at the expense of more instructions, and more CSR-accessing instructions at that.

sdk/core/switcher/entry.S
```diff
@@ -183,222 +190,479 @@ switcher_scheduler_entry_csp:
...
+	 *  gp: caller state, to be spilled, value unused in switcher
+	 *  s0, s1: caller state, to be spilled, value unused in switcher
+	 *  t0: possible caller argument to callee, passed or zered
+	 *  t1: sealed export table entry for the target callee
+	 *      (see LLVM's RISCVExpandPseudo::expandCompartmentCall)
+	 *  a0, a1, a2, a3, a4, a5: possible caller arguments to callee, passed/0ed
+	 *  tp, t2: scratch
+	 */
+	/*
+	 * The caller should back up all caller saved registers.  Spill
```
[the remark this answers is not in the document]
nwf: Please check that you're happy with the new comment here.

sdk/core/switcher/entry.S
```diff
@@ -183,222 +190,479 @@ switcher_scheduler_entry_csp:
...
 	csrw               CSR_MSHWM, sp
 #endif
-#endif // CONFIG_NO_SWITCHER_SAFETY
-.Lout:
-	// Fetch the sealing key
+
+	// Fetch the sealing key, using gp as a scratch scalar
 	LoadCapPCC         cs0, compartment_switcher_sealing_key
-	li                 gp, 9
+	// Atlas: s0: switcher sealing key
```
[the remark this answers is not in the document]
nwf: Took a stab, but likely missed some.

sdk/core/switcher/entry.S
rmn30: The ifdef was useful for benchmarking, not sure if you still want that?

sdk/core/switcher/entry.S
vmurali, line 1: My first set of comments till ".Lswitch_after_zero"

sdk/core/switcher/entry.S
```diff
@@ -119,33 +159,12 @@ switcher_scheduler_entry_csp:
...
-	bnez               t2, .Lforce_unwind
-.endm
-
 /**
  * Zero the stack.  The three operands are the base address, the top address,
  * and a scratch register to use.  The base must be a capability but it must
  * be provided without the c prefix because it is used as both a capability
- * and integer register.  All three registers are clobbered.
+ * and integer register.  All three registers are clobbered, but base and top
+ * are still capabilities afterwards.
```
vmurali, line 167: How is top still a cap, addi instruction is used. c\base is still a cap?

sdk/core/switcher/entry.S
```diff
@@ -184,99 +203,275 @@ switcher_scheduler_entry_csp:
 	.type __Z26compartment_switcher_entryz,@function
 __Z26compartment_switcher_entryz:
 	/*
-	 * Spill caller-save registers carefully.  If we find ourselves unable to do
-	 * so, we'll return an error to the caller (via the exception path; see
-	 * .Lhandle_error_in_switcher).  The error handling path assumes that the
-	 * first spill is to the lowest address and guaranteed to trap if any would.
+	 * FROM: cross-call
+	 * FROM: malice
```
vmurali, line 207: How can this be invoked except through a cross-compartment call?

sdk/core/switcher/entry.S
```diff
@@ -184,99 +203,275 @@ switcher_scheduler_entry_csp:
 	.type __Z26compartment_switcher_entryz,@function
 __Z26compartment_switcher_entryz:
 	/*
-	 * Spill caller-save registers carefully.  If we find ourselves unable to do
-	 * so, we'll return an error to the caller (via the exception path; see
-	 * .Lhandle_error_in_switcher).  The error handling path assumes that the
-	 * first spill is to the lowest address and guaranteed to trap if any would.
+	 * FROM: cross-call
```
vmurali, line 206: Line 190: What are hazard pointers? (Can't comment on a non changed line on git)

sdk/core/switcher/entry.S
```diff
@@ -184,99 +203,275 @@ switcher_scheduler_entry_csp:
...
+	 * FROM: malice
+	 * IRQ: deferred
+	 * LIVE IN: mtdc, ra, sp, gp, s0, s1, t0, t1, a0, a1, a2, a3, a4, a5
+	 *
```
vmurali, line 210: Good so state all registers except tp, t2 (which are both dead as explained below) and zero (which is unnecessary)

sdk/core/switcher/entry.S
```diff
@@ -184,99 +203,275 @@ switcher_scheduler_entry_csp:
...
+	 * Atlas:
+	 *  mtdc: pointer to this thread's TrustedStack
+	 *        (may be 0 from buggy/malicious scheduler thread)
+	 *  ra: caller return address
+	 *      (at the moment, this is ensured because we enter via an
+	 *      IRQ-disabling forward sentry, which requires ra as the destination
+	 *      register of the cjalr the caller used, but we are not relying on
+	 *      this property, and we hope to relax the switcher's IRQ posture)
```
vmurali, line 218: For tail call optimization?

sdk/core/switcher/entry.S
```diff
@@ -184,99 +203,275 @@ switcher_scheduler_entry_csp:
...
+	 *  sp: nominally, caller's stack pointer; will check integrity below
+	 *  gp: caller state, to be spilled, value unused in switcher
+	 *  s0, s1: caller state, to be spilled, value unused in switcher
+	 *  t0: possible caller argument to callee, passed or zered in switcher
+	 *      (specifically, this is the pointer to arguments beyond a0-a5 and/or
+	 *      variadic arguments)
+	 *  t1: sealed export table entry for the target callee
+	 *      (see LLVM's RISCVExpandPseudo::expandCompartmentCall)
```
vmurali, line 226: Maybe link to the documentation in the ISA doc also (Section 5.4)?

sdk/core/switcher/entry.S
```diff
@@ -184,99 +203,275 @@ switcher_scheduler_entry_csp:
...
+	 *  tp, t2: dead
+	 */
+	/*
+	 * By virtue of making a call, the caller is indicating that all caller-save
+	 * registers are dead.  Because we are crossing a trust boundary, the
+	 * switcher must spill callee-save registers.  If we find ourselves unable
+	 * to do so for "plausibly accidental" reasons, we'll return an error to the
+	 * caller (via the exception path; see .Lhandle_error_in_switcher).
+	 * Specifically, the first spill here is to the lowest address and so
+	 * guaranteed to raise a bounds fault if any of the stores here would.
```
vmurali, line 238: Not sure what you mean here. Yes, if the lowest address is less than the base, this is guaranteed. But there could be other bounds check violations (higher than top, etc), no?

sdk/core/switcher/entry.S
```diff
@@ -184,99 +203,275 @@ switcher_scheduler_entry_csp:
...
+
+	/*
+	 * Before we access any privileged state, we can verify the compartment's
+	 * csp is valid. If not, force unwind.  Note that these checks are purely to
+	 * protect the callee, not the switcher itself, which can always bail and
+	 * forcibly unwind the caller.
+	 *
+	 * Make sure the caller's CSP has the expected permissions (including that
+	 * it is a stack pointer, by virtue of being local and bearing SL) and that
+	 * its top and base are 16-byte aligned.  We have already checked that it is
```
vmurali, line 280: Stack sizes are guaranteed to be between the 2^12 and 2^13 range (i.e. between 4089 and 8176) ?

sdk/core/switcher/entry.S
```diff
@@ -184,99 +203,275 @@ switcher_scheduler_entry_csp:
...
-	// we are past the stacks checks. Reload ct2; tp is still as it was
-	cspecialr          ct2, mtdc
-	// ctp points to the current available trusted stack frame.
+	cgetlen            s0, ct2
+	/*
+	 * Atlas update:
+	 *  s0: scalar length of the TrustedStack structure
+	 *  tp: scalar offset of the next available TrustedStack::frames[] element
+	 */
+	// LIVE OUT: mtdc, sp
```
vmurali, line 330: The LIVE OUT doesn't capture, for instance ct2 which is still alive and used. Not sure if LIVE OUT is helping. Just having alive, dead-again and dead-still list would suffice.

sdk/core/switcher/entry.S
```diff
@@ -184,99 +203,275 @@ switcher_scheduler_entry_csp:
...
+	 *  - s1 for the length of the stack suffix to which the callee is entitled
+	 */
+//.Lswitch_stack_chop:
 	cgetaddr           s0, csp
 	cgetbase           s1, csp
 	csetaddr           csp, csp, s1
 	sub                s1, s0, s1
 	csetboundsexact    ct2, csp, s1
 	csetaddr           csp, ct2, s0
+	/*
```
vmurali, line 380: Simpler to state as follows:

// s0 = old csp's address
// csp = {base = old csp's base, length = offset of old csp's top from base, address = old csp's address}

sdk/core/switcher/entry.S
```diff
@@ -184,99 +203,275 @@ switcher_scheduler_entry_csp:
...
+	 * Atlas:
+	 *  s0: address of stack boundary between caller and callee frames
```
vmurali, line 382: What does "stack boundary between ..." mean?

sdk/core/switcher/entry.S
```diff
@@ -184,99 +203,275 @@ switcher_scheduler_entry_csp:
...
+	 *  sp: pointer to stack, with bounds as t2, cursor at boundary in s0
```
vmurali, line 383: new t2 is "dead-again", so referring to it is not very helpful. Better to use the old values instead of a "dead-again" value.

sdk/core/switcher/entry.S
```diff
@@ -184,99 +203,275 @@ switcher_scheduler_entry_csp:
...
-	// Use stack high water mark as base address for zeroing.  If this faults
-	// then it will trigger a force unwind.  This can happen only if the caller
-	// is doing something bad.
+//.Lswitch_shwm_skip_zero:
+	bge                gp, sp, .Lswitch_after_zero
+	/*
+	 * Use stack high water mark as base address for zeroing.  If this faults
+	 * then it will trigger a force unwind.  This can happen only if the caller
+	 * is doing something bad.
+	 */
```
vmurali, line 400: What happens if the high water mark is below the base of the stack?

sdk/core/switcher/entry.S
```diff
@@ -184,99 +203,275 @@ switcher_scheduler_entry_csp:
...
 	csetaddr           ct2, csp, gp
 #endif
-	zero_stack         t2, s0, gp
-.Lafter_zero:
+	zero_stack         /* base = */ t2, /* top = */ s0, /* scratch = */ gp
```
vmurali, line 403: Finished reviewing till this line.

vmurali (commented):
First of many sets of comments.

sdk/core/switcher/entry.S
nwf: Oh, yes, sorry, I'd meant "here would access below the base".  I'll adjust the comment.

sdk/core/switcher/entry.S
nwf: Hm, I didn't think so.  Can you expand your question?

sdk/core/switcher/entry.S
nwf: Hm.  I'd meant LIVE OUT to indicate live on the branch taken condition, which was my (possibly erroneous) interpretation of David's suggestion.  If something else is more useful, I'm happy to scrap it for that!

sdk/core/switcher/entry.S
nwf: Better?

sdk/core/switcher/entry.S
nwf: I have attempted clarification.  Better?

sdk/core/switcher/entry.S
nwf: Then it's off to force unwinding we go... until this thread exits, I think, because we'll try to zero the stack again (in `switcher_skip_compartment_call`, via an exception, then `exception_entry_asm`, `.Lexception_might_handle`, `.Lhandle_error`, `.Lhandle_error_switcher_pcc`, and finally `.Lcommon_force_unwind`, phew!).  That _shouldn't_ happen, but...

nwf (commented):
Thanks @vmurali!  971d8000 includes changes from your suggestions and I've left questions where I didn't understand (sorry!).

sdk/core/switcher/entry.S
```diff
@@ -184,99 +206,285 @@ switcher_scheduler_entry_csp:
...
 	sub                s1, s0, s1
 	csetboundsexact    ct2, csp, s1
 	csetaddr           csp, ct2, s0
+	/*
+	 * Atlas:
+	 *  s0: address of stack boundary between caller and callee frames, that is,
+	 *      the lowest address of the register spill from
+	 *      .Lswitch_entry_first_spill)
+	 *  sp: pointer to stack, with its limit and address set to the address in
+	 *      s0.  The base and permissions have not been altered from sp at
```
vmurali, line 394: Easier to specify what the length is

sdk/core/switcher/entry.S
vmurali: I haven't reached those portions of the code. Would it be easier to fail now, with a check to ensure it's bigger than the base (and fail immediately otherwise)?

Alternatively, is there an invariant that CSR_MSHWM is >= csp.base for any tagged csp? If so, that invariant can be stated, and we can ensure that it's maintained in the code. I suppose this is somewhat trivially true, because CSR_MSHWM is probably set to be the starting csp base for each thread and it can only increase (all CSR_MSHWM writes are protected to ensure that the csp is tagged when setting the high water mark)

vmurali:
> Hm? If you're asking about "and we hope to relax", we'd like to be preemptive during stack zeroing, for example, as it currently impinges upon real-time guarantees.

I think if you force push it, somehow the comments go away.

I believe you are talking about my comment "tail call optimization?". Yeah, I didn't understand what relaxation you are hoping to do later. I still don't understand why preemptive stack zeroing is related to ra as return register for cjalr.

vmurali:
> Hm, I didn't think so. Can you expand your question?

I think this is about stack length. I think 16 byte alignment means you cannot have stack sizes less than 4089 (or something close to 4096). That's the assumption?

vmurali:
> Hm. I'd meant LIVE OUT to indicate live on the branch taken condition, which was my (possibly erroneous) interpretation of David's suggestion. If something else is more useful, I'm happy to scrap it for that!

Just live variables overall is easier. I mean, I am going by traditional hoare tuple: weakest precondition and strongest post condition. In this particular example, the block below needs the t2, while the branch taken probably does not. I will worry about what the branch taken needs at its code rather than here.

nwf:
> > Hm? If you're asking about "and we hope to relax", we'd like to be preemptive during stack zeroing, for example, as it currently impinges upon real-time guarantees.
> 
> I think if you force push it, somehow the comments go away.
> 
> I believe you are talking about my comment "tail call optimization?". Yeah, I didn't understand what relaxation you are hoping to do later. I still don't understand why preemptive stack zeroing is related to ra as return register for cjalr.

Ah, sorry!  There's a lot there to unpack.

The first thing to note is that on entry to `__Z26compartment_switcher_entryz`, `ra` is, in principle, pretty arbitrary; the caller could tell us to return anywhere and we -- the switcher -- would be fine with that.  All we do with their `ra` is spill it after `.Lswitch_entry_first_spill` and then unspill it in `switcher_skip_compartment_call` for use by `cret`.

It happens, however, that the whole of `__Z26compartment_switcher_entryz` right now runs with interrupts deferred, enforced by the sole way to enter that function being via an IRQ-disabling forward sentry.  As a side effect of https://github.com/CHERIoT-Platform/cheriot-sail/pull/54, the only `cjalr` instructions that can successfully invoke such a sentry necessarily use `ra` as their destination register.  Thus we, temporarily, have a stronger guarantee than we need in the cross-call switcher.

They're coupled only in that when we weaken our interrupt disposition to allow the switcher to be preempted, the uselessly strong guarantee about the contents of `ra` will go away.

vmurali:
> They're coupled only in that when we weaken our interrupt disposition to allow the switcher to be preempted, the uselessly strong guarantee about the contents of `ra` will go away.

In the relaxed scenario, you would call the switcher with an interrupt inheriting sentry, and hence can set up any register as the returning register?

sdk/core/switcher/entry.S
```diff
@@ -184,99 +206,285 @@ switcher_scheduler_entry_csp:
...
+	 *  - s0 for the current untrusted stack base address (the lowest address of
+	 *    the register spill we created at .Lswitch_entry_first_spill)
+	 *  - s1 for the length of the stack suffix to which the callee is entitled
+	 */
+//.Lswitch_stack_chop:
 	cgetaddr           s0, csp
 	cgetbase           s1, csp
 	csetaddr           csp, csp, s1
 	sub                s1, s0, s1
 	csetboundsexact    ct2, csp, s1
```
vmurali, line 386: Hmm, what's the guarantee that this wouldn't fail?

Is there a better way to carve things out so that we dont fail here instead of trying to use csetboundsexact and failing? Perhaps get the cap with the same base but smaller length?

sdk/core/switcher/entry.S
```diff
@@ -27,6 +27,21 @@
 #define SPILL_SLOT_pcc 24
 #define SPILL_SLOT_SIZE 32
 
+/*
+ * RISC-V custom-use mcause values for RTOS internal use.
+ * See also include/priv/riscv.h, which defines the same names and values in C
```
rmn30, line 32: maybe we should have a c header file with the macro definitions and include it in both places?

davidchisnall:
> In the relaxed scenario, you would call the switcher with an interrupt inheriting sentry, and hence can set up any register as the returning register?

Yes. I’m not *too* worried about that because we jump to the value from ra only after restoring state. With the current code structure, it would be possible to use the switcher to disable interrupts in an arbitrary piece of code if we exposed an interrupt-inheriting sentry if you:

1. Derive a capability from your PCC and set the address wherever you want.
2. Put that value in cra.
3. Invoke the switcher to jump to an interrupts-disabled function.
4. The switcher resumes still interrupt-disabled.
5. Switcher jumps to cra.

We could avoid this fairly trivially by just force unwinding if cra does not contain a return sentry.

sdk/core/switcher/entry.S
nwf: I think, once we have it, we should use https://github.com/CHERIoT-Platform/cheriot-sail/pull/74; I've left myself a note over on the emerging to-do list at https://github.com/CHERIoT-Platform/cheriot-rtos/issues/334 .

sdk/core/switcher/entry.S
nwf: That's a great idea, thanks.

sdk/core/switcher/entry.S
```diff
@@ -298,98 +512,190 @@ __Z26compartment_switcher_entryz:
...
-	// Enable interrupts of the interrupt-disable bit is not set in flags
-	bnez               t1, .Lskip_interrupt_disable
+
+	/*
+	 * Enable interrupts if the interrupt-disable bit is not set in flags.  See
+	 * loader/types.h's InterruptStatus and ExportEntry::InterruptStatusMask
+	 */
+	andi               t1, tp, 0x10
+	bnez               t1, .Lswitch_skip_interrupt_disable
 	csrsi              mstatus, 0x8
```
rmn30, line 563: This instruction actually enable interrupts, right? In that case the label should be called `skip_interrupts_enable`.

sdk/core/switcher/entry.S
```diff
@@ -298,98 +512,190 @@ __Z26compartment_switcher_entryz:
...
+	 * program counter and the program counter (of the instruction itself) into
+	 * ra (while sealing it to be a backwards-arc sentry).  That is, the value
+	 * we have so carefully been keeping in ra is clobbered, but only after it
+	 * becomes the next program counter.
+	 */
+	// LIVE OUT: *
 	cjalr              cra
 
 	.globl switcher_skip_compartment_call
 switcher_skip_compartment_call:
```
rmn30, line 591: Maybe a more informative name for this label would be `switcher_after_compartment_call`?

sdk/core/switcher/entry.S
```diff
@@ -183,222 +190,479 @@ switcher_scheduler_entry_csp:
...
+	/*
+	 * FROM: malice, above, .Lstack_too_small
+	 * LIVE IN: mtdc, a0, a1
+	 *
+	 * Atlas:
+	 *  mtdc: pointer to this thread's TrustedStack
+	 *  a0, a1: return value(s).  The callee function must ensure that it clears
+	 *          these as appropriate if it is returning 0 or 1 values and not 2.
+	 *  ra, sp, gp: dead or callee state (to be replaced by caller)
+	 *  tp, s0, s1, t0, t1, t2, a2, a3, a4, a5: dead or callee state (to be 0ed)
```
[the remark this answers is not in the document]
nwf: I originally tried to do something like that, but I agree that the rule should be "don't".

nwf:
This is sitting atop #346 at the moment (which hopefully lands quickly).  The `dot_from_switcher.lua` script has grown the ability to cross-check not just `FROM`/`IFROM`/`ITO` annotations but also `IRQ ASSUME` and `IRQ REQUIRE`.  Assuming I have gotten the script and the latter annotations right, there are no more IRQ-related surprises lurking in the switcher.

I would very much appreciate more eyes / the same eyes again on this so that it can hopefully land soon.

PR #324, opened 2024-10-25
Make `unwind.h` a C-compatible header.

hlef:
`unwind.h` does not currently compile as a C header, which is necessary for error handling in the network stack (to use `CHERIOT_*` macros).

Assuming that this is a bug, here is a tentative fix.

### DIFF
--- a/sdk/include/setjmp.h
+++ b/sdk/include/setjmp.h
@@ -30,11 +30,12 @@ struct __jmp_buf
  */
 typedef struct __jmp_buf jmp_buf[1];
 
+__BEGIN_DECLS
 /**
  * C `setjmp` function.  Returns (up to) twice.  First returns 0, returns a
  * value passed to `longjmp` on the second return.
  */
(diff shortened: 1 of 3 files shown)


sdk/include/setjmp.h
```diff
@@ -34,7 +34,11 @@ typedef struct __jmp_buf jmp_buf[1];
  * C `setjmp` function.  Returns (up to) twice.  First returns 0, returns a
  * value passed to `longjmp` on the second return.
  */
-__attribute__((returns_twice)) extern "C" int setjmp(jmp_buf env);
+#if defined(__cplusplus)
+extern "C"
+#endif
```
davidchisnall, line 39: Please use the __BEGIN_DECLS macro like every other header.

davidchisnall (commented):
Please can you also add these to the c compile test to prevent regressions?

hlef:
Thank you for the comments, all done.

davidchisnall (approved):

PR #331, opened 2024-10-31
Make longjmp noreturn.

rmn30:
This is specified in C / C++ standards and may enable very slightly better optitmisation. Not sure what is the most portable way to specify this but this but this seems to work.


nwf (approved):
Seems OK by me, and the documentation is good even if the compiler has already attached ``noreturn`` semantics to the function by virtue of its name.

PR #338, opened 2024-11-07
clang-tidy: Added readability-braces-around-statements check

HU90m:
Ensures there are braces around statements, i.e.

```cpp
// bad:
if (condition)
    statement;
// good:
if (condition)
{
    statement;
}
```


davidchisnall (approved):
Huh, I thought that was on already.  Good catch!

davidchisnall:
Looks like a few things are going to fail.  If you run `clang-tidy --fix` on those files, it should be able to automatically fix them (and then run `clang-format -i` because it may put the braces in a silly place).

davidchisnall:
Looks as if the scheduler one doesn't have a fixit hint for some reason, but the rest appear to, so only one place to fix manually.

PR #339, opened 2024-11-07
Use named constants in the switcher

nwf:
Pull yet more out of #320.  Thanks to @rmn30 for the suggestion to make these self-check a la `assembly-helpers.h`.


davidchisnall (approved):

sdk/core/switcher/misc-assembly.h
```diff
@@ -0,0 +1,43 @@
+// Copyright CHERIoT Contributors.
+// SPDX-License-Identifier: MIT
+
+#pragma once
+#include <assembly-helpers.h>
+
+/*
+ * Constant to represent the raw permissions of the compartment CSP.  We use
+ * this in the switcher, to verify the CSP comes from the compartment is exactly
```
rmn30, line 9:
```suggestion
 * this in the switcher, to verify the permissions of the CSP that comes from the compartment are exactly
```

sdk/include/assembly-helpers.h
```diff
@@ -21,6 +21,25 @@ struct CheckSize
 	static constexpr bool value = Real == Expected;
 };
 
+/**
+ * Export a macro into assembly named `name` with value `value`.  In C++, this
+ * macro will report an error if the provided value does not equal the constexpr
+ * evaluation of `expression`.
+ */
+#	define EXPORT_ASSEMBLY_NAME(name, val)                                    \
```
rmn30, line 29: It took me a minute to work out why we need both of these macros. I guess this one is just the same as `EXPORT_ASSEMBLY_EXPRESSION(name, name, val)` but without a name collision? That makes me think maybe we should use `{}` around the definition below?

sdk/include/priv/riscv.h
```diff
@@ -75,6 +75,7 @@ namespace priv
 	constexpr size_t MCAUSE_LOAD_PAGE_FAULT     = 13;
 	constexpr size_t MCAUSE_STORE_PAGE_FAULT    = 15;
 	constexpr size_t MCAUSE_THREAD_EXIT         = 24;
+	constexpr size_t MCAUSE_THREAD_INTERRUPT    = 25;
 	constexpr size_t MCAUSE_CHERI               = 28;
```
rmn30, line 79: This could be a separate PR but these constants are needed in error handlers so we should make them more publicly available in the SDK.

davidchisnall: This is a public heading (and this is why I don't like abbreviations in names of things: priv is short for 'privileged spec' not 'private').

rmn30: I guess there's nothing stopping users from include `priv/riscv.h` it just seems a bit weird. At any rate we should do this in relevant places in the examples/tests.

sdk/include/assembly-helpers.h
nwf: We might want the names available in C++, maybe, and the invoking source could, in principle, choose to put things in scope to hide them if desired, but because these are otherwise at top-level, just using `{}` doesn't do the trick, eliciting "expected unqualified-id" errors from the compiler.

PR #342, opened 2024-11-08
Fix the build.

davidchisnall:
Don't depend on a newer C++ than our current clang supports.


nwf (approved):

resistor:
FYI, this won't work with clang-17. It requires c++23 for the use of #elifdef

PR #345, opened 2024-11-11
IRQs on in compartment fault handlers

nwf:
The core change here is the addition of _two instructions_ to the switcher.  There's a change to our crash_recovery test that makes sure we're doing the right thing (and which fails when those two new instructions are not present).  And because that wanted to use a symbolic name for one number that was previously somewhat buried, there's a lot of shuffling of symbolic names for things, to hopefully make similar things easier in the future.

### DIFF
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -777,6 +777,9 @@ exception_entry_asm:
 	cmove              csp, ca0
 
 .Linvoke_error_handler:
+	// Enable interrupts before invoking the handler
+	csrsi              mstatus, 0x8
+
 	// Clear all registers except:
 	// cra is set by cjalr.  csp and cgp are needed for the called compartment.
 	// ca0, used for the register state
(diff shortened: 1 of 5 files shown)


sdk/core/switcher/entry.S
```diff
@@ -786,6 +789,12 @@ exception_entry_asm:
 	// Call the handler.
 	cjalr              cra
 
+	/*
+	 * Now that we're back, defer interrupts again before we do anything that
+	 * manipulates the TrustedStack.
+	 */
+	csrci              mstatus, 0x8
+
```
davidchisnall, line 797: Is this necessary?  I believe it's fine to take interrupts here and we'll just handle the sentry on the way back.

davidchisnall (approved):

sdk/core/switcher/entry.S
nwf: I wasn't sufficiently sure to do anything else.  While seemingly simple, this chunk of code connects to the rest of the switcher rather deeply... The "handler asked for a context install" path is probably fine, but force unwind gets us to the cross-call return path, which presently runs with IRQs deferred, and that can get us to the thread-exit path, which *definitely* needs to run with IRQs deferred, since it zeros MTDC.

Would you be OK with this instruction in place for now and a post #320 cleanup of IRQ dispositions?

davidchisnall: I hit approve, so I'm happy as is (it isn't worse than what we have elsewhere), would be nice to have a comment suggesting that this might be safe to remove.

tests/crash_recovery-test.cc
```diff
@@ -10,9 +10,18 @@
 int               crashes = 0;
 std::atomic<bool> expectFault;
 
+static void test_irqs_are_enabled()
+{
+	void *r = __builtin_return_address(0);
+	TEST(__builtin_cheri_type_get(r) == CheriSealTypeReturnSentryEnabling,
```
rmn30, line 16: For future reference: we have `TEST_EQUAL` which gives better error messages in this situation.

PR #346, opened 2024-11-11
scheduler exception path IRQ disposition

nwf:


nwf:
It is at present possible to violate this invariant, by having a thread use an IRQ-enabling sentry from its top-level context to enter the switcher cross-call return path, which it can do, with a little help.  Here's a minimal PoC:

* `helper.cc`:
  ```
  #include "helper.h"
  [[cheri::interrupt_state(enabled)]] void* help(void)
  {
    return __builtin_return_address(0);
  }
  ```
* `helper.h`:
  ```
  #include <compartment.h>
  void* __cheri_compartment("helper") help(void);
  ```
* `top.cc`:
  ```
  #include "helper.h"
  void __cheri_compartment("top") entry()
  {
    asm volatile ("cmove cra, %0; cret" : : "C"(help()));
  }
  ```
* `xmake.lua`:
  ```
  set_project("CHERIoT Scheduler IRQ Exception PoC")
  sdkdir = "../../sdk"
  includes(sdkdir)
  set_toolchains("cheriot-clang")
  option("board")
    set_default("sail")
  compartment("helper")
    add_files("helper.cc")
  compartment("top")
    add_files("top.cc")
  firmware("top_compartment")
    add_deps("freestanding", "debug")
    add_deps("top", "helper")
    on_load(function(target)
        target:values_set("board", "$(board)")
        target:values_set("threads", {
            {
                compartment = "top",
                priority = 1,
                entry_point = "entry",
                stack_size = 0x200,
                trusted_stack_frames = 1
            }
        }, {expand = false})
    end)
  ```

Because this involves a thread exiting, I'm not sure what the best way to integrate this into the RTOS test suite is, but it should probably end up there... a sacrificial thread that blocks on a futex until told to do its thing?

davidchisnall:
We should probably put it in a separate test firmware. Eventually I want to start splitting the test suite up so we can run different parts in parallel.

nwf:
You can test that the small change to the switcher works by running the regression test both before and after that commit is applied (with `git rebase`, say).

Note that the regression test infinite-loops on failure (in the switcher's exception path), since the assertion failure traps while MTDC is 0, which causes the exception path to set MTCC to 0 and then take a trap and then die.  (This PR is not atop #320, and so is missing the pending changes in "switcher: ensure that MTCC is really NULL on fault".)
```
[111432] [M]: 0x8000170C (0x0000) c.illegal 0x0
trapping from M to M to handle illegal-instruction
handling exc#0x02 at priv M with tval 0x00000000
CSR mstatus <- 0x00001800
mem[X,0x80000C04] -> 0x015B
mem[X,0x80000C06] -> 0x03D1
[111433] [M]: 0x80000C04 (0x03D1015B) cspecialrw csp, mtdc, csp
x2 <- 0x00000000 (v:0 0x00000000-0x000000000 l:0x000000000 o:0x0 p:- ------ -- ---)
mem[X,0x80000C08] -> 0x015B
mem[X,0x80000C0A] -> 0xFEA1
[111434] [M]: 0x80000C08 (0xFEA1015B) cmove csp, csp
x2 <- 0x00000000 (v:0 0x00000000-0x000000000 l:0x000000000 o:0x0 p:- ------ -- ---)
mem[X,0x80000C0C] -> 0x1463
mem[X,0x80000C0E] -> 0x0001
[111435] [M]: 0x80000C0C (0x00011463) bne sp, zero, 0x8
mem[X,0x80000C10] -> 0x005B
mem[X,0x80000C12] -> 0x03C1
[111436] [M]: 0x80000C10 (0x03C1005B) cspecialrw cnull, mtcc, csp
mem[X,0x80000C14] -> 0xE406
[111437] [M]: 0x80000C14 (0xE406) c.cscsp cra, 0x8
CHERI TagViolation Reg=c2/csp PC=0x80000C14
trapping from M to M to handle cheri
handling exc#0x1C at priv M with tval 0x00000042
CSR mstatus <- 0x00001800
CHERI TagViolation Reg=PCC PC=0x00000000
trapping from M to M to handle cheri
handling exc#0x1C at priv M with tval 0x00000402
CSR mstatus <- 0x00001800
CHERI TagViolation Reg=PCC PC=0x00000000
...
```
We should probably teach sail that that's a good excuse to stop simulating.

sdk/core/scheduler/main.cc
```diff
@@ -258,6 +258,15 @@ namespace sched
 	                                               size_t        mepc,
 	                                               size_t        mtval)
 	{
+#if !defined(NDEBUG)
```
davidchisnall, line 261: We don’t really use these macros anywhere, this should probably be a constexpr if on the scheduler debug flag.

sdk/core/scheduler/main.cc
```diff
@@ -258,6 +258,15 @@ namespace sched
...
+		/* Ensure that we got here from an IRQ-s deferred context */
+		void *returnAddress = __builtin_return_address(0);
+		Debug::Assert(
+		  __builtin_cheri_type_get(returnAddress) ==
```
davidchisnall, line 265: This can use Capability::type, which you can do without a cast if you make the 263 a Capability instead of a bare pointer.

sdk/core/scheduler/main.cc
```diff
@@ -258,6 +258,15 @@ namespace sched
 	                                               size_t        mepc,
 	                                               size_t        mtval)
 	{
+		if constexpr (DebugScheduler)
+		{
+			/* Ensure that we got here from an IRQ-s deferred context */
+			auto returnAddress = Capability{__builtin_return_address(0)};
```
davidchisnall, line 264: I would write this as:

```suggestion
			Capability returnAddress{__builtin_return_address(0)};
```

But I don't think it matters too much.

davidchisnall (approved):

vmurali:
> It is at present possible to violate this invariant, by having a thread use an IRQ-enabling sentry from its top-level context to enter the switcher cross-call return path, which it can do, with a little help.


What invariant?

rmn30:
I think the invariant that `sched_exception_entry` runs with interrupts disabled.

vmurali:
I saw it on the commit message. I think it's better to put the info here also, as this PR is referred to in the code. Specifically:

The thread exit path zeros out MTDC as part of transitioning to the
scheduler compartment.  It's bad if we can get there with IRQs enabled,
so ensure that we cannot from switcher_skip_compartment_call.

PR #350, opened 2024-11-13
token_unseal: use cgettop

nwf:
Shave two instructions off the fast unsealer.  It's helpful to remember the ISA. :)


davidchisnall (approved):
The compiler didn’t support cgettop when this was written!

rmn30:
Just checking: this change means that `ct0` is no longer clobbered. I think that is OK as it contains the user's key which is not secret?

davidchisnall:
We never actually use *c*t0, so we could replace line 60 with a mv / cgetaddr instead of a cmove and just extract the value.  I don't think it makes a different for security, but it means that we are still clobbering the capability (which is not secret), just leaking the integer value (which is also not secret, but conveys no rights by itself).

nwf:
Oh, yea, good call.  Done, and some other cleanups while in the neighborhood; hopefully nothing objectionable.

davidchisnall (approved):

PR #353, opened 2024-11-15
Add __multi3() to crt library

### LINKED ISSUE #352: add __multi3() to crt library
The compiler believes that the x25519 layer in libhydrogen has a dependency on __multi3(), even though it can generate the code fine if asked to multiply two uint64_t  values

PhilDay-CT:
The compiler believes that the x25519 layer in libhydrogen has a dependency on __multi3(), even though it can generate the code fine if asked directly to multiply two uint64_t values.    So we'll humour it by giving it that as an implemenation.  

Closes #352

### DIFF
--- a/sdk/lib/crt/arith64.c
+++ b/sdk/lib/crt/arith64.c
@@ -25,6 +25,7 @@
 // - Add asm labels so that the compiler doesn't need to know the C++ type encodings for builtins.
 // - Use standard integer types.
 // - Add SPDX tags for license from the upstream repository (commit 426b7578ecfb5ce7c841e738613cff2a261214eb)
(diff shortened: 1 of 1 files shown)


sdk/lib/crt/arith64.c
```diff
@@ -300,3 +302,9 @@ typedef union
     __divmoddi4(a, b, &r);
     return r;
 }
+
+// Return the product of a and b
+[[clang::no_builtin]] arith64_u64 __multi3(arith64_u64 a, arith64_u64 b)
```
davidchisnall, line 307: I think this should be s64 not u64, it doesn’t actually affect codegen.

davidchisnall (approved):

PR #355, opened 2024-11-18
Allocator: expose minimum chunk size information, fix test

nwf:
`allocator-test.cc` needs to account for underlying object sizes.  Expose the information it needs to do so and update documentation on heap interface functions.


nwf:
(This fixes the CI failure in #29, at least locally.)

rmn30 (approved):

PR #357, opened 2024-11-20
Sonata v1.0 SPI block driver

HU90m:
The main changes:

1. Chip select (CS) lines are now controlled by a register in the SPI controller
   itself.
2. FIFO depths are now available in the 'info' register, and the FIFOs
   are reduced to 16 entries apiece.
3. Ensure that zero-byte transfer requests do not result in a deadlock
   between software and hardware; the controller will attempt to start
   but cannot perform a zero-byte transfer.
    *  Zero-byte blocking_write/read can be used instead to synchronise
       with the controller core.
4. Interrupts for the SPI block added.
5. Added internal loopback. Internal loopback within the SPI controller
   is trivial and very useful for testing; simultaneous transmit and
   receive shall result in all transmitted bytes being replicated
   exactly in the receive FIFO; we can test permutations of polarity and
   phase settings.
6. Added software reset. Software reset is required to recover in the
   event of an error or indeed just a re-download of the software
   without an intervening IP block reset presently. Clear the TX FIFO,
   reset the controller core and then clear the RX FIFO.
7. Adapted ethernet driver to use of the builtin SPI chip select


HU90m:
It looks like the coding conventions check is failing due to a name collision between the old SPI driver `v0.2/platform-spi.hh` and the new `platform-spi.hh`.

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,403 @@
...
+	    /// Asserted whilst the transmit FIFO level is at or above the
+	    /// watermark.
+	    InterruptTransmitWatermark = 1 << 3,
+	    /// Asserted whilst the transmit FIFO is empty.
+	    InterruptTransmitEmpty = 1 << 2,
+	    /// Asserted whilst the receive FIFO level is at or below the watermark.
+	    InterruptReceiveWatermark = 1 << 1,
+	    /// Asserted whilst the receive FIFO is full.
+	    InterruptReceiveFull = 1 << 0,
+	  } Interrupt;
```
nwf, line 23: Er, perhaps this is just a polarity thing, but I'd expect the `InterruptTransmitWatermark` to fire when the transmit FIFO is _below_ its watermark (that is, it has room to accept a positive natural number of bytes before overflow) and, dually, `InterruptReceiveWatermakr` to fire when the receive FIFO is _above_ its watermark (that is, it can source at least watermark-many bytes).

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,403 @@
...
+	 * Rendered documentation is served from:
+	 * https://lowrisc.org/sonata-system/doc/ip/spi.html
+	 */
+	template<size_t NumChipSelects = 4>
+	struct Generic : private utils::NoCopyNoMove
+	{
+		/**
+		 * The Sonata SPI block doesn't currently have support for interrupts.
+		 * The following registers are reserved for future use.
+		 */
```
nwf, line 147: Stale comment, given the code below?

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -0,0 +1,403 @@
...
+		uint32_t transmitFifo;
+		/**
+		 * Information about the SPI controller. This register reports the
+		 * depths of the transmit and receive FIFOs within the controller.
+		 */
+		uint32_t info;
+		/**
+		 * Chip Select lines; clear a bit to transmit/receive to/from the
+		 * corresponding peripheral.
+		 */
```
nwf, line 187: What happens if this has a population count other than 1?

nwf (commented):
Some nits, but generally LGTM!

sdk/include/platform/sunburst/platform-spi.hh
HU90m: Good catch! You're spot on.

- https://github.com/lowRISC/sonata-system/blob/736837cda6de2d7b6dbbe86a513d18d2eb4155e2/rtl/ip/spi/rtl/spi.sv#L168
- https://github.com/lowRISC/sonata-system/blob/736837cda6de2d7b6dbbe86a513d18d2eb4155e2/rtl/ip/spi/rtl/spi.sv#L165

sdk/include/platform/sunburst/platform-spi.hh
HU90m: All chip selects can be controlled individually, so one can have a population count of 0 (all enabled) to the number of chip selects (all disabled). (Chip selects are active low.)

https://github.com/lowRISC/sonata-system/blob/736837cda6de2d7b6dbbe86a513d18d2eb4155e2/rtl/ip/spi/rtl/spi.sv#L220-L226

Improved the comment to  better cover this.

nwf: I apologize for the polarity error, but that wasn't really the question... SPI isn't (usually) an open-drain bus, and while SCK and MOSI might be OK with multiple receivers selected (and so, presumably, not tristated)... having multiple chip-selects active at once sure seems like it's going to do bad things to the drivers on MISO.

I know SPI isn't always bidirectional, and so maybe there are as few as zero drivers on MISO, but this really feels like something needs more blinking light warnings.

nwf (approved):
I think I'm happy modulo a request for an expanded comment.  Thanks!

sdk/include/platform/sunburst/platform-spi.hh
HU90m: That's a good point. I've changed the function so that it defaults to de-asserting the other chip select. People will now have to explicitly ask for the other chip selects to remain asserted.

nwf (approved):

Issue #334, opened 2024-11-05
Future small tweaks to switcher?


nwf:
While writing (and trying to actually finish) #320, follow-up work has sprung to mind.  Just jotting down so I don't forget.  This list might grow with time.

- @davidchisnall's "assume MSHWM now" comment; it's been a long time since we've driven without it (though @rmn30 notes that it could be useful to keep configurable for benchmarking)
- `.Lswitch_trusted_stack_exhausted` looks an awful lot like a subset of `switcher_skip_compartment_call`.  Can we readily shuffle the latter so that the former can be made to look like `.Lswitch_stack_too_small`?
- `.Lhandle_error_try_stackful` and `.Lhandle_error_try_stackless` both want the value `0xffff` in `s1`; can we shuffle things around so that that's live-in to both rather than being `li`-d separately in each?

davidchisnall:
One more for the list:

 - [ ] Use MScratch as the temporary home of CSP so MTDC is always a trusted stack or null.

PR #358, opened 2024-11-22
Next round of changes to the switcher

nwf:
Pulling from https://github.com/CHERIoT-Platform/cheriot-rtos/issues/334 .  This is much, much smaller than #320 and so hopefully that much easier to review.  As usual, commit-by-commit is probably the right approach.

I am only so confident in the `mscratchc` changes, as they are not mechanically checked, though the version here does run the test suite on my machine (and several nearby versions, in some sense, do not).  We might want to tear the bottom commits off this PR, as often happens.

### DIFF
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -77,6 +77,7 @@
  *
  *  - "LIVE IN:", a list of live (in) registers at this point of the code and/or
  *    - "*": the entire general purpose register file (no CSRs or SCRs implied)
+ *    - "callee-save": the psABI callee-save registers
  *    - "mcause"
  *    - "mtdc"
  *    - "mtval"
@@ -1824,10 +1825,14 @@ exception_entry_asm:
 	.type __Z23trusted_stack_has_spacei,@function
 __Z23trusted_stack_has_spacei:
 	/*
-	 * LIVE IN: mtdc, a0
+	 * FROM: malice
+	 * IRQ ASSUME: deferred
+	 * LIVE IN: mtdc, callee-save, ra, a0
 	 *
 	 * Atlas:
 	 *  mtdc: pointer to TrustedStack (or nullptr if from buggy scheduler)
+	 *  ra: return pointer (guaranteed because this symbol is reachable only
+	 *      through an interrupt-disabling forward-arc sentry)
 	 *  a0: requested number of trusted stack frames
 	 */
 	li                 a2, TrustedStackFrame_size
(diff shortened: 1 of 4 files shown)


sdk/core/switcher/entry.S
```diff
@@ -1659,6 +1659,7 @@ exception_entry_asm:
 	 *  mtdc: pointer to TrustedStack
 	 */
 	csrw               mcause, MCAUSE_THREAD_EXIT
+	csrw               mtval, MCAUSE_THREAD_EXIT
```
rmn30, line 1725: Worth a comment here to indicate that this avoids a small info leak to the scheduler?

sdk/core/switcher/entry.S
```diff
@@ -274,21 +274,19 @@ __Z26compartment_switcher_entryz:
...
 	 */
-	csc               cs0, SPILL_SLOT_cs0(ct2)
-	csc               cs1, SPILL_SLOT_cs1(ct2)
-	csc               cgp, SPILL_SLOT_cgp(ct2)
-	csc               cra, SPILL_SLOT_pcc(ct2)
-	cmove             csp, ct2
+	csc               cs0, (SPILL_SLOT_cs0-SPILL_SLOT_SIZE)(csp)
+	csc               cs1, (SPILL_SLOT_cs1-SPILL_SLOT_SIZE)(csp)
+	csc               cgp, (SPILL_SLOT_cgp-SPILL_SLOT_SIZE)(csp)
+	csc               cra, (SPILL_SLOT_pcc-SPILL_SLOT_SIZE)(csp)
```
rmn30, line 306: Unfortunately the negative offset means we don't get to use compressed `c.scsp` instruction here. `ct2` is also not accessible to the compressed `c.csc` so the instruction saved is still a net win. Remind me why we can't do the `cincoffset` before into `sp`? So that we can recover `csp` on entry in the case where the cincoffset takes it below `csp.base`?

nwf: Thanks much for this comment.  We can, instead:
```
	csc               cs0, (SPILL_SLOT_cs0-SPILL_SLOT_SIZE)(csp)
	cincoffset        csp, csp, -SPILL_SLOT_SIZE
	csc               cs1, SPILL_SLOT_cs1(csp)
	csc               cgp, SPILL_SLOT_cgp(csp)
	csc               cra, SPILL_SLOT_pcc(csp)
```
This uses one uncompressed insruction (the first), and then compresses the rest.

Much of the rest of the switcher could stand careful analysis to move our working set into the registers available to more compressed instructions...

sdk/core/switcher/entry.S
```diff
@@ -1021,6 +1050,14 @@ exception_entry_asm:
 	csetaddr           csp, csp, gp
 	cunseal            csp, ca0, csp
 	// Atlas update: sp: unsealed target thread trusted stack pointer
+	/*
+	 * LOCAL SEAL: if the scheduler has shed GL(obal) of the reference it gave
+	 * us in a0, then sp will also lack GL(obal) after unsealing.  This
+	 * reference is not stored in memory (in the switcher, anyway), just mtdc.
+	 * However, when this reference is extracted and sealed for the next
+	 * context switch (in .Lexception_scheduler_call), the result will lack
+	 * GL(obal), which will likely prove challenging for the scheduler.
```
rmn30, line 1082: > which will likely prove challenging for the scheduler

I think this is understatement?! I'm not so familiar with the scheduler but presumably it keeps the sealed trusted stack pointers in global data structures. It sounds like this would only allow the scheduler to crash itself so not really an issue but I wonder if there are other ways a non-GL context pointer could make its way into the scheduler? `switcher_interrupt_thread` looks like it needs investigating.

PR #360, opened 2024-11-27
Sonata Pinmux driver

HU90m:
A driver for Sonata v1.0's Pinmux


marnovandermaas (approved):
Thanks Hugo, this looks correct to me from a code review.

HU90m:
I think in the case of the Pinmux it's best to make sure sink names match the canonical name used documentation, RTL and schematic. The sinks are enum classes, so breaking the code style shouldn't cause much confusion here.

nwf (changes requested):
It looks like platform-pinmux.hh is generated from [sonata-system's platform-pinmux.hh.tpl](https://github.com/lowRISC/sonata-system/blob/658a74130ba8436c4a5619671b7cace2179d6ce4/sw/cheri/common/platform-pinmux.hh.tpl) via its [top_gen](https://github.com/lowRISC/sonata-system/blob/658a74130ba8436c4a5619671b7cace2179d6ce4/util/top_gen/generator.py).  It'd be good to...
- have a comment to that effect in the file
- update the `.hh.tpl` in light of the changes made here (`NOLINT`, names of some function arguments and locals, &c).

FWIW, I ran `git diff --word-diff` having clobbered `sonata-system` `v0.3-403-g658a741`'s `sw/cheri/common/platform-pinmux.hh` with the one from this PR; that might be the easiest way to find the requisite changes to the template?

I have not attempted to validate that `top_gen` is consistent between the RTL and C++ it generates, much preferring to merely assume so. :)

HU90m:
↑ rebased onto main, ↓ added comment pointing to what generates the file.

*See https://github.com/lowRISC/sonata-system/pull/376*

nwf:
Thanks for doing that.  One nit aside, this LGTM.

nwf:
Want to update the `platform-pinmux.hh` herein with the result of that PR?  I think it's good to go after that.

HU90m:
> Want to update the `platform-pinmux.hh` herein with the result of that PR? I think it's good to go after that.

The `sonata-system` repo uses the lowRISC C++ formatting, *unlike `sonata-software`*, so the output of the topgen script has to be lowRISC formatted.

This PR has already been updated with the updated with the latest output, only run through clang-format with the CHERIoT settings.

davidchisnall:
I missed in review, but this seems to have an Apache license.  Is it possible to relicense to match the rest of the repository?

PR #361, opened 2024-11-27
Simplified the sonata PWM driver

HU90m:


sdk/include/platform/sunburst/platform-pwm.hh
```diff
@@ -46,23 +30,45 @@ struct SonataPulseWidthModulation
...
+		void output_set(uint8_t period, uint8_t dutyCycle) volatile
+		{
+			this->period    = period;
+			this->dutyCycle = dutyCycle;
+		}
+	};
+
+	/// A convenience structure that can map onto multiple PWM outputs.
+	template<size_t NumberOfPwms = 6>
+	struct Array
```
davidchisnall, line 55: Is there a reason that this isn't a `std::array`?  It looks like it's equivalent to a subset of `std:array`.

HU90m: I originally used `std::array` but found a lot of `std::array`'s methods don't work with `utils::NoCopyNoMove` elements.

marnovandermaas (approved):
This looks correct to me from a code inspection.

sdk/include/platform/sunburst/platform-pwm.hh
```diff
@@ -1,37 +1,21 @@
 #pragma once
+#include <array>
```
davidchisnall, line 2: I think this is stale now?

sdk/include/platform/sunburst/platform-pwm.hh
```diff
@@ -46,23 +30,45 @@ struct SonataPulseWidthModulation
...
+		}
+	};
+
+	/**
+	 * There are six general purpose PWM outputs are general purpose that can be
+	 * pinmuxed to different outputs.
+	 */
+	using General = Array<6>;
+	/// There is one dedicated PWM for the LCD backlight.
+	using Lcd = Output;
```
davidchisnall, line 73: Maybe call this `LcdBacklight`?  It's a bit confusing to have a thing called Lcd that isn't the LCD interface.

davidchisnall (approved):

sdk/include/platform/sunburst/platform-pwm.hh
HU90m: Ah yes good spot!

nwf:
I think this is ready modulo some small deltas.  Would you like to make those @HU90m or should I just make a follow-up commit?

marnovandermaas:
> I think this is ready modulo some small deltas. Would you like to make those @HU90m or should I just make a follow-up commit?

Hugo is currently out for vacation I believe so might be better to merge and make a follow-up commit to speed things along. Thanks, Wes!

HU90m:
> > I think this is ready modulo some small deltas. Would you like to make those @HU90m or should I just make a follow-up commit?
> 
> Hugo is currently out for vacation I believe so might be better to merge and make a follow-up commit to speed things along. Thanks, Wes!

I was planning on quickly doing these changes this afternoon now I have a little time, but do feel free to merge now if you want the changes in pronto

PR #362, opened 2024-11-27
token_unseal: wield three unsealing authorities (& some tidying)

nwf:
Merge only if https://github.com/CHERIoT-Platform/cheriot-sail/pull/87 is not reverted.

### DIFF
--- a/sdk/core/loader/boot.cc
+++ b/sdk/core/loader/boot.cc
@@ -1267,15 +1267,31 @@ extern "C" SchedulerEntryInfo loader_entry_point(const ImgHdr &imgHdr,
 	              sizeof(void *),
 	              PermissionSet{Permission::Global, Permission::Unseal});
 
-	setSealingKey(imgHdr.allocator(), Allocator);
+	/*
+	 * The token library unseals both static and dynamic objects, sometimes
+	 * either and sometimes with static knowledge of which is expected.  To
+	 * avoid `li; csetaddr; csetbounds` sequences, we give it a separate cap
+	 * for each case.
+	 */
 	setSealingKey(imgHdr.token_library(),
 	              Allocator,
 	              2, // Allocator and StaticToken
 	              0,
 	              PermissionSet{Permission::Global, Permission::Unseal});
+	setSealingKey(imgHdr.token_library(),
+	              StaticToken,
+	              1,
+	              sizeof(void *),
+	              PermissionSet{Permission::Global, Permission::Unseal});
+	setSealingKey(imgHdr.token_library(),
+	              Allocator,
+	              1,
+	              2 * sizeof(void *),
+	              PermissionSet{Permission::Global, Permission::Unseal});
+
 	constexpr size_t DynamicSealingLength =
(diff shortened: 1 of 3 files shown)


sdk/core/token_library/token_unseal.S
```diff
@@ -127,8 +115,12 @@ _Z16token_obj_unsealP10SKeyStructP10SObjStruct:
 	.hidden  _Z23token_obj_unseal_staticP10SKeyStructP10SObjStruct
 	.globl   _Z23token_obj_unseal_staticP10SKeyStructP10SObjStruct
 _Z23token_obj_unseal_staticP10SKeyStructP10SObjStruct:
-	li       a2, CheriSealTypeStaticToken
-	j        .Ltoken_unseal_internal
+	auipcc     ca2, %cheriot_compartment_hi(__sealingkey)
+	clc        ca2, %cheriot_compartment_lo_i(_Z23token_obj_unseal_staticP10SKeyStructP10SObjStruct)(ca2)
+	li         t0, CheriSealTypeStaticToken
+    csetaddr   ca2, ca2, t0
+    csetbounds ca2, ca2, 1
```
rmn30, line 122: I guess we could avoid these extra instructions by having three versions of `__sealingKey` in memory and loading the appropriate one? I think that would be a net win on space and cycles?

nwf: Yeah, I had just the same realization this morning: a `csetaddr; csetbounds` pair alone is the size of a capability.

nwf:
For the record, assuming I can count correctly: The instructions proper shrink by two bytes in aggregate (hooray compression), and we go from 25 instructions on all success paths to either 21 (_static or _dynamic) and 23 (dynamic decision), but the whole thing goes from 134 bytes to 148 bytes.  That drops to 21 instructions on all success paths and we claw back eight of bytes once https://github.com/CHERIoT-Platform/cheriot-sail/pull/87 is everywhere.

sdk/core/loader/boot.cc
```diff
@@ -1267,15 +1267,31 @@ extern "C" SchedulerEntryInfo loader_entry_point(const ImgHdr &imgHdr,
...
-	setSealingKey(imgHdr.allocator(), Allocator);
+	/*
+	 * The token library unseals both static and dynamic objects, sometimes
+	 * either and sometimes with static knowledge of which is expected.  To
+	 * avoid `li; csetaddr; csetbounds` sequences, we give it a separate cap
+	 * for each case.
+	 */
 	setSealingKey(imgHdr.token_library(),
 	              Allocator,
 	              2, // Allocator and StaticToken
```
rmn30, line 1278: Could we have a `static_assert` here that `StaticToken == (Allocator + 1)`?

nwf: It's already there, further up:
https://github.com/CHERIoT-Platform/cheriot-rtos/blob/049fb9e328f383171937c32ec1927abd9b61eef1/sdk/core/loader/boot.cc#L155-L158

rmn30 (approved):

PR #363, opened 2024-12-02
tests/misc: check sealed vs. local/global

nwf:


tests/misc-test.cc
```diff
@@ -216,13 +216,56 @@ void check_capability_set_inexact_at_most()
...
+
+	/*
+	 * Use CAndPerm to shed Global from our o cap.
+	 * Spell this a little oddly to make sure we get CAndPerm with a mask of
+	 * all 1s but Global.  Using oLocal2.permissions().without() would do a
+	 * cgetperm and then candperm.
+	 */
+	Capability<void> oLocal2 = o;
+	oLocal2.permissions() &=
+	  PermissionSet::from_raw(-1).without(Permission::Global);
```
nwf, line 255: I'm a little worried that even this spelling isn't quite enough.  I'd not used `::omnipotent()` because I failed to notice that `::from_raw()` _also_ masks with `valid_permission_mask()` and so I'm worried that we're presently relying on the compiler to know that bits above those set by `valid_permission_mask()` are irrelevant?  Maybe we should have a `PermissionSet::allButGlobal` constant that works for the stringent requirement of https://github.com/CHERIoT-Platform/cheriot-sail/pull/83 ?

tests/misc-test.cc
```diff
@@ -240,13 +240,56 @@ namespace
...
+	  oLocal1.is_valid() && (oLocal1.type() == o.type()) &&
+	    (oLocal1.permissions() == o.permissions().without(Permission::Global)),
+	  "Loading global sealed cap through non-LoadGlobal authority gone wrong: "
+	  "{} {}",
+	  o,
+	  oLocal1);
+
+	/*
+	 * Use CAndPerm to shed Global from our o cap.
+	 * Spell this a little oddly to make sure we get CAndPerm with a mask of
```
rmn30, line 273: I think we should add a `drop_perms` method to `Capability` to do this. The architectural permissions were ordered to optimize dropping specific permissions in this way for common permissions to drop (e.g. LM, LG).

sdk/include/cheri.hh
```diff
@@ -812,6 +812,15 @@ namespace CHERI
 			return permission_set_from_pointer(ptr);
 		}
 
+		/**
+		 * Drop some permissions from this capability
```
rmn30, line 816: Is it worth extending this comment to say that you must use this if dropping global from a sealed cap and that it may produce very slightly smaller code?

tests/misc-test.cc
```diff
@@ -240,13 +240,55 @@ namespace
...
+	Capability<Capability<void>> oPNoLoadGlobal = oP;
+	oPNoLoadGlobal.drop_permissions(Permission::LoadGlobal);
+	Capability<void> oLocal1 = *oPNoLoadGlobal;
+
+	TEST(oLocal1.is_valid(),
+	     "Loading global sealed cap through non-LoadGlobal invalid");
+	TEST_EQUAL(oLocal1.type(),
+	           o.type(),
+	           "Loading global sealed cap through non-LoadGlobal bad type");
+	TEST_EQUAL(static_cast<PermissionSet>(oLocal1.permissions()),
```
rmn30, line 267: I had to add this cast in the same situation recently and it is slightly annoying. Could we have an implicit conversion or is that a bad idea? You can also get rid of if you make o1 const as there is a const version of the permissions method.

sdk/include/cheri.hh
```diff
@@ -812,6 +812,20 @@ namespace CHERI
...
+		/**
+		 * Drop some permissions from this capability.
+		 *
+		 * Because this function computes the permission mask by clearing bits
+		 * in the PermissionSet::omnipotent() all-ones mask, rather than from
+		 * the set of permissions currently held by this Capability, it is safe
+		 * to use to clear Global permission on a sealed capability.
+		 */
+		template<std::same_as<Permission>... Permissions>
+		void drop_permissions(Permissions... drop)
```
davidchisnall, line 824: We already have a without method on PermissionSet, please can you make the naming of this consistent?

PR #369, opened 2024-12-06
Various changes needed to make network stack reset more reliable.

hlef:
To perform a network stack reset, we must go through all locks and futexes used by all threads in the network stack and set them for destruction. This means waking up all waiters, making them leave, and making it impossible for other threads to block on these locks and futexes again.

Some of these locks are dynamically allocated.

To be able to go through them at reset time, we keep track of them in a linked list. At reset time we go through the list and set all locks and futexes for destruction.

One current limitation of the network stack is that, if this list gets corrupted, resetting into a clean state will be impossible. This is a relevant attack vector as it allows permanent DoS on the network stack.

In this PR, I introduce a mechanism to address this limitation.

- Blocking in "steps": instead of attempting to acquire a lock with the timeout in one chunk, try to acquire the lock in a loop of steps of duration `step`, until the timeout expires. In this way we are guaranteed that the thread will become live again in a bounded amount of time (`step`), and try to re-acquire the lock: we can thus free the lock through heap-free-all, and have the "lost" thread crash when re-trying to acquire the lock.
- Blocking with "conditions": however, it is not always possible to heap-free-all to free the lock, e.g., if the lock was allocated with a caller's quota. In this case, we can call the lock with a condition that checks, at each "step", if a reset is ongoing (e.g., very simply through checking a bit in the state machine word) to prevent the "lost" thread from blocking again.

This PR implements this in lock guards and event groups.


hlef:
Note that the related network stak PR is https://github.com/CHERIoT-Platform/network-stack/pull/54.

sdk/include/locks.hh
```diff
@@ -282,6 +282,39 @@ class LockGuard
 		try_lock(timeout);
 	}
 
+	/// Constructor, attempts to acquire the lock with a timeout, in steps
+	/// of a given duration, checking the `condition` function at every
+	/// iteration. If `condition` returns `true`, stop trying to acquire
+	/// the lock.
```
davidchisnall, line 288: Please use /** for doc comments that are longer than a single line.

sdk/include/locks.hh
```diff
@@ -282,6 +282,39 @@ class LockGuard
...
+	[[nodiscard]] explicit LockGuard(Lock    &lock,
+	                                 Timeout *timeout,
+	                                 Ticks    step,
+	                                 auto condition) requires(TryLockable<Lock>)
```
davidchisnall, line 292:
```suggestion
	                                 auto condition= []() { return false; }) requires(TryLockable<Lock>)
```

Should eliminate the need for the other constructor overload.

sdk/lib/event_group/event_group.cc
```diff
@@ -8,6 +8,17 @@
...
+/**
+ * Length of a locking step for even group locks and futex words. When we
+ * attempt to grab an event group lock or wait on a futex, do so in steps of
+ * `EventGroupLockTimeoutStep`. That way, if the user of the event group
+ * crashes and we somehow do not have a pointer to the event group lock anymore
+ * (and thus cannot reset it), we have a guarantee that the blocking thread
+ * will become live again within a bounded amount of time, and crash as it is
+ * trying to re-acquire the lock if we free it through heap-free-all.
+ */
+static constexpr const Ticks EventGroupLockTimeoutStep = MS_TO_TICKS(500);
```
davidchisnall, line 20: It would be nice to make this configurable somehow.  Waking up every 500ms will really hurt power efficiency for a load of things that are mostly asleep.

Maybe make it a parameter and, in the FreeRTOS wrapper, expose a define for configuring it?  That way the network stack can set it to a smallish value.

hlef: Sounds good, agree.

davidchisnall:
@hlef, are you still working on this?  Do we still need it?

hlef:
> @hlef, are you still working on this? Do we still need it?

Well, I am not working on it right now because this is blocking on @nwf. We would like to avoid the polling, and Wes was thinking of switcher changes that would introduce a notion of "interrupted compartment invocation", and of an API that would mark all invocations of a given compartment as interrupted. I believe that Wes shared a work document with you at some point (will re-send privately).

PR #373, opened 2024-12-10
Improve a comment and fix heap_can_free.

davidchisnall:
Previously `heap_can_free` would succeed in a situation where `heap_free` would fail with an insufficient-stack error.

### DIFF
--- a/sdk/core/allocator/main.cc
+++ b/sdk/core/allocator/main.cc
@@ -923,10 +923,14 @@ __cheriot_minimum_stack(0x1c0) ssize_t
 	return 0;
 }
 
-__cheriot_minimum_stack(0xf0) int heap_can_free(SObj  heapCapability,
-                                                void *rawPointer)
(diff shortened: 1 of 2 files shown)


sdk/core/allocator/main.cc
```diff
@@ -923,17 +923,22 @@ __cheriot_minimum_stack(0x1c0) ssize_t
...
 
-__cheriot_minimum_stack(0xf0) int heap_can_free(SObj  heapCapability,
-                                                void *rawPointer)
+__cheriot_minimum_stack(0x260) int heap_can_free(SObj  heapCapability,
+                                                 void *rawPointer)
 {
-	STACK_CHECK(0xf0);
+	// This function requires much less space, but we claim that we require as
+	// much as `heap_free` so that a call to `heap_free` will not fail due to
+	// insufficient stack immediately after `heap_can_Free` has said that it's
```
nwf, line 931:
```suggestion
	// insufficient stack immediately after `heap_can_free` has said that it's
```

nwf (approved):

PR #377, opened 2024-12-12
Heap render debugging facility

nwf:
This has been kicking around locally and undergoing the occasional modification, but it might be useful to push into the tree even if we never use it in most cases?

### DIFF
--- a/sdk/core/allocator/alloc.h
+++ b/sdk/core/allocator/alloc.h
@@ -2811,4 +2811,155 @@ class MState
 	{
 		ABORT();
 	}
+
+#if HEAP_RENDER
+	public:
+	/**
+	 * "Render" the heap for debugging.
+	 */
+	template<bool Asserts = false>
+	void render()
+	{
+		size_t measuredAllocated = 0, measuredFree = 0, measuredQuarantined = 0;
+
+		auto toAddr = [](void *p) {
+			return static_cast<ptraddr_t>(CHERI::Capability{p}.address());
+		};
+
+		auto header =
+		  static_cast<MChunkHeader *>(static_cast<void *>(heapStart));
+
+		Debug::log("Dumping MState={} start={} end={}",
+		           toAddr(this),
+		           toAddr(heapStart),
+		           heapStart.top());
+
+		while (toAddr(header) != heapStart.top())
+		{
+			Debug::log("  header {}: size={} inuse={} pinuse={}",
+			           toAddr(header),
+			           header->size_get(),
+			           header->isCurrInUse,
+			           header->isPrevInUse);
+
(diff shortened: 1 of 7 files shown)


davidchisnall (commented):
Looks useful, but shouldn't be behind `NDEBUG`.

rmn30 (approved):
I haven't reviewed the render code in detail but given this is entirely guarded by ifdefs I can't see any harm in merging. Maybe add a call to `render_heap` in a test?

rmn30:
I tried this locally by adding a call to `heap_render` in the allocator test. It seems to work but I had to pass both `--debug-allocator=y` and `--allocator-rendering=y` to `xmake config` to get it to show anything. We should add this call in the test and enable the flag in CI debug builds.

nwf:
Hm.  Maybe we shouldn't use the generic `Debug` so that `--allocator-rendering` and `--debug-allocator` are orthogonal.  Have pushed patch to that effect; please advise.

rmn30:
That's reasonable but I wonder if we also want a mode that will just do the assertions without the verbosity? Probably something we should add if we ever want it.

nwf:
> That's reasonable but I wonder if we also want a mode that will just do the assertions without the verbosity? Probably something we should add if we ever want it.

Oh, that's a great idea and easy, just move the `using RenderDebug` into the function and add another template paramater... done. :)

rmn30:
I don't think that will work as written because `ConditionalDebug::Assert` is only enabled when the condition is true. I think it will work if you use `ConditionalDebug::Invariant` instead.

nwf:
Some *truly* terrifying C++ later...

sdk/core/allocator/alloc.h
```diff
@@ -2811,4 +2811,163 @@ class MState
 	{
 		ABORT();
 	}
+
+#if HEAP_RENDER
+	public:
+	/**
+	 * "Render" the heap for debugging.
+	 */
+	template<bool Asserts = false, bool Chatty = true>
```
rmn30, line 2820: I think Asserts should default to `true`. They don't look that expensive compared to the debug logs (I guess `ds::linked_list::is_well_formed` may be linear in length of list?).

rmn30 (approved):
Looks fine. I will defer to @davidchisnall  for the `debug.hh` bits...

sdk/core/allocator/alloc.h
nwf: Thanks!

nwf: > I guess `ds::linked_list::is_well_formed` may be linear in length of list?

No, is just a check of the well-formedness of the node.  Could probably be better named, in retrospect.

PR #375, opened 2024-12-12
Allocator test

sleffler:
I have a target platform where the heap is 32KB. I want to run the test suite but the allocator test fails because the current (fixed) BigAllocSize is too big. This change calculates BigAllocSize from the heap size. The change works for the sail board config but not my target w/ 32KB of heap:
```
19:19:30.9381 [INFO] uart5: [output] [35mallocator-test.cc:376[31m Invariant failure[35m in test_claims[36m
19:19:30.9400 [INFO] uart5: [output] Heap capability still valid after releasing last claim: 0x50078200 (v:1 0x50078200-0x50078280 l:0x80 o:0x0 p: G RWcgm- -- ---)
```
Don't immediately see the problem so looking for help...

PR #378, opened 2024-12-12
allocator: fix free space accounting

nwf:
While investigating #375, I found that the allocator occasionally double-counts free space.  The data structure is otherwise fine, but this probably means we eventually start telling clients that things could succeed when they can't.

The fix is to move the onus of free space accounting to the callers of `mspace_free_internal`, which know whether they're returning memory already accounted as allocated, quarantined, or free.

### DIFF
--- a/sdk/core/allocator/alloc.h
+++ b/sdk/core/allocator/alloc.h
@@ -1207,6 +1207,7 @@ class MState
 			           "quota is {})",
 			           header->size_get(),
 			           quota);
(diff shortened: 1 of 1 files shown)


rmn30 (commented):
Should mention here that it is the caller's responsibility to update `heapFreeSize`, or why not use a boolean parameter?

rmn30 (approved):

PR #379, opened 2024-12-16
catch up with the new compiler

nwf:
I thought this would be easy, but alas.  So in lieu of a finished artifact, here's where things stand right now.  Hopefully this still passes the old CI, but it's causing linker failures locally.  I'll sic @resistor on it. :)


nwf:
This is frustrating.  I am unable to reproduce the ibex release build CI failure locally (with, admittedly, the new compiler), but can reproduce it in the current devcontainer and have noted that...
- Changing `--stack-usage-check-allocator` to `n` removes the failure, and
- it is possible to raise the stack check thresholds so that a `--debu-allocator=y` `-m release` build succeeds, specifically, for future reference as follows, ...
  - `heap_quarantine_empty` `0xf0`
  - `heap_allocate` `0x2b0`
  - `heap_free` `0x2e0`
  - `heap_free_all` `0x1f0`
  - `token_sealed_unsealed_alloc` `0x2e0`
  - `token_obj_can_destroy` `0x210`
- ... but even at those elevated thresholds, setting `--debug-allocator=n` fails in the same way.

Hopefully those data are useful to someone, possibly future me, because I'm going to set this aside for the moment.

rmn30:
Good to see the new warnings catching some genuine issues :-)

nwf:
ff029bdc passes CI with the new compiler; see https://github.com/nwf/cheriot-rtos/actions/runs/12585466330 .

nwf:
Ah ha.  And now it passes CI with the old compiler, too. :D

sdk/core/allocator/main.cc
```diff
@@ -1183,7 +1192,7 @@ __cheriot_minimum_stack(0x280) SObj
 			return sealed;
 		}
 	}
-	heap_free(heapCapability, obj);
+	(void)heap_free(heapCapability, obj);
```
davidchisnall, line 1195: Would be nice to have a comment here.  This can happen only if the caller gives us a malicious output value.  Actually, we probably can elide this and return sealed unconditionally.  If you give us a valid `unsealed` pointer and we don't write through it, you can still free with the sealed pointer.

sdk/core/loader/boot.cc
```diff
@@ -1460,6 +1460,7 @@ extern "C" SchedulerEntryInfo loader_entry_point(const ImgHdr &imgHdr,
 	// invoke the exception entry point.
 	auto exportEntry = build<ExportEntry>(
 	  imgHdr.scheduler().exportTable,
+	  // XXX NWF
```
davidchisnall, line 1463: ???

sdk/core/scheduler/main.cc
```diff
@@ -31,9 +31,10 @@ using namespace CHERI;
 /**
  * Exit simulation, reporting the error code given as the argument.
  */
-void simulation_exit(uint32_t code)
+int simulation_exit(uint32_t code)
```
davidchisnall, line 34: Why is this an `int`?  I don't think any code should ever be expected to handle this returning.  The fallback mode should probably be to infinite loop.

sdk/include/fail-simulator-on-error.h
```diff
@@ -67,7 +67,7 @@ compartment_error_handler(ErrorState *frame, size_t mcause, size_t mtval)
 		DebugErrorHandler::log("Unhandled error {} at {}", mcause, frame->pcc);
 	}
 
-	simulation_exit(1);
+	(void)simulation_exit(1);
```
davidchisnall, line 70: As above, so far every call to this function has discarded the return result.  It either exits (and doesn't return) or doesn't exit (and we're not in a simulator that can exit).  We get no information from the non-void return.

sdk/include/simulator.h
```diff
@@ -3,14 +3,18 @@
...
 #include <compartment.h>
+#include <errno.h>
 #include <stdint.h>
 
 #ifdef SIMULATION
 /**
  * Exit simulation, reporting the error code given as the argument.
  */
-[[cheri::interrupt_state(disabled)]] void __cheri_compartment("sched")
+[[cheri::interrupt_state(disabled)]] int __cheri_compartment("sched")
```
davidchisnall, line 13: We can just silence the warning here, I think.  Or we can rename this to `scheduler_exit_simulation` and have a `void` wrapper that calls it.

sdk/include/stdlib.h
```diff
@@ -282,7 +282,7 @@ ssize_t __cheri_compartment("alloc")
  * state.  It can block indefinitely if another thread is allocating and
  * freeing memory while this runs.
  */
-void __cheri_compartment("alloc") heap_quarantine_empty(void);
+int __cheri_compartment("alloc") heap_quarantine_empty(void);
```
davidchisnall, line 285: Please update the comment to say that this returns 0 or -ECOMPARTMENTFAIL.

sdk/lib/cxxrt/guard.cc
```diff
@@ -72,8 +72,9 @@ namespace
 		void unlock()
 		{
 			assert(high == LockBit);
-			high = 0;
-			futex_wake(&high, std::numeric_limits<uint32_t>::max());
+			high    = 0;
+			int res = futex_wake(&high, std::numeric_limits<uint32_t>::max());
+			assert(res >= 0);
```
davidchisnall, line 77: Please don't use `assert`, I don't think we use it anywhere else in the codebase.  It is there *purely* because C99 requires it, not because we should use it.

sdk/core/loader/boot.cc
nwf: Whoops!  Sorry, that was a note to self in some of the switcher reworking that got committed accidentally.

sdk/core/allocator/main.cc
nwf: Moved work to #401.

sdk/core/scheduler/main.cc
nwf: It's a cross-compartment call, so `void` trips the compiler's warning.  I'll annotate it as `noreturn` and make it loop, tho'.

sdk/lib/cxxrt/guard.cc
nwf: Would you prefer a different spelling or merely discarding the result entirely?

sdk/core/scheduler/main.cc
davidchisnall: It's better to silence the warning.  None of the callers gracefully handle it failing in a way that cares about return results, they just continue if they are not in a simulation.  It's always a best-effort thing for debugging.

sdk/lib/cxxrt/guard.cc
nwf: #404

sdk/core/allocator/main.cc
```diff
@@ -224,7 +224,10 @@ namespace
 			// Release the lock before sleeping
 			g.unlock();
 			Timeout smallSleep{1};
-			thread_sleep(&smallSleep);
+			if (thread_sleep(&smallSleep) < 0)
```
rmn30, line 227: Not going to block merge but would be nice to have a union type for this. Did you audit all calls to `thread_sleep`, even ones that didn't trigger the `nodiscard` warning?

sdk/core/allocator/software_revoker.h
```diff
@@ -9,7 +9,7 @@
  * revocation pass, it will scan a region of memory and then return.
```
rmn30, line 9: Ought to have a comment here describing the return value. Not a blocker.

sdk/core/software_revoker/revoker.cc
```diff
@@ -159,15 +159,17 @@ namespace
...
-void revoker_tick()
+int revoker_tick()
 {
 	// If we've been asked to run, make sure that we're running.
 	if (state == State::NotRunning)
 	{
 		advance();
 	}
 	// Do some work.
-	return scan_range();
```
rmn30, line 170: Why was there a return in a void function!?

sdk/include/thread_pool.h
```diff
@@ -94,7 +94,14 @@ namespace thread_pool
...
 			}
 			(*fn)();
-			token_obj_destroy(MALLOC_CAPABILITY, key, static_cast<SObj>(rawFn));
+			/*
+			 * This fails only if the thread pool runner compartment can't make
+			 * cross-compartment calls to the allocator at all, since we're
+			 * in its initial trusted activation frame and near the beginning
+			 * (highest address) of its stack.
+			 */
+			(void)token_obj_destroy(
```
rmn30, line 103: Worthy of a debug assertion?

tests/allocator-test.cc
```diff
@@ -471,21 +499,19 @@ namespace
 		     ptr2);
 		state = 2;
 		// Yield to allow the hazards to be dropped.
-		Timeout t{1};
-		thread_sleep(&t);
+		TEST(sleep(1) >= 0, "Failed to yield to drop hazards");
 		// Try a double free.  This may logically succeed, but should not affect
 		// our quota.
-		heap_free(SECOND_HEAP, ptr);
+		(void)heap_free(SECOND_HEAP, ptr);
```
rmn30, line 505: Why doesn't this one test the return value of `heap_free`?

rmn30 (approved):
LGTM

sdk/include/thread_pool.h
nwf: Somewhat tricky since this is a header included in the users of the thread pool compartment.

tests/allocator-test.cc
nwf: Hm, I don't know why I just dropped that one!  Fixed.

PR #382, opened 2024-12-18
Introduce tests.extra/hardware_revoker_IRQs

nwf:


nwf:
With #384 and the gateware fixes mentioned therein, this passes.  Hopefully it's a good regression test and assist for platform bringup.

rmn30 (approved):

PR #387, opened 2024-12-19
allocator-test/fuzz: stochastic necromancy test

nwf:
Hold on to a random collection of freed pointers (stored before we free them) and, after each round, scan the collection looking for asserted tags.  This could happen if an object is prematurely released from quarantine before the revoker gets to the copy here in the collection.


nwf:
While in the neighborhood, add a basic preflight test to hopefully catch things like the load barrier or revoker not working or being misconfigured (inspired by the report in https://github.com/CHERIoT-Platform/cheriot-rtos/pull/375, even if it isn't the actual problem over there, in the end).

rmn30 (approved):
Nice!

PR #388, opened 2024-12-20
Add trylock to ticket locks.

davidchisnall:
It's not impossible to do, it's just annoying, I realised while writing the book paragraph explaining that it's impossible.


davidchisnall:
Updating with offline discussion: this code is currently wrong because the unlock path does not do a futex wake if there are no outstanding waiters. Not sure if it’s worth fixing, we currently have no users of flag locks.

PR #391, opened 2024-12-23
Hardware revoker: correct wait for indicated epoch

nwf:
The commentary says that we wait for the indicated (even) epoch to have finished, but in practice we were merely waiting for the next one to begin.  Because has_revocation_finished_for_epoch() had the intended semantics and is the actual function used to gate release from quarantine, wait_for_completion()'s incorrect waiting was merely ineffecient, making malloc_internal spin a bit in the "Quarantine has enough memory to satisfy allocation" case.

### DIFF
--- a/sdk/include/platform/ibex/platform-hardware_revoker.hh
+++ b/sdk/include/platform/ibex/platform-hardware_revoker.hh
@@ -121,7 +121,11 @@ namespace Ibex
 		}
 
 		/**
-		 * Queries whether the specified revocation epoch has finished.
+		 * Queries whether the specified revocation epoch has finished, or,
+		 * if `AllowPartial` is true, that it has (at least) started.
+		 *
+		 * `epoch` must be even, as memory leaves quarantine only when
+		 * revocation is not in progress.
 		 */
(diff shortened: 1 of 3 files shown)


nwf:
Sat atop #379, this passes in the updated devcontainer with the new SAFE (which also brings the new compiler): https://github.com/nwf/cheriot-rtos/actions/runs/12585500071

sdk/include/platform/ibex/platform-hardware_revoker.hh
```diff
@@ -121,7 +121,11 @@ namespace Ibex
...
 
 		/**
-		 * Queries whether the specified revocation epoch has finished.
+		 * Queries whether the specified revocation epoch has finished, or,
+		 * if `AllowPartial` is true, that it has (at least) started.
+		 *
+		 * `epoch` must be even, as memory leaves quarantine only when
+		 * revocation is not in progress.
 		 */
 		template<bool AllowPartial = false>
```
davidchisnall, line 130: Do we have any users of this template parameter left?

davidchisnall (approved):

PR #395, opened 2024-12-31
Prune stale FLUTE support

nwf:


azure-pipelines.yml
```diff
@@ -6,9 +6,6 @@ resources:
   - pipeline: LLVM
     project: CHERI-MCU
     source: LLVM
```
davidchisnall, line 8: I think we can lose this file entirely.

davidchisnall (approved):

PR #396, opened 2025-01-01
fix xmake nightly

nwf:
I, uh, fixed(?) it, in that it now runs the test we mean to be running to completion successfully, but I don't like it.

There were several things wrong:
- `xmake`'s profile script uses `source`, which `sh` doesn't (and doesn't have to) understand
- https://github.com/xmake-io/xmake/issues/6030
- Copy-paste mistake meant that we weren't passing a board JSON to `xmake config` if we got there.

### DIFF
--- a/.github/workflows/test-new-xmake-nightly.yml
+++ b/.github/workflows/test-new-xmake-nightly.yml
@@ -5,6 +5,15 @@ on:
     - cron: '0 0 * * *'
   workflow_dispatch:
 
+defaults:
+  run:
+    # The `sh` used by default does not understand `source` which `xmake` uses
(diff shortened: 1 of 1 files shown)


nwf:
For future!me: try `daemontools` `fghack` instead of `strace`.

davidchisnall (approved):

nwf:
1dff2d3 runs cleanly, and much faster, with `fghack` instead of `strace`: https://github.com/nwf/cheriot-rtos/actions/runs/12580665704 .  Going to merge this version unless there are objections.

davidchisnall (approved):

Issue #397, opened 2025-01-02
Disable machine outlining at build time when targeting boards without weakened return sentries


resistor:
See the SAIL change in https://github.com/CHERIoT-Platform/cheriot-sail/issues/85 and the compiler issue https://github.com/CHERIoT-Platform/llvm-project/issues/46

When building the RTOS, we need to add `-mllvm -disable-global-outlining` to CFLAGS / CXXFLAGS when targeting boards that won't have the SAIL fix incorporated.

PR #398, opened 2025-01-02
Allow boards to specify CXFLAGS, and use that to disable machine outlining for Sonata.

resistor:
Fixes https://github.com/CHERIoT-Platform/cheriot-rtos/issues/397

### DIFF
--- a/sdk/boards/sonata-prerelease.json
+++ b/sdk/boards/sonata-prerelease.json
@@ -93,6 +93,7 @@
         "ipconfigDRIVER_INCLUDED_RX_IP_CHECKSUM=1",
         "ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM=1"
     ],
+    "cxflags": "-enable-machine-outliner=never",
     "driver_includes" : [
         "../include/platform/sunburst",
         "../include/platform/ibex",
--- a/sdk/xmake.lua
+++ b/sdk/xmake.lua
@@ -350,6 +350,13 @@ rule("firmware")
 			end)
 		end
 
+		-- Add cxflags to all dependencies.
+		local add_cxflags = function (cxflags)
+			visit_all_dependencies(function (target)
+				target:add('cxflags', cxflags, {force = true})
+			end)
+		end
+
 		local software_revoker = false
 		if board.revoker then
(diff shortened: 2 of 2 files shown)


davidchisnall:
Functionality looks good.  I'm a little unsure about naming.  Boards will probably also want to provide lists of extensions that they support at some point, which will be assembly flags.  We also treat defines separately.  Are there any cases where we may want to pass different flags for C and C++?  Currently, xmake uses cflags for C flags, cxxflags for C++ flags, and cxflags for C/C++ flags.  It might make sense to use cxflags here for consistency?

sdk/boards/sonata-0.2.json
```diff
@@ -75,6 +75,7 @@
         "ipconfigDRIVER_INCLUDED_RX_IP_CHECKSUM=1",
         "ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM=1"
     ],
+    "cflags": "-enable-machine-outliner=never",
```
davidchisnall, line 78: I *think* this is not needed on 0.2.  I'll check, but I believe 0.2 has an old enough version of Ibex that it doesn't introduce the problem that the latest one fixes.

resistor:
> Functionality looks good. I'm a little unsure about naming. Boards will probably also want to provide lists of extensions that they support at some point, which will be assembly flags. We also treat defines separately. Are there any cases where we may want to pass different flags for C and C++? Currently, xmake uses cflags for C flags, cxxflags for C++ flags, and cxflags for C/C++ flags. It might make sense to use cxflags here for consistency?

I don't think there are any cases today where we'd want to pass them differently, but there could conceivably be in the future. I'll rename to cxflags.

sdk/boards/sonata-0.2.json
resistor: You had previously said it was needed on 0.2 here: https://github.com/CHERIoT-Platform/cheriot-sail/issues/85#issuecomment-2517482580

I'm happy to do whatever, but I don't know how to find the correct answer.

resistor:
Updated naming and removed the flags from Sonata 0.2. Can you confirm that no other boards need the change?

davidchisnall (approved):

PR #400, opened 2025-01-06
Adapt STACK_CHECK macros to use the new __builtin_cheriot_get_specified_minimum_stack() when available.

resistor:


sdk/include/debug.hh
```diff
@@ -819,20 +819,20 @@ namespace
...
-	template<StackCheckMode Mode, size_t Expected, DebugContext Fn>
+	template<StackCheckMode Mode, DebugContext Fn>
 	class StackUsageCheck
 	{
 		/**
 		 * The expected maximum.  This class is templated on the function name
 		 * and so there is one copy of this per function.
 		 */
-		static inline size_t stackUsage = Expected;
+		static inline size_t stackUsage;
```
nwf, line 829:
```suggestion
		size_t stackUsage;
```

The change from a template parameter to a constructor parameter is... incomplete?  On both the "old" llvm-13 and new compiler, as it stands, this PR...

1. creates and *mutates* a `static` object in globals, and
2. stores a pointer to that static object on the stack, which is why CI is balking.

Here's the prolog of `heap_quota_remaining`, before (`-`) and after (`+`):

```
 800039b0 <heap_quota_remaining(SObjStruct*)>:
 ; heap_quota_remaining(SObjStruct*)():
 ; /home/cheriot/cheriot-rtos/tests/../sdk/core/allocator/main.cc:844
 ; {
-800039b0: 79 71                cincoffset      csp, csp, -48
-800039b2: 06 f4                csc     cra, 40(csp)
-800039b4: 22 f0                csc     cs0, 32(csp)
-800039b6: 26 ec                csc     cs1, 24(csp)
+800039b0: 39 71                cincoffset      csp, csp, -64
+800039b2: 06 fc                csc     cra, 56(csp)
+800039b4: 22 f8                csc     cs0, 48(csp)
+800039b6: 26 f4                csc     cs1, 40(csp)
 800039b8: db 04 a5 fe          cmove   cs1, ca0
-800039bc: 28 00                cincoffset      ca0, csp, 8
-800039be: 5b 24 05 01          csetbounds      cs0, ca0, 16
 
-800039c2 <.LBB0_6>:
+800039bc <.LBB0_6>:
 ; .LBB0_6():
+; /home/cheriot/cheriot-rtos/sdk/include/debug.hh:835
+;              constexpr inline explicit StackUsageCheck(size_t Expected) { stackUsage = Expected; }
+800039bc: db 95 c1 ff          cincoffset      ca1, cgp, -4
+800039c0: 13 05 00 09          addi    a0, zero, 144
+800039c4: 2e e4                csc     ca1, 8(csp)
+800039c6: 88 c1                csw     a0, 0(ca1)
```

Dropping the `static inline` on `stackUsage` undoes that change, presumably because the compiler has better visibility into all references to it.  The new compiler still won't pass CI yet until we bump some limits, but it is definitely _worse_ with the `static inline` on `stackUsage` than without.

resistor: The original code also had it as a `static inline`, and mutated the value in the destructor, so that's not a change per se.

I think the right fix here is to make the builtin constexpr, but that will require some investigation.

PR #401, opened 2025-01-06
allocator: correct token alloc error path

nwf:


sdk/core/allocator/main.cc
```diff
@@ -1183,7 +1190,11 @@ __cheriot_minimum_stack(0x280) SObj
 			return sealed;
```
davidchisnall, line 1190:
```suggestion
```

sdk/core/allocator/main.cc
```diff
@@ -1183,7 +1190,11 @@ __cheriot_minimum_stack(0x280) SObj
...
 		}
 	}
-	heap_free(heapCapability, obj);
+	/*
+	 * We get here only if we have failed to store the unsealed handle through
+	 * the out parameter.  Destroy the allocation we just made.
+	 */
+	token_obj_destroy(heapCapability, key, sealed);
 	return INVALID_SOBJ;
```
davidchisnall, line 1198:
```suggestion
return sealed;
```

nwf:
The SAFE revoker bug strikes again.  Leave CI green by leaving the explicit test of this error path for a future PR.

sdk/core/allocator/main.cc
```diff
@@ -1175,16 +1175,24 @@ __cheriot_minimum_stack(0x280) SObj
...
 		      Permission::Store, Permission::LoadStoreCapability}>(unsealed))
 		{
 			*unsealed = obj;
-			return sealed;
 		}
 	}
-	heap_free(heapCapability, obj);
-	return INVALID_SOBJ;
+	/*
+	 * We get here only if we have failed to store the unsealed handle through
```
rmn30, line 1192: What am I missing? This doesn't seem right.

nwf: Whoops, thanks.  Comment was from back when there was still a `return sealed` up above.

rmn30 (approved):
Looks reasonable to me.

PR #404, opened 2025-01-07
cxxrt guard: stop using cassert, check futex_wake result

nwf:


sdk/lib/cxxrt/guard.cc
```diff
@@ -1,12 +1,14 @@
...
 // SPDX-License-Identifier: MIT
 
-#include <cassert>
 #include <cdefs.h>
+#include <debug.hh>
 #include <futex.h>
 #include <limits>
 #include <stdint.h>
 
+using Debug = ConditionalDebug<DEBUG_CXXRT, "cxxrt">;
```
davidchisnall, line 10: I don’t see the xmake option that sets this, so I’m a bit confused about how it compiles at all.

sdk/lib/cxxrt/guard.cc
```diff
@@ -54,6 +56,8 @@ namespace
 
 		/**
 		 * Acquire the lock.
+		 *
+		 * This is safe only in IRQ-deferred context.
```
davidchisnall, line 60: Why?

sdk/lib/cxxrt/guard.cc
```diff
@@ -62,7 +66,7 @@ namespace
 			{
 				futex_wait(&high, LockBit);
 			}
-			assert(high == 0);
+			Debug::Assert(high == 0, "Corrupt guard word");
```
davidchisnall, line 69: It would be nice to log the address of the guard word here, so someone can look up that address. It will have a symbol name that matches the global being described.

sdk/lib/cxxrt/guard.cc
```diff
@@ -71,9 +75,10 @@ namespace
 		 */
 		void unlock()
 		{
-			assert(high == LockBit);
-			high = 0;
-			futex_wake(&high, std::numeric_limits<uint32_t>::max());
+			Debug::Assert(high == LockBit, "Corrupt guard word");
+			high    = 0;
+			int res = futex_wake(&high, std::numeric_limits<uint32_t>::max());
+			Debug::Invariant(res == 0, "futex_wake failed; possible deadlock");
```
davidchisnall, line 81: I’m not sure I like having this as an invariant. If a compartment is single threaded or otherwise ensures that statics are not concurrently initialised, it’s fine for this to fail. In this function, we don’t know if that is the case.

sdk/lib/cxxrt/guard.cc
nwf: It already existed so I didn't have to add it in this PR.

sdk/lib/cxxrt/guard.cc
nwf: A) whoops, that should be `>= 0`, not `== 0`

B) `futex_wake` fails (`< 0`) only if the cap is invalid or there's not enough room on the (trusted or untrusted) stack for the cross-call.

sdk/lib/cxxrt/guard.cc
davidchisnall: Oh, hmm.  This should be an atomic, I guess this predated support for atomics.  Unless there's a header somewhere defining these functions, it *isn't* called with interrupts disabled, so that would be a bug.

sdk/lib/cxxrt/guard.cc
davidchisnall: Right, case B is the one I'm thinking about.  It's fine for the futex wake to fail if there are no waiters.  We don't use a waiters flag, so we can't tell if that's the case.  In a lot of uses, this is used for lazy initialisation, but not for locking, so the failure is benign then.

sdk/lib/cxxrt/guard.cc
nwf: The callers are given prototypes that mark them as IRQ deferring with `DEFINE_ATOMIC_LIBCALL` at the top of the file.

davidchisnall: Ah, right.  I wonder if we should (after this PR is merged) refactor this whole thing to use priority inheriting flag locks.  The lock half of the word is entirely ours to play with.

sdk/lib/cxxrt/guard.cc
nwf: I suppose, but I think I'd rather trip an assertion in initialization than quietly fail to wake a thread after the system's flying?

davidchisnall: If it’s an assert, that’s fine, you can turn it off. If it’s an invariant, you get a crash in your compartment instead of working code.

davidchisnall (approved):

PR #405, opened 2025-01-08
Use the new __cheriot_minimum_stack__ when performing stack checks.

resistor:


rmn30 (approved):
Nice! What's the oldest compiler version we aim to support at the moment? Looks like `__CHERIOT__` was only added last month so to be kind to be people using older compilers we could check if it is defined?

sdk/core/allocator/alloc_config.h
```diff
@@ -34,5 +34,15 @@ constexpr StackCheckMode StackMode =
 #endif
   ;
 
-#define STACK_CHECK(expected)                                                  \
-	StackUsageCheck<StackMode, expected, __PRETTY_FUNCTION__> stackCheck
+#if __CHERIOT__ >= 20250108
```
rmn30, line 37:
```suggestion
#if defined(__CHERIOT__) && (__CHERIOT__ >= 20250108)
```

resistor:
> Nice! What's the oldest compiler version we aim to support at the moment? Looks like `__CHERIOT__` was only added last month so to be kind to be people using older compilers we could check if it is defined?

It looks like @davidchisnall added it in October. Is that recent enough that we need to worry about it?

davidchisnall:
There’s a ‘your compiler is too old’ warning somewhere in cdefs.h. 

I wanted to add a wrapper macro that took a version and evaluated to true if the compiler defined at least that version, false if it didn’t and unconditionally false for the compiler that didn’t define the macro. We should probably do that.

Until we do 1.0, we *can* drop support for older compilers, but I would like to avoid it unless there’s a really good reason.

resistor:
Can someone click the merge button?

PR #409, opened 2025-01-08
treewide: clang format and tidy for the new compiler

nwf:
This gets CI closer to passing, but I am not sure what to do about

> sdk/include/debug.hh:419:3: warning: Address of stack memory associated with local variable 'args' is still referred to by the stack variable 'arguments' upon returning to the caller.  This will be a dangling reference [clang-analyzer-core.StackAddressEscape]


davidchisnall:
I also missed in review of this one that the license doesn't match the rest of the project.  Is it possible to relicense?

PR #412, opened 2025-01-10
Remove sched namespace in scheduler

rmn30:
- **Fix scheduler compartment name used in DECLARE_INTERRUPT_CAPABILITY.**
- **Remove 'sched' namespace in scheduler.**

### DIFF
--- a/sdk/core/scheduler/main.cc
+++ b/sdk/core/scheduler/main.cc
@@ -188,10 +188,6 @@ namespace
 		return {shouldRecalculatePriorityBoost, woke};
 	}
 
-} // namespace
-
-namespace sched
-{
 	using namespace priv;
 
 	/**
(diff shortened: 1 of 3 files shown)


sdk/core/scheduler/main.cc
```diff
@@ -170,243 +170,235 @@ namespace
 
 } // namespace
 
-namespace sched
```
davidchisnall, line 173: These things should all be moved to an anonymous namespace, not exported functions in the global namespace.

Ideally, the `priv` namespace should also be removed.

rmn30: OK, I've done the first thing. Had to move exception_entry and scheduler_entry out of the anonymous namespace. The `priv` namespace should be a separate PR, I think.

davidchisnall: I don't think you pushed?  It is still putting everything in the global namespace.  Making it an anonymous namespace should remove most of the whitespace changes.

davidchisnall (approved):

PR #413, opened 2025-01-10
Improve futex_wake behaviour.

davidchisnall:
The current code yields and schedules a new thread if any equal or higher-priority thread has become runnable.

The new code decomposes these cases.  If a new thread is runnable and should run now, the current thread yields.  If a new thread is runnable but is the same priority as the current thread, the scheduler ensures that a timer interrupt is ready but does not yield immediately.

This improves the test suite performance by around 2% and a producer-consumer microbenchmark by about 35%.

### DIFF
--- a/sdk/core/scheduler/main.cc
+++ b/sdk/core/scheduler/main.cc
@@ -14,7 +14,6 @@
 #include <futex.h>
 #include <interrupt.h>
 #include <locks.hh>
-#include <new>
 #include <priv/riscv.h>
 #include <riscvreg.h>
 #include <simulator.h>
@@ -122,6 +121,26 @@ namespace
 	 */
 	static constexpr auto UnboundedSleep = std::numeric_limits<uint32_t>::max();
 
+	enum FutexWakeKind
+	{
+		/**
+		 * The futex wake did not make any threads runnable that would be
+		 * scheduled preemptively.
+		 */
+		NoYield,
+		/**
+		 * The futex wake made a thread at the current priority level runnable,
+		 * the caller should ensure that there is a timer interrupt scheduled
+		 * to make the current thread yield later.
+		 */
+		YieldLater,
+		/**
+		 * The futex wake made a thread at a higher priority level runnable,
+		 * the caller should yield to allow the other thread to run immediately.
+		 */
+		YieldNow,
+	};
+
 	/**
 	 * Helper that wakes a set of up to `count` threads waiting on the futex
 	 * whose address is given by the `key` parameter.
@@ -134,11 +153,10 @@ namespace
 	 *    dropped back to the previous priority.
 	 *  - The number of sleeper that were awoken.
 	 */
-	std::tuple<bool, bool, int>
+	std::tuple<bool, int>
 	futex_wake(ptraddr_t key,
 	           uint32_t  count = std::numeric_limits<uint32_t>::max())
 	{
-		bool shouldYield                    = false;
 		bool shouldRecalculatePriorityBoost = false;
 		// The number of threads that we've woken, this is the return value on
 		// success.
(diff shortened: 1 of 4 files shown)


rmn30 (approved):
Looks reasonable, but could do with review by @nwf too.

sdk/core/scheduler/main.cc
```diff
@@ -530,7 +553,28 @@ __cheriot_minimum_stack(0xa0) int futex_wake(uint32_t *address, uint32_t count)
...
-	auto [shouldYield, shouldResetPrioirity, woke] = futex_wake(key, count);
+	auto [shouldResetPrioirity, woke] = futex_wake(key, count);
+
+	FutexWakeKind shouldYield = NoYield;
+
+	if (woke > 0)
+	{
+		auto *thread = Thread::current_get();
+		if (!thread)
+		{
```
nwf, line 564: Just to confirm my understanding: this is the case of `exception_entry` calling `futex_wake`?

sdk/core/scheduler/main.cc
```diff
@@ -551,12 +595,19 @@ __cheriot_minimum_stack(0xa0) int futex_wake(uint32_t *address, uint32_t count)
...
-		yield();
+		case YieldLater:
+			Timer::ensure_tick();
+			break;
+		case YieldNow:
+			yield();
+			break;
+		case NoYield:
+		{
+		}
```
nwf, line 610: Purely a stylistic question: why not `break;`?

sdk/core/scheduler/thread.h
```diff
@@ -543,6 +539,17 @@ namespace
...
 		}
 
+		/**
+		 * Returns true if the thread has run for a complete tick.  This must
+		 * be called only on the currently running thread.
+		 */
+		bool has_run_for_full_tick()
+		{
+			Debug::Assert(this == current,
+			              "Only the current thread is running");
```
nwf, line 549: That's a bit of a tautology, isn't it?

```suggestion
			              "Ask only about currently running thread");
```

sdk/core/scheduler/thread.h
```diff
@@ -564,8 +571,12 @@ namespace
 		///@}
 		/// Pointer to the list of the resource this thread is blocked on.
 		ThreadImpl **sleepQueue;
-		/// If suspended, when will this thread expire. The maximum value is
-		/// special-cased to mean blocked indefinitely.
+		/**
+		 * If suspended, when will this thread expire. The maximum value is
+		 * special-cased to mean blocked indefinitely.
+		 *
+		 * When a thread is running, this the time at which it was scheduled.
```
nwf, line 578: Again, just to check my understanding: Time of the _last_ scheduling event and not the time of any _upcoming_ scheduling event because there might not actually be one, if this is the only runnable thread in the system, and besides it's cheaper to not do the addition unless someone asks in `has_run...`?

nwf (commented):
LGTM, modulo nits and some questions to check my understanding.

sdk/core/scheduler/main.cc
davidchisnall: Also fine, I just wanted to make sure all cases were covered.  Break is probably better.

sdk/core/scheduler/thread.h
davidchisnall: Yes, this when you were scheduled.  When you do futex wake, you may have been running for more then one tick, or less than one tick.

sdk/core/scheduler/thread.h
rmn30: Wes' version is verging on a parody of an ancient Chinese wisdom. "Ask not a suspended thread how long it has been running for."

davidchisnall: May you be scheduled at interesting times.

PR #414, opened 2025-01-13
Further fixes for enabling warnings on cross-compartment void returns.

resistor:


sdk/lib/cxxrt/guard.cc
```diff
@@ -73,7 +73,7 @@ namespace
 		{
 			assert(high == LockBit);
 			high = 0;
-			futex_wake(&high, std::numeric_limits<uint32_t>::max());
+			(void)futex_wake(&high, std::numeric_limits<uint32_t>::max());
```
nwf, line 76: This is https://github.com/CHERIoT-Platform/cheriot-rtos/pull/404

nwf (commented):
Thanks for doing this; sorry to have only done so for the bits that fed into the test suite.

I wonder, though... a lot of the `return -1`-s that this introduces are effectively at top-level and silently lead to thread exit.  I'd ask if we could make them more vocal, but a bunch of them are "we can't call into the uart compartment" kinds of failure, so that kind of vocalization is unavailable... perhaps `Debug::Invariant`, so at least there's a trap?

resistor:
> Thanks for doing this; sorry to have only done so for the bits that fed into the test suite.

No worries. You got all the hard bits.

> I wonder, though... a lot of the `return -1`-s that this introduces are effectively at top-level and silently lead to thread exit. I'd ask if we could make them more vocal, but a bunch of them are "we can't call into the uart compartment" kinds of failure, so that kind of vocalization is unavailable... perhaps `Debug::Invariant`, so at least there's a trap?

I'll look into this.

resistor:
> Thanks for doing this; sorry to have only done so for the bits that fed into the test suite.
> 
> I wonder, though... a lot of the `return -1`-s that this introduces are effectively at top-level and silently lead to thread exit. I'd ask if we could make them more vocal, but a bunch of them are "we can't call into the uart compartment" kinds of failure, so that kind of vocalization is unavailable... perhaps `Debug::Invariant`, so at least there's a trap?

Done. PTAL.

examples/07.error_handling/hello.cc
```diff
@@ -2,18 +2,20 @@
...
 #include <fail-simulator-on-error.h>
 
 /// Thread entry point.
-void __cheri_compartment("hello") entry()
+int __cheri_compartment("hello") entry()
 {
 	// Try writing a string with a missing null terminator
 	char maliciousString[] = {'h', 'e', 'l', 'l', 'o'};
-	write(maliciousString);
+	(void)write(maliciousString);
```
nwf, line 13: Should this check for `ECOMPARTMENTFAIL`?

benchmarks/allocation/alloc.cc
```diff
@@ -34,7 +35,10 @@ void __cheri_compartment("allocbench") run()
 		auto quota = heap_quota_remaining(MALLOC_CAPABILITY);
 		Debug::Invariant(quota == MALLOC_QUOTA, "Quota remaining {}, should be {}", quota, MALLOC_QUOTA);
 		Debug::log("Flushing quarantine");
-		heap_quarantine_empty();
+		Debug::Invariant(heap_quarantine_empty() >= 0,
```
nwf, line 38: I think in _benchmark_ code, these might want to be `Debug::Assert`, so that they compile away in release mode when we're measuring cycles?

benchmarks/allocation/alloc.cc
```diff
@@ -34,7 +35,10 @@ void __cheri_compartment("allocbench") run()
...
+		                 "Unable to call heap_quarantine_empty");
```
nwf, line 39: Hm.  I worry that "Unable to call" is potentially misleading... In many cases, the callees indeed cannot fail, so it's accurate, but in the case where they _can_ fail?

nwf (changes requested):
Thanks!  I like that a lot better.  Some comments below.

benchmarks/allocation/alloc.cc
resistor: This doesn't work out super well, because the call won't evaluated when the assert is compiled out.

examples/07.error_handling/hello.cc
resistor: Done throughout

benchmarks/allocation/alloc.cc
resistor: Made the message clearer throughout

nwf (approved):

benchmarks/allocation/alloc.cc
```diff
@@ -9,12 +9,13 @@ using Debug = ConditionalDebug<DEBUG_ALLOCBENCH, "Allocator benchmark">;
...
  * bytes, report how long it takes.
  */
-void __cheri_compartment("allocbench") run()
+int __cheri_compartment("allocbench") run()
 {
 	// Make sure sail doesn't print annoying log messages in the middle of the
 	// output the first time that allocation happens.
 	free(malloc(16));
-	heap_quarantine_empty();
+	Debug::Invariant(heap_quarantine_empty() != -ECOMPARTMENTFAIL,
```
rmn30, line 17:
```suggestion
	Debug::Invariant(heap_quarantine_empty() == 0,
```
This invariant is not strong enough as, according to the comment on `heap_quarantine_empty`, other failures such as `ENOTENOUGHSTACK` and `ENOTENOUGHTRUSTEDSTACK` are possible. This applies to all other calls to this and perhaps other compartment calls?

tests/allocator-test.cc
```diff
@@ -271,7 +272,8 @@ namespace
 		 * failure we see below isn't because we aren't allowing the revocation
 		 * state machine to advance.
 		 */
-		heap_quarantine_empty();
+		TEST(heap_quarantine_empty() != -ECOMPARTMENTFAIL,
+		     "Compartment call to heap_quarantine_empty failed");
```
rmn30, line 276:
```suggestion
		TEST_EQUAL(heap_quarantine_empty(), 0, "Compartment call to heap_quarantine_empty failed");
```
As per previous comment but in the test suite we have this handy macro.

rmn30 (changes requested):

benchmarks/allocation/alloc.cc
rmn30: I think we'd like to know about any failures so would prefer the stronger invariant.

rmn30 (approved):

resistor:
@rmn30 @nwf Can one of you hit merge on this for me?

PR #419, opened 2025-01-14
Fix multiwaiter public header.

davidchisnall:
It had a load of left-over things in it from when the scheduler knew about event groups.

### DIFF
--- a/sdk/include/multiwaiter.h
+++ b/sdk/include/multiwaiter.h
@@ -34,54 +34,24 @@
 #include <timeout.h>
 
 /**
- * The kind of event source to wait for in a multiwaiter.
- */
-enum EventWaiterKind
-{
-	/// Event source is an event channel.
-	EventWaiterEventChannel,
-	/// Event source is a futex.
-	EventWaiterFutex
-};
-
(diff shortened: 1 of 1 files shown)


nwf (approved):
With the test caught up, it LGTM.  Please squash, not rebase, merge for ease of any future bisections.

nwf:
Mm, on second thought, there's a bit more that can go along these lines...

nwf:
(There's a bit of cruft, like https://github.com/CHERIoT-Platform/cheriot-rtos/blob/bea973dae848c109147967ce37f5a0ad1738382d/sdk/core/scheduler/multiwait.h#L36 and the fact that `::readyEvents` is used as a 24-bit boolean, that could also be cleaned up, or left in case we ever wanted something like it again.  I think it's not enough cruft that removing it would save bytes in an `EventWaiter` object?)

davidchisnall:
I think all of the responses are now a single bit (event fired or did not fire) so we might be able to pack them into a bitfield in the header rather than a per-event field. These things make some paths O(n) so a limit of well under 32 events per waiter is fine.

PR #422, opened 2025-01-15
allocator-test: probe revocation in more places

nwf:
Explicitly test for revocation in all of globals, the stack, and the trusted stack.


nwf:
This also needs a test of capabilities in static shared objects.

nwf (commented):
I cannot approve my own PR, but thank you for seeing the work through to the end.  This LGTM.

davidchisnall (approved):

davidchisnall (approved):

PR #423, opened 2025-01-16
Add some limited JSON Patch support for board description files.

davidchisnall:
Boards can now be provided as .patch files that are JSON documents containing two nodes:

 - `base` is the board that this is based on, which can be another .patch file, a full path (note: relative paths are not yet supported, but they should be for out-of-tree BSPs).
 - `patch` is a JSON Patch.

The patch is a subset of RFC 6902.  Current limitations:

 - It doesn't handle escapes in keys or any weird keys.
 - It assumes that numbers in the JSON Pointer path are array indexes and will do the wrong thing if you try to replace or add an object field with a number as a key.
 - It doesn't implement the copy, move, or test operations, only add, remove, and replace.

This is sufficient to be useful.  The Arty A7 and Sonata Simulator JSON files are now replaced with patches.


sdk/xmake.lua
```diff
@@ -253,21 +253,143 @@ target("cheriot.software_revoker")
...
+	if not patch.base then
+		print("Board file " .. boardFile .. " does not specify a base")
+		return nil
+	end
+	local _, baseFile = board_file_for_name(patch.base)
+	local base = load_board_file(json, baseFile)
+
+	-- If a string value is a number, return it as number, otherwise return it
+	-- in its original form.
+	function asNumberIfNumber(value)
```
nwf, line 308: Hrm.  This function isn't just `tonumber(value) or value` (numeric 0 is _true_ in Lua, and `tonumber` won't return `false`) because we want to keep things Lua can't exactly represent as strings?  I don't think we're likely to _notice_ the difference in practice, but defense in depth is probably a good idea.

sdk/xmake.lua
```diff
@@ -253,21 +253,143 @@ target("cheriot.software_revoker")
...
+			return nil
+		end
+
+		-- Parse the JSON Pointer into an array of filed names, converting
+		-- numbers into Lua numbers if we see them.  This is not quite right,
+		-- because it doesn't handle field names with / in them, but we don't
+		-- currently use those for anything.  It also assumes that we really do
+		-- mean array indexes when we say numbers.  If we have an object with
+		-- "3" as the key and try to replace 3, it will currently not do the
+		-- right thing.
```
nwf, line 345: The bit about numeric path components might be somewhat easy to solve with `is_marked_as_array`?  Might not be worth doing yet, but just a comment with the name so we don't forget, perhaps.

sdk/xmake.lua
```diff
@@ -339,9 +460,10 @@ rule("firmware")
 			visit_all_dependencies_of(target, callback)
 		end
 
-		local boarddir, boardfile = board_file(target);
-		local board = json.loadfile(boardfile)
-
+		local boarddir, boardfile = board_file_for_target(target);
+		local board = load_board_file(json, boardfile)
+		print("Board file saved as ", target:targetfile()..".board.json")
+		json.savefile(target:targetfile()..".board.json", board)
```
nwf, line 466: Oh thank you.  It's so obnoxious when build systems hide their computed state. :)

sdk/xmake.lua
davidchisnall: I’m not sure I understand. JSON keys are strings for objects or numbers for dictionaries. Will this do the wrong thing if a key is called “true”?

sdk/xmake.lua
```diff
@@ -253,21 +253,143 @@ target("cheriot.software_revoker")
...
+		end
+		return value
+	end
+
+	-- Heuristic to tell a Lua table is probably an array in Lua
+	-- This is O(n), but n is usually very small, and this happens once per
+	-- build so this doesn't really matter.
+	function isarray(t)
+		local i = 1
+		for k, v in pairs(t) do
```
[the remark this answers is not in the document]
davidchisnall: Aha! It’s not documented, I expected something like that to exist. With that, we can just do the coercion of keys to numbers when the thing is an array, so that simplifies it a lot.

sdk/xmake.lua
davidchisnall: It’s mostly there because I didn’t want to rewrite all of this logic for cheriot-audit (though, at least, the JSON library that it uses has JSON Patch support). This also means that we can use any JSON extensions that xmake supports and not worry because other tools can just use the output (which is plain.simple JSON).

sdk/xmake.lua
nwf: Lua's a strange beast.  `tonumber("true") or "true"` is the string `"true"`, not the boolean `true`, because `nil or X` evaluates to `X` unchanged, and `X or Y` with `X` neither `nil` or `false` evaluates to `X`.

davidchisnall: That is the desired behaviour. Booleans do not want this kind of special casing, only numbers (technically, only integers). But given the ability to accurately tell if a JSON thing is an array, I don’t think this function is needed.

sdk/xmake.lua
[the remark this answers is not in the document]
davidchisnall: It seemed like such a good idea.  Unfortunately, it doesn't seem to work on my xmake build (I'm guessing it's using the built-in JSON parser rather than json-c and there's a bug).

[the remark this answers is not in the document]
davidchisnall: I rewrote it to use this function if it works and use a fallback if not.  The fallback is correct, though potentially O(n) in terms of the underlying tree, rather than just the top-level array: it will convert the entire tree to JSON and then see if the first char is '['.  If that function works, it's O(1).

Even with the fallback, execution time is negligible with the current examples in the tree, so I don't care too much.

sdk/boards/sonata-simulator.patch
```diff
@@ -0,0 +1,15 @@
...
+  "base": "sonata-prerelease",
+  "patch": [
+    {
+      "op": "replace",
+      "path": "/simulation",
+      "value": true
+    },
+    {
+      "op": "replace",
+      "path": "/simulator",
```
rmn30, line 11: Shouldn't there be a run here somewhere?

PR #424, opened 2025-01-17
s/simulator/run_command/

davidchisnall:
Having two board description nodes at the same level, one called simulator and one called simulation is error prone and will cause people to be confused.

The simulator node is also now poorly named.  On Sonata, for example, it doesn't run a simulation, it loads onto the board.  The same will likely happen with ASICs.

Given that it's both ambiguous and just plain wrong, replace it with run_command.

The xmake still works with simulator, if someone has an out-of-tree board file.  This will go away at some point.

### DIFF
--- a/sdk/boards/ibex-arty-a7-100.patch
+++ b/sdk/boards/ibex-arty-a7-100.patch
@@ -71,7 +71,7 @@
     },
     {
       "op": "remove",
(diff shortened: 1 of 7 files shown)


rmn30 (approved):
As someone who was temporarily confused by this I approve.

PR #425, opened 2025-01-20
Add hyperram support on Sonata v1.

rmn30:
We place all the rw data in SRAM and (almost) all code and RO data in HyperRAM. This requires changing the linker script to support linking code either at lower (on most current boards) or higher address (Sonata v1).


nwf (changes requested):

scripts/model_output/sonata-simulator/examples/test-suite.txt
```diff
@@ -0,0 +1 @@
+All tests finished
```
davidchisnall, line 1: This file won't be needed after rebase on #428

davidchisnall (approved):
Looks good once #428 is done.

nwf (approved):

PR #427, opened 2025-01-22
Adopt compiler support for sealed capabilities

davidchisnall:
This still supports the old mode and, because some existing code may require use of SObj, provides a compatibility mode for migration.

With the sealed branch of the compiler, there is a __sealed_capability qualifier.  This qualifier is part of the type system and sealed capabilities cannot be dereferenced (the hardware enforces this, now the compiler does as well).  The seal and unseal builtins are perform the conversion between a `T*` and a `T*__sealed_capability`.

### DIFF
--- a/sdk/core/loader/debug.hh
+++ b/sdk/core/loader/debug.hh
@@ -132,8 +132,8 @@ namespace
 		/**
 		 * Append a capability.
 		 */
-		template<typename T>
-		__always_inline void append(CHERI::Capability<T> capability)
+		template<typename T, bool IsSealed>
+		__always_inline void append(CHERI::Capability<T, IsSealed> capability)
 		{
 			append(static_cast<const void *>(capability.get()));
 		}
--- a/sdk/include/cheri.hh
+++ b/sdk/include/cheri.hh
@@ -384,15 +384,104 @@ namespace CHERI
 		       representable_length(size) == size;
 	}
 
+#if __has_extension(cheri_sealed_pointers)
+	// clang-format treats __sealed_capability differently depending on whether
+	// it knows it's a type qualifier or thinks it's an identifier.  Turn off
+	// clang-format here until we've bumped the version in CI.
+	// clang-format off
+	/**
+	 * Type trait to remove sealed from a capability type.
+	 */
+	template<typename T>
+	struct remove_sealed; // NOLINT(readability-identifier-naming)
+
+	/**
+	 * Specialisation to
+	 */
+	template<typename T>
+	struct remove_sealed<T *__sealed_capability>
+	{
+		using type = T *; // NOLINT(readability-identifier-naming)
+	};
+
+	/**
+	 * Type trait for checking if a type is a sealed capability.
+	 */
+	template<typename T>
+	struct is_sealed_capability // NOLINT(readability-identifier-naming)
+	{
+		static constexpr bool value = // NOLINT(readability-identifier-naming)
+		  false;
+	};
+
+	/**
+	 * Type trait for checking if a type is a sealed capability.
+	 */
+	template<typename T>
+	struct is_sealed_capability<T *__sealed_capability>
+	{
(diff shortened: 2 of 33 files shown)


sdk/include/cheri.hh
```diff
@@ -1055,27 +1119,32 @@ namespace CHERI
...
-			return *this;
+			return {reinterpret_cast<CHERI_SEALED(T *)>(
+			  __builtin_cheri_seal(ptr, key))};
 		}
 
 		/**
 		 * Unseal this capability with the given key.
 		 */
-		Capability<T> &unseal(void *key)
+		[[nodiscard]] Capability<T, false> unseal(void *key)
```
nwf, line 1139: Do you want to `requires IsSealed` here?

sdk/include/cheri.hh
```diff
@@ -1055,27 +1119,32 @@ namespace CHERI
...
 		{
-			ptr = static_cast<T *>(__builtin_cheri_unseal(ptr, key));
-			return *this;
+#if __has_extension(cheri_sealed_pointers) &&                                  \
+  defined(CHERIOT_NO_SEALED_POINTERS)
```
nwf, line 1142: Do you mean `!defined` here as everywhere else?  I don't see `CHERIOT_NO_SEALED_POINTERS` documented anywhere, so I'm not quite sure what it's supposed to be doing...

sdk/core/loader/debug.hh
```diff
@@ -132,8 +132,8 @@ namespace
 		/**
 		 * Append a capability.
 		 */
-		template<typename T>
-		__always_inline void append(CHERI::Capability<T> capability)
+		template<typename T, bool IsSealed>
+		__always_inline void append(CHERI::Capability<T, IsSealed> capability)
```
nwf, line 136: I wonder if it's worth a check and warning if the sealed state of the capability differs from the static annotation?

sdk/include/cheri.hh
davidchisnall: `CHERIOT_NO_SEALED_POINTERS` is for opting out of the typed sealed pointer things so that existing code that uses `SObj` should work.  This one is the other way around because it's specifically for using a compiler that *does* support sealing in the type system, which changes the type signature of the seal and unseal builtins.

sdk/core/loader/debug.hh
davidchisnall: That would need to be in the debug library and would need us to use a different type to pass in sealed things, but is probably a good idea.

sdk/include/cheri.hh
```diff
@@ -1055,27 +1119,32 @@ namespace CHERI
...
 		/**
 		 * Seal this capability with the given key.
 		 */
-		Capability<T> &seal(void *key)
+		[[nodiscard]] Capability<T, true> seal(void *key)
 		{
-			ptr = static_cast<T *>(__builtin_cheri_seal(ptr, key));
-			return *this;
+			return {reinterpret_cast<CHERI_SEALED(T *)>(
+			  __builtin_cheri_seal(ptr, key))};
```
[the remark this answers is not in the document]
davidchisnall: This won't be merged without a blog explaining the changes.

sdk/include/stdlib.h
```diff
@@ -40,7 +41,10 @@ struct AllocatorCapabilityState
 	uintptr_t reserved[2];
 };
 
-struct SObjStruct;
+/**
+ * Type for allocator capabilities.
+ */
+typedef CHERI_SEALED(struct AllocatorCapabilityState *) AllocatorCapability;
```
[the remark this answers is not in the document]
davidchisnall: In the old version, everything was a `SObj` and no one got around to adding fake types.  It would have been a good idea though.

We'll need to make the same changes in the network stack, which also uses a bunch of software-defined capabilities.

PR #428, opened 2025-01-22
Use magic string to exit on sonata-simulator.

rmn30:
This means we no longer need the brittle and broken model output files and can simplify the run-sonata-sim script considerably.
The testing-model-output xmake option is also redundant.


nwf (approved):

sdk/include/platform/sunburst/platform-simulation_exit.hh
```diff
@@ -0,0 +1,25 @@
...
+#	include <platform-uart.hh>
+#	include <string_view>
+
+static void platform_simulation_exit(uint32_t code)
+{
+#	if DEVICE_EXISTS(uart0)
+	auto uart = MMIO_CAPABILITY(Uart, uart0);
+#	elif DEVICE_EXISTS(uart)
+	auto uart = MMIO_CAPABILITY(Uart, uart);
+#	endif
```
davidchisnall, line 17: Our normal style for this is to move the `auto UART=` out of the ifdef block. This makes it clearer that the existence of the variable is not conditional.

sdk/include/platform/sunburst/platform-simulation_exit.hh
```diff
@@ -0,0 +1,25 @@
...
+	// The following magic string will cause sonata simulator to exit.
+	using namespace std::string_view_literals;
```
davidchisnall, line 19: I am not convinced that this is more readable (it’s definitely longer) than just using the long-form constructor.

rmn30: If I do `std::string_view s {"magic string"}` there is a call to `strlen` which means I have to link the string library into hello_world and any other example that doesn't already have it. Using a string_view literal (or just writing the C-style loop) avoids this.

scripts/run-sonata-sim.sh
```diff
@@ -25,27 +25,20 @@ fi
...
-		fi
-		LOOP_TRACKER=$((LOOP_TRACKER+1))
-	done
-	# Timeout was hit so no success.
-	pkill -P $$
+# Run the simulator with a timeout and capture the exit status. The || is needed so that set -e doesn't kill us if this fails
+# and we can report the error gracefully. The timeout is chosen so that the testsuite can run in CI. It's probably
+# be a bit long for other purposes.
+result=0
+timeout --foreground 20m ${SONATA_SIMULATOR} -E "${SONATA_SIMULATOR_BOOT_STUB}" -E "$1" || result=$?
```
davidchisnall, line 32: Please don't depend on non-POSIX things.  If we need to use timeout in CI, that's fine (it's a known environment) but requiring developers to install GPLv3 things is mean.

davidchisnall (changes requested):

sdk/include/platform/sunburst/platform-simulation_exit.hh
davidchisnall: Computers :-(

marnovandermaas:
> This means we no longer need the brittle and broken model output files and can simplify the run-sonata-sim script considerably. The testing-model-output xmake option is also redundant.

Why are you removing the model output? This allows for smoke testing the Sonata system until we can run the full test suite on Sonata with the HyperRAM

davidchisnall (approved):

davidchisnall:
@marnovandermaas This is half of a pair of PRs, the other one enables the test suite on Sonata, but we want to merge this one first or we need to add a file and then immediately remove it.

rmn30:
@marnovandermaas there were a couple of problems with the model output and the way it was being used:

1. It contained addresses which are liable to change often, making an exact match of the output very brittle.
2. The run-sonata-sim script used `grep -f` to check for a match which matches as any line of output matches any line in the output file, meaning it would kill the simulation after 1 second in most cases and the examples wouldn't finish executing.

I agree that having model output might be nice but it wasn't working properly as was. I've added a grep for `failure` in the output which should catch cases of assertion failure. I'm open to better solutions.

PR #434, opened 2025-01-24
Fix the LED numbering for Sonata 1.0 and later.

davidchisnall:


marnovandermaas (approved):
This looks correct to me from a code review. Note that we're planning to upstream a rewrite of the Sonata GPIO driver soon.

PR #435, opened 2025-01-27
Add sprintf to stdio.

rmn30:
We can use the length of the output capability to calculate the length to pass to vnsprintf, making sprintf safe! Also fixes an error in the prototype for vnsprintf.

### DIFF
--- a/sdk/include/stdio.h
+++ b/sdk/include/stdio.h
@@ -8,6 +8,7 @@
 #include <compartment-macros.h>
 #include <stdarg.h>
 #include <stddef.h>
+#include <cheri-builtins.h>
 
 #define PRT_MAX_SIZE (0x80)
 #define EOF (-1)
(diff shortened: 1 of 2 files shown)


tests/stdio-test.cc
```diff
@@ -1,4 +1,5 @@
 #include <cstdio>
+#include <cstring>
```
rmn30, line 2: My editor appears to have added this include automatically, presumably because of using strcmp. Doesn't appear to be required but also not doing any harm?

sdk/include/stdio.h
```diff
@@ -65,10 +66,19 @@ static inline int printf(const char *format, ...)
...
+int __cheri_libcall vsnprintf(char *str,
                               size_t      size,
                               const char *format,
                               va_list     ap);
+
+static inline int sprintf(char *str, const char *format, ...)
+{
+	va_list args;
+	va_start(args, format);
+	int ret = vsnprintf(str, cheri_top_get(str) - cheri_address_get(str), format, args);
```
rmn30, line 78: I thought about creating a helper in `cheri-builtins.h` for top - address e.g. `cheri_remaining_get`, would you like that?

tests/stdio-test.cc
```diff
@@ -19,5 +20,9 @@ int test_stdio()
 	TEST(strcmp(buffer, "-42") == 0,
 	     "snprintf(\"%d\", -42) gave {}",
 	     std::string_view{buffer, BufferSize});
+	sprintf(buffer, "%x", 6 * 9);
+	TEST(strcmp(buffer, "36") == 0,
+	     "sprintf(\"%x\", 6 * 9) gave {}",
+	     std::string_view{buffer, BufferSize});
```
davidchisnall, line 26: If you're going to create a string view anyway, you can just use operator == in the comparison rather than strcmp.

rmn30: OK, I used string_view and replaced TEST with TEST_EQUAL. How's that?

davidchisnall (approved):

PR #436, opened 2025-01-29
sonata: Added I2C devices to the board description

HU90m:


nwf (approved):
Seems OK, though I'm curious where the `0x80` comes from, since https://github.com/lowRISC/sonata-system/blob/main/data/top_config.toml doesn't mention it.

PR #437, opened 2025-01-29
Sonata v1.* GPIO driver

HU90m:


sdk/include/platform/sunburst/v0.2/platform-gpio.hh
```diff
@@ -0,0 +1,143 @@
...
+
+	/**
+	 * Is the current target Sonata 1.0.0 or later?
+	 */
+	static bool constexpr Sonata1OrLater =
+#if SONATA >= CHERIOT_VERSION_TRIPLE(1, 0, 0)
+	  true
+#else
+	  false
+#endif
```
davidchisnall, line 50: This probably isn't needed since this is just 0.2 now.

sdk/include/platform/sunburst/platform-gpio.hh
```diff
@@ -0,0 +1,444 @@
...
+/**
+ * Represents the state of Sonata's joystick, where each possible input
+ * corresponds to a given bit in the General GPIO's input register.
+ *
+ * Note that up to 3 of these bits may be asserted at any given time: pressing
+ * down the joystick whilst pushing it in a diagonal direction (i.e. 2 cardinal
+ * directions).
+ *
+ */
+enum class SonataJoystick : uint16_t
```
nwf, line 15: Should this be within the `struct SonataGpioBoard` rather than at top-level?

sdk/include/platform/sunburst/platform-gpio.hh
```diff
@@ -0,0 +1,444 @@
...
+	{
+		return (debouncedInput & gpio_bit(index, input_mask())) > 0;
+	}
+};
+
+/**
+ * A driver for Sonata's Board GPIO (instance 0).
+ *
+ * Documentation source:
+ * https://lowrisc.org/sonata-system/doc/ip/gpio.html
```
nwf, line 148: Perhaps examples in the documentation of the right `MMIO_CAPABILITY` invocation for these would be useful?  (Or even static factory methods, since they are backed by singletons in Sonata hardware?)

sdk/include/platform/sunburst/platform-gpio.hh
```diff
@@ -138,6 +292,153 @@ struct SonataGPIO
...
+ */
+struct SonataGpioRaspberryPiHat : SonataGpioBase
+{
+	/**
+	 * The mask of bits of the `output` register that contain meaningful GPIO
+	 * output.
+	 */
+	static constexpr uint32_t output_mask()
```
davidchisnall, line 311: I think this doesn't do what you think.  You probably want to either make this virtual (not ideal) or use CRTP so that the superclass can use it.

sdk/include/platform/sunburst/platform-gpio.hh
```diff
@@ -0,0 +1,444 @@
...
+};
+
+/**
+ * A simple driver for the Sonata's GPIO. This struct represents a single
+ * GPIO instance, and the methods available to interact with that GPIO.
+ *
+ * GPIO instances can be implemented via subclass structs which optionally
+ * override the three static mask attributes and/or add their own instance-
+ * specific functionality. The subclass itself should be provided as a
+ * template class parameter when declaring inheritance.
```
AlexJones0, line 31: Nit: I think I forgot to remove this last sentence from the comment when I removed the CRTP pattern from this driver; it doesn't make sense any more. The previous sentence should also probably say "three static mask ~attributes~ methods".

sdk/include/platform/sunburst/v0.2/platform-gpio.hh
HU90m: Done. I haven't removed `CHERIOT_VERSION_TRIPLE`. It isn't used anywhere else that I could see. Should I remove it?

HU90m:
Changed the driver to use template parameters for the masks.

sdk/include/platform/sunburst/platform-gpio.hh
HU90m: This documentation page is meant to be purely a hardware reference as oppose to how-to/tutorial. The page does have a lot of room for improvement. We could also create a page with some minimal how-to guides that the drivers could point to, but I don't think this should live in the sonata-system book.

nwf (commented):
Looks generally sensible to me.  I'll test it later today.

sdk/include/platform/sunburst/platform-gpio.hh
```diff
@@ -1,78 +1,186 @@
...
-	uint32_t debouncedThreshold;
-	uint32_t raspberryPiHeader;
-	uint32_t raspberryPiMask;
-	uint32_t arduinoShieldHeader;
-	uint32_t arduinoShieldMask;
+	uint32_t outputEnable;
+
+	/**
+	 * Returns the bit corresponding to a given GPIO index. Will mask out
+	 * the bit if this is outside of the provided mask.
```
rmn30, line 41:
```suggestion
	 * Returns a mask with a single bit set corresponding to the given GPIO index
	 * or 0 if that bit is not set in mask.
```

sdk/include/platform/sunburst/platform-gpio.hh
```diff
@@ -1,78 +1,186 @@
...
+	 */
+	constexpr static uint32_t gpio_bit(uint32_t index, uint32_t mask)
+	{
+		return (1 << index) & mask;
+	}
+
+	/**
+	 * Set the output bit for a given GPIO pin index to a specified value.
+	 * This will only have an effect if the corresponding bit is first set
+	 * to `0` (i.e. output) in the `output_enable` register, and if the pin
```
rmn30, line 51:
```suggestion
	 * to `1` (i.e. output) in the `output_enable` register, and if the pin
```
This appears to contradict the comment and code for `set_output_enable` below.

sdk/include/platform/sunburst/platform-gpio.hh
```diff
@@ -1,78 +1,186 @@
...
-	  false
-#endif
-	  ;
+	bool read_input(uint32_t index) volatile
+	{
+		return (input & gpio_bit(index, InputMask)) > 0;
+	}
+
+	/**
+	 * Read the debounced input value for a given GPIO pin index. FOr this
```
rmn30, line 98:
```suggestion
	 * Read the debounced input value for a given GPIO pin index. For this
```

sdk/include/platform/sunburst/platform-gpio.hh
rmn30: I think static factory method / function would make sense.

sdk/include/platform/sunburst/platform-gpio.hh
```diff
@@ -1,78 +1,186 @@
...
+	/**
+	 * Represents the state of Sonata's joystick, where each possible input
+	 * corresponds to a given bit in the General GPIO's input register.
+	 *
+	 * Note that up to 3 of these bits may be asserted at any given time:
+	 * pressing down the joystick whilst pushing it in a diagonal direction
+	 * (i.e. 2 cardinal directions).
+	 *
+	 */
+	enum class SonataJoystick : uint16_t
```
rmn30, line 127:
```suggestion
	enum class [[clang::flag_enum]] SonataJoystick : uint16_t
```
As this can have more than one value we should tell the compiler that. I would also prefer the name `SonataJoystickDirection` or similar. `SonataJoystick` sounds to me like the actual GPIO register, which this isn't.

rmn30: I would actually like `read_joystick` to return a `SonataJoystickValue` class that has a variable of this enum as a member and convenience methods like `is_pressed()`.

sdk/include/platform/sunburst/platform-gpio.hh
```diff
@@ -1,78 +1,186 @@
...
+  /**
+   * The mask of bits of the `output` register that contain meaningful
+   * GPIO output.
+   */
+  uint32_t OutputMask = 0xFFFF'FFFF,
+  /**
+   * The mask of bits of the `input` and `debouncedInput` registers that
+   * contain meaningful GPIO input.
+   */
+  uint32_t InputMask = OutputMask,
```
rmn30, line 26: Should this be negated?

sdk/include/platform/sunburst/v0.2/platform-gpio.hh
rmn30: I think it's fine to keep `CHERIOT_VERSION_TRIPLE` as we might need it for other things in future and we still define `SONATA` in that format.

sdk/include/platform/sunburst/platform-gpio.hh
```diff
@@ -138,6 +246,13 @@ struct SonataGPIO
...
+		return static_cast<SonataJoystick>(input & Inputs::Joystick);
 	}
 };
+
+using SonataGpioRaspberryPiHat = SonataGpioBase<0x0FFF'FFFF>;
+using SonataGpioArduinoShield  = SonataGpioBase<0x0000'3FFF>;
+using SonataGpioPmod           = SonataGpioBase<0x0000'00FF>;
+using SonataGpioPmod0          = SonataGpioPmod;
+using SonataGpioPmod1          = SonataGpioPmod;
+using SonataGpioPmodC          = SonataGpioBase<0x0000'003F>;
```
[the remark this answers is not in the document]
HU90m: Ah that's a short cut because so many of the GPIO devices happen to have the same number of input, output and output enable pins. The masks are all used for different registers.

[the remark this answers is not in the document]
rmn30: This was my confusion: inputs and outputs are separate registers but all these things have the same number of input / output pins so same mask for both.

sdk/include/platform/sunburst/platform-gpio.hh
nwf: I _think_ this is fine: most GPIO drivers can be configured to either I or O, these days.

sdk/include/platform/sunburst/platform-gpio.hh
```diff
@@ -136,8 +294,15 @@ struct SonataGPIO
...
 	 */
-	SonataJoystick read_joystick() volatile
+	JoystickValue read_joystick() volatile
 	{
-		return static_cast<SonataJoystick>(input & 0x1f);
+		return {static_cast<JoystickDirection>(input & Inputs::Joystick)};
 	}
 };
+
+using SonataGpioRaspberryPiHat = SonataGpioBase<0x0FFF'FFFF>;
```
nwf, line 303: Perhaps these should be structs inheriting from their current definitions and providing static factory methods for the singletons on Sonata.  I'm content to leave that for a later PR, tho'.

sdk/include/platform/sunburst/platform-gpio.hh
```diff
@@ -1,78 +1,236 @@
...
+	 * to query the value.
+	 */
+	class JoystickValue
+	{
+		JoystickDirection direction;
+
+		public:
+		JoystickValue(JoystickDirection direction) : direction(direction) {}
+
+		bool is_direction_enabled(JoystickDirection direction)
```
nwf, line 155: "enabled"?  Maybe "active"?

nwf (commented):
Nits aside, this looks OK.  I should still pull it and test (sorry).

sdk/include/platform/sunburst/platform-gpio.hh
rmn30: Yeah I was confused.

sdk/include/platform/sunburst/platform-gpio.hh
rmn30: Maybe pressed?

nwf (approved):

PR #438, opened 2025-01-30
Add support for levels in the ConditionalDebug class.

davidchisnall:
The class is now instantiated with a threshold level, from None to Error (for backwards compatibility, true means Information and false means None).

Log messages also provide a log level (Information if none is provided) and will fire if the log level is greater than or equal to the threshold.

### DIFF
--- a/sdk/include/debug.hh
+++ b/sdk/include/debug.hh
@@ -5,13 +5,11 @@
 #include <__debug.h>
 #include <cheri.hh>
 #include <compartment.h>
-#include <concepts>
 #include <cstddef>
 #include <platform-uart.hh>
 #include <string.h>
 #include <switcher.h>
 
-#include <array>
 #include <string_view>
 #include <type_traits>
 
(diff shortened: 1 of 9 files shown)


sdk/include/debug.hh
```diff
@@ -592,26 +666,31 @@ namespace
 	 * using Debug = ConditionalDebug<DebugFoo, "Foo">;
```
rmn30, line 666: Does this comment need updating?

sdk/include/debug.hh
```diff
@@ -592,26 +666,31 @@ namespace
...
 	 * ```
 	 */
-	template<bool Enabled, DebugContext Context>
+	template<DebugLevelTemplateArgument Threshold, DebugContext Context>
 	class ConditionalDebug
 	{
+		static constexpr bool Enabled = Threshold != DebugLevel::None;
```
rmn30, line 672: Where is this used?

rmn30: Ah, I see it is used in `ConditionalDebug::Invariant`.

rmn30 (commented):
Looks good but I don't understand how it is backwards compatible with `ConditionalDebug<false, "blah">`. Looks like that would create a Information level ConditionalDebug?

rmn30 (approved):
Ah, that's what DebugLevelTemplateArgument is for. I still think you could update the comments.

PhilDay-CT:
Would be nice if one or more of the examples was updated to show how this can be uses.

PR #440, opened 2025-01-31
rmn30/sonata joystick

rmn30:
- **Add sprintf to stdio.**
- **Fix sonata joystick GPIO driver for Sonata v1.**


rmn30:
Ooops, forgot about #437 .

PR #446, opened 2025-02-06
CILS accessors

nwf:


davidchisnall (commented):
For C++ it would be nice to have something that returns a reference to the slot that stores the pointer so that you can set and get it. Maybe also for C, return a `void**`, rather than separate set and get (maybe set and get as wrappers?)

davidchisnall (approved):

PR #448, opened 2025-02-12
Un-mangle libc string function names so they can benefit from LLVM optimizations.

resistor:


resistor:
This should address https://github.com/CHERIoT-Platform/llvm-project/issues/107

sdk/include/cdefs.h
```diff
@@ -94,7 +94,7 @@ using _Bool = bool;
  * other places) to avoid the namespace pollution from including `string.h` but
  * is either constant folded in the front end or expanded to a libcall.
  */
-unsigned __builtin_strlen(const char *str) __asm__("_Z6strlenPKc");
+unsigned __builtin_strlen(const char *str);
```
davidchisnall, line 97: I believe you can remove this declaration entirely.  Without it, clang will use the built-in definition.

sdk/include/string.h
```diff
@@ -6,27 +6,37 @@
...
-size_t __cheri_libcall strlen(const char *str);
-int __cheri_libcall    strncmp(const char *s1, const char *s2, size_t n);
-char *__cheri_libcall  strncpy(char *dest, const char *src, size_t n);
-int __cheri_libcall    strcmp(const char *s1, const char *s2);
-char *__cheri_libcall  strnstr(const char *haystack,
-                               const char *needle,
-                               size_t      haystackLength);
-char *__cheri_libcall  strchr(const char *s, int c);
-size_t __cheri_libcall strlcpy(char *dest, const char *src, size_t n);
+#define DECLARE_STRING_LIBCALL(rtype, name, ...)                               \
```
davidchisnall, line 9: It would be good to have a comment here for what this is doing (declaring libcall functions with unmangled names) and why it exists (to permit compiler optimisations).

Minor naming nit, but are all of the things where we care about compiler libcalls optimisations string functions?  If not, we should maybe call this `DECLARE_STANDARD_LIBCALL` or similar.

sdk/include/string.h
```diff
@@ -6,27 +6,37 @@
...
+DECLARE_STRING_LIBCALL(int, strcmp, const char *s1, const char *s2)
+DECLARE_STRING_LIBCALL(char *,
+                       strnstr,
+                       const char *haystack,
+                       const char *needle,
+                       size_t      haystackLength)
+DECLARE_STRING_LIBCALL(char *, strchr, const char *s, int c)
+DECLARE_STRING_LIBCALL(size_t, strlcpy, char *dest, const char *src, size_t n)
+
+#undef DECLARE_STRING_LIBCALL
```
davidchisnall, line 39: If you add a CHERIOT prefix on this, we can move it into cdefs.h.  There's a similar macro in lib/atomic and it would be nice to refactor that to use this one at the same time.

resistor: Done

resistor:
Ping

PR #453, opened 2025-02-16
Switch to [[cheriot::interrupt_state]]

resistor:


davidchisnall:
When was this introduced? We need to bump the compiler-too-old warning in cdefs.h if we’re doing this without a macro.

resistor:
> When was this introduced? We need to bump the compiler-too-old warning in cdefs.h if we’re doing this without a macro.

I added it on January 13.

resistor:
> When was this introduced? We need to bump the compiler-too-old warning in cdefs.h if we’re doing this without a macro.

I've added a version check, as there wasn't a straightforward one as-yet.

PR #457, opened 2025-02-18
Add test for compilation of setjmp / longjmp in C file.

rmn30:
This is a regression test for https://github.com/CHERIoT-Platform/llvm-project/issues/118 .


rmn30:
With current compiler this will fail to build if `-fno-builtin` is removed from `sdk/xmake.lua`. Eventually we want to do that and then this will become a proper regression test.

davidchisnall (approved):

PR #460, opened 2025-02-19
Remove -fno-builtin to enable LLVM libcall optimizations.

resistor:
Use -fno-builtin-setjmp and -fno-builtin-longjmp to workaround https://github.com/CHERIoT-Platform/llvm-project/issues/118 when using a compiler not new enough to contain the fix.


resistor:
I think this should be good to go in since https://github.com/CHERIoT-Platform/cheriot-rtos/commit/9675a9592f551c071051ad0595c16da9ce3608f4 is already in.

compile_flags.txt
```diff
@@ -10,7 +10,8 @@ riscv32-unknown-unknown
 -fshort-wchar
 -nostdinc
 -fomit-frame-pointer
--fno-builtin
+-fno-builtin-setjmp
+-fno-builtin-longjmp
```
davidchisnall, line 14: These are only used for clangd, so don't matter too much.

davidchisnall (approved):
LGTM

PR #463, opened 2025-02-20
Pass -fno-builtin-printf, too

nwf:
Otherwise, for example, building our network stack's MQTT compartment with debugging will get some of its printf()s replaced with putchar() and puts(), which we don't have.


davidchisnall (approved):

davidchisnall:
I suspect a better fix is to add these in our stdio compartment, but for now it's good to avoid this.

PR #464, opened 2025-02-20
Small improvements to the token APIs.

davidchisnall:
Expose `token_unseal` as the type-safe wrapper for everything (C macro, C++ templated function).  Make `token_allocate` not use the `Sealed` wrapper when it's not needed.

The macro currently triggers a compiler bug.

### DIFF
--- a/sdk/include/token.h
+++ b/sdk/include/token.h
@@ -238,4 +238,18 @@ __always_inline T *token_unseal(SKey key, Sealed<T> sealed)
 {
 	return static_cast<T *>(token_obj_unseal(key, sealed));
 }
+
 #endif // __cplusplus
+
+#if __has_extension(cheri_sealed_pointers)
+#	ifdef __cplusplus
+template<typename T>
+__always_inline T *token_unseal(SKey key, T *__sealed_capability sealed)
+{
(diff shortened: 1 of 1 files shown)


sdk/include/token.h
```diff
@@ -238,4 +238,18 @@ __always_inline T *token_unseal(SKey key, Sealed<T> sealed)
...
+
+#if __has_extension(cheri_sealed_pointers)
+#	ifdef __cplusplus
+template<typename T>
+__always_inline T *token_unseal(SKey key, T *__sealed_capability sealed)
+{
+	return static_cast<T *>(token_obj_unseal(key, sealed));
+}
+#	else
+#		define token_unseal(key, sealed) /*NOLINT*/                           \
```
nwf, line 252: Why won't this clash with the `token_unseal` function above, since its `T` is inferrable and so its invocations can be spelled without wakas before parens?

davidchisnall: Because there is no type that is both a `T *__sealed_capability` and a `Sealed<T>`

davidchisnall: The macro is visible only in C code.

nwf (approved):

PR #467, opened 2025-02-24
MCP251XFD CAN bus examples

GrassHopper1977:
## General Description
This is a generic MCP251XFD driver. The MCP2517FD and MCP2518FD are SPI to CAN drivers that support CAN2.0 and CAN FD. The driver is largely based on [Emandhal's Generic CAN251XFD Driver](https://github.com/Emandhal/MCP251XFD) and we have left the license intact.
The example is designed to run on the Sonata board, however, this is a generic driver with all the hardware specific changes located in one single file.

### CAN Devices Supported

- MCP2517FD
- MCP2518FD 

### sdk/boards/include/driver/MCP251XFD
This is the generic driver. We've altered the code so that it is all based in header files so that you don't need to add any extra C to the `xmake.lua` file when building. The code includes a CRC in a sub folder. We've not tested this yet but the device does include an option to add CRC to all SPI messages.

### examples/12.CAN_driver/
Sonata specific: This is an implementation of the MCP251XFD driver. The sonata specific interface code is found in `/examples/12.CAN_driver/driver/interface.cc` and it merely defines some hardware specific SPI and timing code.

### examples/11.basic_spi
Sonata specific: This example shows how to read and write on the Sonata's SPI interface. It shows the use of the bi-directional SPI transfer function that we added. Bi-directional transfer is often used in SPI drivers so it makes sense to add support for it.

### sdk/include/platform/sunburst/platform-spi.hh
Added a bi-directional blocking SPI transfer method (`void blocking_transfer(const uint8_t txData[], uint8_t rxData[], uint16_t len)`) as the driver (and many other SPI based drivers) expect a bi-directional transfer.
We also added another version of chip_select_assert() that could take the Chip Select index as a parameter as it is not always possible to specify which chip select you will be using ahead of time. I'm sure that there is a way of implementing this driver without the need for this additional method but I'm running out of time on this project.

## Items to be Tested
 - [x] Receive CAN2.0 with Standard IDs 
 - [x] Transmit CAN2.0 with Standard IDs 
 - [ ] Receive CAN2.0 with Extended IDs
 - [ ] Transmit CAN2.0 with Extended IDs
 - [ ] Receive CAN FD
 - [ ] Transmit CAN FD
 - [ ] Sonata Hardware on RPi Header
 - [ ] Sonata Hardware on other Headers
 - [x] Software polling Tx and Rx FIFOs
 - [ ] Interrupt polling Tx and Rx FIFOs
 - [ ] Message timing
 - [x] Receive Message Filters
 - [ ] Test driving two separate instances to 2 different ICs
 - [ ] CRC for all SPI messages

## Suggested Improvements
 - [ ] The `struct MCP251XFD` could be converted to a class allowing us to have private and public methods and allowing us to no longer need to have to keep passing the `struct` around.
 - [ ] Many static functions have now become public as a result of moving everything into the header. This should be made private methods.
 - [ ] There are many `#define` statements that could become `constexpr`. Many of these could safely be private to the class.
 - [ ] Putting it all in a namespace and a class would mean that we can drop the `MCP251XFD_` prefix (and occasional postfix) from everything.
 - [ ] The Sonata specific examples made require a different home rather than the `examples` folder. Maybe a hardware specific subfolder?
 - [ ] We don't any CHERI specific code in the driver. It may not be needed but Tx and Rx buffers could be made read or write only for additional protection before passing to sub functions. This will be more important for multi-threaded implementations.


examples/11.basic_spi/README.md
```diff
@@ -0,0 +1,44 @@
+<!--
+Copyright lowRISC Contributors.
+SPDX-License-Identifier: Apache-2.0
```
davidchisnall, line 3: Please can you make this match the existing copyright for this repo?

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -318,6 +318,54 @@ namespace SonataSpi
...
+			if (len)
+			{
+				control = ControlReceiveEnable | ControlTransmitEnable;
+				start   = len;
+	
+				uint32_t transmitAvailable = 0;
+				uint16_t txCnt = 0;
+				uint16_t rxCnt = 0;
+				uint8_t dummy;
+				while(rxCnt < len)
```
nwf, line 346: It'd be nicer if this were IRQ-driven instead, if it can be, especially if the SPI is operating at low clock speeds.  We should probably do the same with the UART driver, as another example...

sdk/include/platform/sunburst/platform-spi.hh
```diff
@@ -318,6 +318,54 @@ namespace SonataSpi
...
+				{
+					// Wait for at least one byte to be available in the RX FIFO
+					if((txCnt < len) && ((status & StatusTxFifoLevel) < 8)) {
```
nwf, line 349: Please don't poll the tx fifo fill level directly, as that will fail to adapt to hardware with differing FIFO depths.  Poll on `StatusTxFifoFull` instead.

sdk/include/driver/MCP251XFD/crc/CRC16_CMS.hh
```diff
@@ -0,0 +1,122 @@
...
+    Author:       FMA
+    Version:      1.0
+    Date (d/m/y): 14/05/2020
+    Description:  CRC16-CMS implementation
+                  The CRC16-CMS polynomial is x^16 + x^15 + x^2 + 1 (0x8005)
+                  Does not use RefIN and RefOUT, the initial value 0xFFFF
+                  The result is XORed with 0x0000
+                  http://reveng.sourceforge.net/crc-catalogue/16.htm
+
+    History :
```
nwf, line 12: License?

sdk/include/driver/MCP251XFD/crc/CRC16_CMS.hh
```diff
@@ -0,0 +1,122 @@
...
+//-----------------------------------------------------------------------------
+/// @cond 0
+/**INDENT-OFF**/
+#ifdef __cplusplus
+}
+#endif
+/**INDENT-ON**/
+/// @endcond
+//-----------------------------------------------------------------------------
+#endif /* CRC16_CMS_H_INC */
```
nwf, line 122: I think I'd rather see this as a library `.cc` instead of a header.  It seems plausible to me that multiple compartments would want this CRC (it's also the polynomial chosen for MODBUS, for example).

sdk/include/driver/MCP251XFD/crc/CRC16_CMS.hh
```diff
@@ -0,0 +1,122 @@
...
+//#define CRC16CMS_NOTABLE  // Uncomment to calculate the CRC on the fly instead of using the lookup table.
+
+#ifndef CRC16CMS_NOTABLE
+/*! @brief Compute a byte stream with CRC16-CMS with a table
+ *
+ * @param[in] *data Is the pointed byte stream
+ * @param[in] size Is the size of the pointed byte stream
+ * @return The CRC computed
+ */
+uint16_t ComputeCRC16CMS(const uint8_t* data, uint32_t size)
```
nwf, line 42: Very minor: if you make this (and the tableless counterpart below) take the initial `crc` value (and not do the "final xor" below, even though it's all zeros), then it can be used for streaming updates as well as whole-buffer operations, too.

nwf (commented):
Thanks!

Some initial comments at a glance.  Longer review still to come.

Please don't commit Windows' `:Zone.Identifier` files.

sdk/include/driver/MCP251XFD/crc/CRC16_CMS.hh
GrassHopper1977: Good question. It was part of the original driver that I adapted so I guess it's the same licence? I'm not quite ssure what the rules about changing a licence are.

sdk/include/platform/sunburst/platform-spi.hh
GrassHopper1977: Yes, I just combined the existing Tx and Rx examples. I've not got long left on this project but I may have time.

sdk/include/driver/MCP251XFD/crc/CRC16_CMS.hh
GrassHopper1977: I got the CRC from here and moved all the code into the header to make it easier to include with a single line: https://github.com/Emandhal/MCP251XFD/tree/master/CRC
This repo was licensed under MIT but it is possible that code was taken from elsewhere. I will message the author of the driver and ask about the license.

GrassHopper1977: I've modified a CRC8 function that I wrote years ago to be CRC16 and swapped the file.

sdk/include/driver/MCP251XFD/crc/CRC16_CMS.hh
GrassHopper1977: I've changed the CRC file completely to a modified version of a CRC8 function that I wrote years ago. You can now pass in a CRC table, initial value and final XOR. That should make it a little more flexible.

sdk/include/platform/sunburst/platform-spi.hh
GrassHopper1977: I've only got two weeks left on this project and that includes documenting it so I probably won't have time to implement an interrupt driven version.

GrassHopper1977:
I've added some more example code to play with I've just updated the PR with the details.

examples/11.basic_spi/basic_spi.cc
```diff
@@ -0,0 +1,281 @@
...
+// else will use the blocking_write() and blocking_read().
+#define SPI_TRANSFER_TEST
+
+// The SPI clock calculation
+// The settings is the length of a half period of the SPI clock, measured in system clock cycles reduced by 1.
+// The system clock is 40000000Hz (40MHz - 25ns)
+// The Arduino code example uses a 1MHz SPI clock. Which gives us a clock period of 1E-06 or 1uS.
+// Therefore, we want to try to set the half period to 500ns. So our setting, s, will be:
+// s = (500 / 25ns) - 1 = 20 - 1 = 19
+#define SPI_CLOCK_SPEED_SETTING	(19)
```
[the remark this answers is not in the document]
GrassHopper1977: I tried to get it working but I kept coming up with the wrong answer 😳

examples/11.basic_spi/README.md
```diff
@@ -0,0 +1,40 @@
...
+This will set up two SPI ports on the RPi header to query 2 MCP2518FD CAN controllers. This code won't fully drive them, instead it will just read one of the registers.
+
+## Notes
+* We will use the pinmux to enable the RPi ports to connect to the SPI modules and then we will activate the modules.
+* The SPI driver ius supposed to be capable of driving the CE (Chip Enable) but that would prevent us from a write, then read operations (SPI is bi-directional). Also, we couldn't get the CE lines to work driectly from the SPI.
+* The SPI driver has blocking read and blocking write functions already written.
+* SPI transfers are inherently bi directional so, in the long term, we may require a blocking_transfer function, however, it may not be needed for this example.
+* We will be using the Waveshare 2-CH CAN FD HAT for the purposes of testing. Details can be found here: https://www.waveshare.com/wiki/2-CH_CAN_FD_HAT
+* SPI0 is connected to other devices. Only SPI1 and SPI2 are coinnected to the RPi header. So SPI1 is connected to RPi SPI0 and SPI2 is connected to RPi SPI1 (confused yet?)
+* The SPI interrupts are currently not working but registers are reserved for them.
```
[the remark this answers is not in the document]
GrassHopper1977: Stale comment form before V1.0. I checked this file so many times and missed all this.

examples/11.basic_spi/basic_spi.cc
```diff
@@ -0,0 +1,281 @@
...
+	// Setting up the pinmux.
+	auto pinSinks = MMIO_CAPABILITY(SonataPinmux::PinSinks, pinmux_pins_sinks);
+	auto blockSinks = MMIO_CAPABILITY(SonataPinmux::BlockSinks, pinmux_block_sinks);
+
+	// SPI1 - COPI
+	if(true == pinSinks->get(SonataPinmux::PinSink::rph_g10).select(1)) {
+			Debug::log("Success! Set RPi GPIO10 to SPI1_COPI");
+	} else {
+		Debug::log("ERROR! Failed to set RPi GPIO10 to SPI1_COPI");
+	}
```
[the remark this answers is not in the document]
GrassHopper1977: That's more succinct. I'll try that.

sdk/include/driver/MCP251XFD/README.md
```diff
@@ -0,0 +1,54 @@
+# MCP251XFD
```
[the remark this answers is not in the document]
GrassHopper1977: Yes, maybe. @davidchisnall suggested this location but maybe it makes more sense a sub module.
Then the hardware specific drivers could be in a sub folder in the submodule and it would keep it all together.
Would you like the CRC code to stay in this repo? In which case, where would you like it placing?

PR #468, opened 2025-02-27
Add a more generally usable function for flushing quarantine.

davidchisnall:
This is similar to the existing one that was added for the test suite, but now takes a timeout and gracefully exits if the timeout expires.


sdk/core/allocator/main.cc
```diff
@@ -863,22 +863,53 @@ __cheriot_minimum_stack(0x90) ssize_t
 	return cap->quota;
 }
 
-__cheriot_minimum_stack(0xd0) int heap_quarantine_empty()
+__cheriot_minimum_stack(0xd0) int heap_quarantine_flush(Timeout *timeout)
 {
 	STACK_CHECK(0xd0);
```
rmn30, line 868: Aside: now that we have `__cheriot_minimum_stack__` can we introduce a variant of this macro that doesn't need repeat the stack size?

sdk/include/stdlib.h
```diff
@@ -291,17 +291,36 @@ ssize_t __cheri_compartment("allocator")
   heap_quota_remaining(AllocatorCapability heapCapability);
 
 /**
- * Block until the quarantine is empty.
+ * Try to empty the quarantine and defragment the heap.
  *
- * This should be used only in testing, to place the system in a quiesced
- * state.  It can block indefinitely if another thread is allocating and
- * freeing memory while this runs.
+ * This will start revocation sweep and then try to empty the quarantine.  In
```
rmn30, line 296:
```suggestion
 * This will start a revocation sweep and then try to empty the quarantine.  In
```

sdk/include/stdlib.h
```diff
@@ -291,17 +291,36 @@ ssize_t __cheri_compartment("allocator")
...
+__attribute__((overloadable)) int __cheri_compartment("allocator")
+  heap_quarantine_flush(Timeout *timeout);
+
+/**
+ * Compatibility implementation of `heap_quarantine_empty`.  This should be
+ * used only in debugging and testing because it is not guaranteed to terminate
+ * if another thread is allocating memory.
  */
-int __cheri_compartment("allocator") heap_quarantine_empty(void);
+static int heap_quarantine_empty(void)
```
rmn30, line 319: Could mark this deprecated and add a new version called `heap_quarantine_flush_block_forever`? Hopefully that's a sufficiently scary name to discourage inappropriate use and also consistent with naming convention?

sdk/core/allocator/main.cc
```diff
@@ -863,22 +863,54 @@ __cheriot_minimum_stack(0x90) ssize_t
...
+	}
+
+	if (LockGuard g{lock, timeout})
 	{
-		if (!gm->quarantine_dequeue())
+		while (gm->heapQuarantineSize > 0)
 		{
-			revoker.system_bg_revoker_kick();
+			auto epoch = revoker.system_epoch_get();
+			if (!gm->quarantine_dequeue())
```
nwf, line 880: This probably wants to be repeated, like `while (gm->quarantine_dequeue()) { ; }`, possibly with a notion of time elapsing, because there may be lots of (that is, more than 4) objects for us to remove from quarantine before we need to kick and wait for any revocation passes to take place.  We should re-check `gm->heapQuarantineSize > 0` afterwards in case the entire quarantine was on the finished or (already-revoked) pending rings.

sdk/core/allocator/main.cc
```diff
@@ -863,22 +863,54 @@ __cheriot_minimum_stack(0x90) ssize_t
...
+				revoker.system_bg_revoker_kick();
+			}
+			if constexpr (Revocation::Revoker::IsAsynchronous)
+			{
+				wait_for_background_revoker(timeout, epoch, g);
+			}
+			else
+			{
+				(void)epoch;
+				// Drop and reacquire the lock while yielding.
```
rmn30, line 891: This sequence could be extracted as a method on `LockGuard` so that this would become `if (!g.unlock_temporarily(timeout, 1)) return -ETIMEDOUT;`. Do we have other instances of this pattern?

sdk/core/allocator/main.cc
davidchisnall: I've refactored this a lot, what do you think of the new version?

sdk/include/stdlib.h
davidchisnall: We probably could just move this into the test suite, but I don't like breaking existing APIs, even if they're silly.

sdk/core/allocator/main.cc
```diff
@@ -305,27 +333,8 @@ namespace
...
-						{
-							/* Unable to sleep; bail out */
-							return nullptr;
-						}
-						if (!reacquire_lock(timeout, g, smallSleep.elapsed))
-						{
-							return nullptr;
-						}
-					}
+					wait_for_background_revoker(
```
rmn30, line 336: Should this check the return value?

davidchisnall: Yes.  That was an existing bug, but it's now fixed.

sdk/include/stdlib.h
```diff
@@ -291,17 +291,36 @@ ssize_t __cheri_compartment("allocator")
...
- * This should be used only in testing, to place the system in a quiesced
- * state.  It can block indefinitely if another thread is allocating and
- * freeing memory while this runs.
+ * This will start a revocation sweep and then try to empty the quarantine.  In
+ * normal operation, the allocator will remove a small number of allocations
+ * from quarantine on each allocation.  Allocations are not coalesced until
+ * they are moved from quarantine, so this can cause fragmentation.  If you
+ * have just freed a lot of memory (for example, after resetting a compartment
+ * and calling `heap_free_all`), especially if you have freed a lot of small
+ * allocations, then calling this function will reduce fragmentation.
```
nwf, line 302: I don't think it's a certainty?

```suggestion
 * allocations, then calling this function will likely reduce fragmentation.
```

sdk/core/allocator/main.cc
nwf: Running `while(gm->quarantine_dequeue()) { ; }` twice with a revocation in the middle will get _most_ of the quarantine most of the time (anything finished or already revoked prior to its invocation, and anything in quarantine prior to its kick of the revoker), it won't necessarily get things freed during its own operation.  That appears to be deliberate, and is probably fine semantics, but the documentation could be a little more clear, perhaps.

sdk/include/stdlib.h
```diff
@@ -291,17 +291,42 @@ ssize_t __cheri_compartment("allocator")
...
+ * not valid.
  */
-int __cheri_compartment("allocator") heap_quarantine_empty(void);
+__attribute__((overloadable)) int __cheri_compartment("allocator")
+  heap_quarantine_flush(Timeout *timeout);
+
+/**
+ * Compatibility implementation of `heap_quarantine_empty`.  This should be
+ * used only in debugging and testing because it is not guaranteed to terminate
+ * if another thread is allocating memory.
```
nwf, line 323: Since `heap_quarantine_flush` is now one-pass, this should also be guaranteed to terminate, now... or should it loop around `_flush`?

sdk/core/allocator/main.cc
```diff
@@ -216,20 +217,44 @@ namespace
...
 			{
 				return false;
 			}
-			if (!reacquire_lock(timeout, g, smallSleep.elapsed))
+		}
+		return true;
+	}
+
+	template<typename T = Revocation::Revoker>
+	bool wait_for_background_revoker(Timeout                   *timeout,
```
nwf, line 239: At a glance, this implementation looks to be the same as the above?

davidchisnall: Hmm, after the refactoring the without-interrupt one might now be equivalent to the synchronous one, yes.

nwf: That still looks to be true, assuming I'm not missing something; can we cut this down to one rather than two?

Alternatively, should the one below more aggressively kick (that is, make incremental progress on) the revocation, rather than just one "unit" of work per `has_revocation_finished_for_epoch` call?

sdk/include/stdlib.h
nwf: I think this still needs updating one way or the other?

sdk/include/stdlib.h
nwf: I'd be in favor of moving this to the test suite under a different name and marking this one as deprecated.

nwf (commented):
Some nits, but generally LGTM.  Could merge now and fix nits later, if desired.

nwf (approved):
Approving with permission. ;)

PR #473, opened 2025-03-07
Reduce scheduler permissions slightly.

davidchisnall:
The scheduler doesn't require store permission for futexes when it does futex_wake, so don't require any permissions here.  This means that you can spuriously wake any thread waiting on a futex that you have any pointer to, even a read-only one, but the API permits spurious wakes.

The benefit from this is that the scheduler cannot tamper with a futex word.  For example, if a futex is a trivial lock where 1 indicates locked and 0 indicates unlocked, the scheduler currently can both wake waiters and set the lock value to 0 making two threads believe that they acquired the same lock.  With this change, an attacker who gets arbitrary code execution in the scheduler cannot do this, they can only cause threads to wake, see that the lock is already acquired, and then go back to sleep.

The C++ atomics (used for the locks library) now enforce this least privilege when calling the scheduler.

These changes increase the size of the test suite by 32 bytes.

### DIFF
--- a/sdk/include/futex.h
+++ b/sdk/include/futex.h
@@ -67,10 +67,14 @@ __always_inline static int futex_wait(const uint32_t *address,
  * Wakes up to `count` threads that are sleeping with `futex[_timed]_wait` on
  * `address`.
  *
- * The `address` argument must permit storing four bytes of data after the
- * address. This call does not store to the address but requiring store
- * permission prevents a thread from waking up a futex that it cannot possibly
- * have moved to a different state.
+ * The `address` argument must be a valid, unsealed, pointer with a length of at
+ * least four after the address but the scheduler does not require any explicit
+ * permissions.  The scheduler never needs store access to the futex word.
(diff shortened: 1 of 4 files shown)


sdk/include/futex.h
```diff
@@ -67,10 +67,14 @@ __always_inline static int futex_wait(const uint32_t *address,
  * Wakes up to `count` threads that are sleeping with `futex[_timed]_wait` on
  * `address`.
  *
- * The `address` argument must permit storing four bytes of data after the
- * address. This call does not store to the address but requiring store
- * permission prevents a thread from waking up a futex that it cannot possibly
- * have moved to a different state.
+ * The `address` argument must have a length of at least four after the address
```
nwf, line 70: May be worth more fully enumerating what `check_pointer` is going to ask for: valid, unsealed, and length >= 4.

nwf (approved):

davidchisnall:
Hmm, I thought the tests ran, but it appears that they actually fail to wake in the queue tests.

PR #475, opened 2025-03-10
Remove storage classes from explicit instantiations.

resistor:
clang-19 warns that this is not allowed.


resistor:
Interesting, this errors with with clang-18. I guess we'll need to wait for clang-19 to roll out.

davidchisnall (approved):

PR #476, opened 2025-03-10
Fix clang-tidy modernization warning with clang-19

resistor:


resistor:
@davidchisnall What is this code actually trying to do? The implementation of `&operator=(T *t)` seems highly questionable, as it's storing a reference to a temporary...

davidchisnall:
@nwf , this is your code, I believe?

resistor:
@nwf ping

nwf:
IIRC, it's attempting to provide an implementation of `{ proxy = ptr } -> std::same_as<P &>;` after which `proxy` will act as though it had value `ptr`?

resistor:
@nwf This line seems wrong: https://github.com/CHERIoT-Platform/cheriot-rtos/blob/c551d357ee5da1a0f03714f7f9389844f3aeda4e/sdk/include/ds/pointer.h#L99

It's storing a reference to a function argument in `Pointer`, so the reference outlives the lifetime of the value.

davidchisnall:
The original code was correct.  The constructor captured a reference to a pointer.  The assignment operator then stored the pointer argument (as a value) *through* the captured reference.

In the new code, the reference to a pointer is captured but `std::reference_wrapper<T*>::operator=` *rebinds* the reference, it does not store the value through the reference.  This updates the reference wrapper to refer to the on-stack / in register argument.  Any subsequent operation will just break.  This could be fixed, I believe, by doing `ref.get() = p.ref;` on line 107.

I don't see any value from `std::reference_wrapper` here.  We don't want to be able to rebind the reference after construction (that's why we're using a reference!).  The right fix is probably to disable this clang-tidy warning on line 82.  Does clang-tidy explain why it thinks this is a good idea?  It seems to make the code less readable for no benefit.

nwf:
I'll raise a PR with some expanded commentary on this tautological implementation to check my understanding and hopefully reduce confusion in the future.

resistor:
> We should also probably report an issue against this clang-tidy FIXIT hint. Replacing a type with one with totally different semantics for `operator=` is a _great_ way of introducing subtle bugs. I can imagine adding that to clang-tidy as a supply-chain attack on a specific project.

FWIW, the fixit didn't directly suggest replacing the type, just using `= default`.

resistor:
Proposing a new fix that makes the intent of this code more clear, and convinces clang-tidy that this is not a candidate for a default assignment operator.

sdk/include/ds/pointer.h
```diff
@@ -104,7 +104,7 @@ namespace ds::pointer
 
 			__always_inline Pointer<T> &operator=(Pointer const &p)
 			{
-				ref = p.ref;
+				ref = (T *){p.ref};
```
davidchisnall, line 107: Doesn't this trigger a clang-tidy warning about C-style casts?

resistor: I’ll double check tomorrow but I didn’t see one when I ran it. It’s intended to be a direct-list-initialization, not a C-style cast.

davidchisnall: Hmm.  I find that less readable than the original, but maybe that's just me.

resistor: How would you feel about `ref = static_cast<T*>(p.ref)` or `*&ref = p.ref`?

resistor: Switched to `static_cast`

davidchisnall (approved):

PR #477, opened 2025-03-10
Suppress identifier readability lints in places where function names are part of ABI.

resistor:


nwf (approved):
Seems sensible enough to me!

PR #481, opened 2025-03-17
Disable interrupts if a thread's entry point is marked as

### LINKED ISSUE #129: Loader does not check interrupt status for thread entry points
If you export an entry point function with interrupts disabled and use it as a thread entry point, it will run with interrupts enabled.  This is annoying for realtime control loops and is fixable: we look up the entry point by its export table entry and so can set the relevant bit in mstatus if necessary.  Doing this for the highest-priority thread means that it runs, uninterrupted, until it explicitly yields (and then resumes with interrupts disabled).

davidchisnall:
interrupts-disabled.

Fixes #129

### DIFF
--- a/tests/test-runner.cc
+++ b/tests/test-runner.cc
@@ -77,11 +77,26 @@ compartment_error_handler(struct ErrorState *frame, size_t mcause, size_t mtval)
 	return ErrorRecoveryBehaviour::InstallContext;
 }
 
+__attribute__((noinline, weak)) void *pcc_as_sentry()
+{
+	return __builtin_return_address(0);
+}
+
 /**
  * Test suite entry point.  Runs all of the tests that we have defined.
  */
+[[cheriot::interrupt_state(disabled)]]
 int __cheri_compartment("test_runner") run_tests()
 {
+	// This is started as an interrupts-disabled thread, make sure that it
+	// really is!  This should always be CheriSealTypeReturnSentryDisabling,
+	// but older ISA versions didn't have separate forward and backwards
+	// sentries and so allow any kind of interrupt-disabled sentry.
+	Capability sealedPCC = pcc_as_sentry();
+	TEST((sealedPCC.type() == CheriSealTypeReturnSentryDisabling) ||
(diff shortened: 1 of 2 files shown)


tests/test-runner.cc
```diff
@@ -77,11 +77,23 @@ compartment_error_handler(struct ErrorState *frame, size_t mcause, size_t mtval)
...
 /**
  * Test suite entry point.  Runs all of the tests that we have defined.
  */
+[[cheriot::interrupt_state(disabled)]]
 int __cheri_compartment("test_runner") run_tests()
 {
+	// This is started as an interrupts-disabled thread, make sure that it really is!
+	Capability sealedPCC = pcc_as_sentry();
+	TEST((sealedPCC.type() == CheriSealTypeReturnSentryDisabling) ||
+	       (sealedPCC.type() == CheriSealTypeSentryDisabling),
```
rmn30, line 94: Shouldn't it always be a return sentry? Why do you need the `||`?

tests/test-runner.cc
```diff
@@ -77,11 +77,23 @@ compartment_error_handler(struct ErrorState *frame, size_t mcause, size_t mtval)
...
 
+__attribute__((noinline, weak)) void *pcc_as_sentry()
+{
+	return __builtin_return_address(0);
+}
+
 /**
  * Test suite entry point.  Runs all of the tests that we have defined.
  */
+[[cheriot::interrupt_state(disabled)]]
```
rmn30, line 88: This will only apply to the top level test runner, right? All the tests that run in compartments will default to interrupts enabled?

tests/test-runner.cc
davidchisnall: Sonata 0.2

tests/test-runner.cc
davidchisnall: Yes, otherwise the test suite would have broken.

rmn30 (approved):
I think it's fine but bit of a scary thing to do.

PR #482, opened 2025-03-17
sunburst: default to debug being switched off

HU90m:
Debug flag should probably be consistent among drivers and should probably default to being off


nwf:
I wonder if we should have a way for the board `JSON` files to define `--debug-XXX` options in the `xmake.lua` and have the board driver headers use those.

PR #491, opened 2025-03-25
New external IRQ latency benchmark

nwf:
This uses existing peripherals that can be used as IRQ "loopbacks",
which fire some appreciable delay after a software event, to measure
cycle, instruction, or HPMC counters of interest.  Presently, the Ibex
revoker engine or the Sunburst UART1 can be used as triggers; it should
be easy to add other sources.

### DIFF
--- a/sdk/include/platform/ibex/platform-hardware_revoker.hh
+++ b/sdk/include/platform/ibex/platform-hardware_revoker.hh
@@ -204,6 +204,31 @@ namespace Ibex
 			// timeout parameter, we fail either way.
 			return false;
 		}
+
+		/**
+		 * Reveal the interrupt futex word.
+		 *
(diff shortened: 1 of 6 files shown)


sdk/include/platform/ibex/platform-hardware_revoker.hh
```diff
@@ -24,7 +24,7 @@ namespace Ibex
 {
 	class HardwareRevoker
 	{
-		private:
+		public:
```
davidchisnall, line 27: I don’t like making these fields public just for a benchmark. What API are we missing?

nwf: Good call.  Have a hopefully much better spin.

davidchisnall (approved):

PR #498, opened 2025-03-28
sdk/xmake: support board mixins

nwf:
Mixins are "baseless" patches, or just a list of JSON patch directives.

Add an example to remove the use of HyperRAM on Sonata 1.x builds, removing the `"data_memory"` key and moving the `"instruction_memory"` map's values to those of SRAM.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -85,6 +85,25 @@ jobs:
           xmake
         done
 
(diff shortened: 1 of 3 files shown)


rmn30:
Looks useful. Could we add to the ci matrix?

davidchisnall (approved):

PR #502, opened 2025-04-01
Fully resolve commands in $PATH to an absolute path when searching for LLVM utilities.

resistor:


scripts/includes/helper_find_llvm_install.sh
```diff
@@ -7,7 +7,7 @@ DEV_CONTAINER_BIN="/cheriot-tools/bin/"
 #             `llvm-objdump`, `llvm-objcopy`)
 find_llvm_tool() {
 	TOOL_NAME=$1
-	LLVM_TOOL=${TOOL_NAME}
+	LLVM_TOOL=$(command -v ${TOOL_NAME})
```
davidchisnall, line 10: This is making it find the system tools in preference to the CHERIoT ones, which is why CI is failing.

nwf: I am confused; shouldn't `type` search the `$PATH` like `command -v` does?

nwf: Oh, hm.  `type` on an empty string is not an error, so if `command -v` doesn't find `${TOOL_NAME}`, then we return the empty string, which is surely wrong.

PR #505, opened 2025-04-06
More ds::linked_list fun

nwf:


sdk/include/ds/concepts.h
```diff
@@ -0,0 +1,30 @@
...
+namespace ds
+{
+
+	/**
+	 * Concepts meant to label functions as field projectors and their
+	 * container_of-like inverses.
+	 *
+	 * @{
+	 */
+
```
rmn30: These don't appear to be used anywhere?

nwf: Good catch, thanks!

sdk/include/ds/linked_list.h
```diff
@@ -324,33 +325,50 @@ namespace ds::linked_list
...
+	 * In order to guarantee termination, `elem` must always remain reachable
+	 * from the iterator.
 	 */
-	template<cell::HasCellOperations Cell, typename F>
-	__always_inline bool search(Cell *elem, F f)
+	template<bool Reverse = false, cell::HasCellOperations Cell, typename F>
+	__always_inline auto search(Cell *elem, F f)
 	{
-		return search(static_cast<Cell *>(elem->cell_next()), elem, f);
+		return search<Reverse>(static_cast<Cell *>(elem->cell_next()), elem, f);
```
rmn30, line 371: Why is this cast required?

sdk/include/ds/linked_list.h
```diff
@@ -527,19 +527,20 @@ namespace ds::linked_list
 	{
 
 		/** Cons cell using two pointers */
-		class Pointer
+		template<typename Type>
+		class PointerF
```
rmn30: Excuse my ignorance, but why 'F'?

rmn30: Ah from the commit message, for Functor I guess.

sdk/include/ds/linked_list.h
nwf: Because C++ isn't willing to do an implicit conversion from a pointer proxy to a pointer here:
```
sdk/include/ds/linked_list.h:346:40: note: candidate template ignored: could not match 'Cell *' against 'ds::pointer::proxy::Pointer<ds::linked_list::cell::Pointer>'
  346 |         __always_inline std::optional<Cell *> search(Cell *from, Cell *to, F f)
```

(ETA: sorry for the delay; went spelunking through https://en.cppreference.com/w/cpp/language/adl.html and friends to see if there was something convenient to cite.  I am at a loss.)

sdk/include/ds/linked_list.h
nwf: Yeah, a Haskell-ism that crept in.  I should probably stick to this repo's policy and spell it out.

sdk/include/ds/linked_list.h
rmn30: Fair enough, was reading commits in order and thought it was odd considering it wasn't required in the loop in 'search'.

sdk/include/ds/linked_list.h
```diff
@@ -526,133 +526,168 @@ namespace ds::linked_list
...
-		class Pointer
+		namespace functor
 		{
-			Pointer *prev, *next;
 
-			public:
-			Pointer()
+			/** Cons cell using two pointers */
+			template<typename Type>
+			class Pointer
```
rmn30, line 626: Should this be called a `PointerConsCell` or something a bit more descriptive? It doesn't really look like a Pointer...

sdk/include/ds/linked_list.h
```diff
@@ -526,133 +526,168 @@ namespace ds::linked_list
...
+			};
+
+			/**
+			 * Encode a linked list cons cell as a pair of addresses (but
+			 * present an interface in terms of pointers).  CHERI bounds on the
+			 * returned pointers are inherited from the pointer to `this` cons
+			 * cell.
+			 */
+			template<typename Type>
+			class __cheri_no_subobject_bounds PtrAddr
```
rmn30, line 659: As above, perhaps a more descriptive name?

sdk/include/ds/linked_list.h
rmn30: Hmm I missed that this is in a namespace, but I still think the name is a bit confusing (especially as it is the same as `Ds:pointer::proxy::Pointer`.

sdk/include/ds/linked_list.h
```diff
@@ -526,133 +526,168 @@ namespace ds::linked_list
...
+
+				__always_inline bool cell_is_doubleton()
+				{
+					return prev == next;
+				}
+			};
+
+		} // namespace functor
+
+		struct Pointer : public functor::Pointer<Pointer>
```
rmn30, line 770: Oh god, what's going on?! So many `Pointer`s. I guess this is CRTP?

sdk/include/ds/linked_list.h
```diff
@@ -526,133 +526,168 @@ namespace ds::linked_list
...
+				{
+					return prev == next;
+				}
+			};
+
+			/**
+			 * Encode a linked list cons cell as a pair of addresses (but
+			 * present an interface in terms of pointers).  CHERI bounds on the
+			 * returned pointers are inherited from the pointer to `this` cons
+			 * cell.
```
rmn30: This should probably mention `Offset` somewhere.

sdk/include/ds/linked_list.h
nwf: Yeah, it's a `ds::linked_list::cell::Pointer`, but please feel free to suggest alternatives!

sdk/include/ds/linked_list.h
nwf: Yeah.  I guess I could try using "deducing `this`" instead?

sdk/include/ds/linked_list.h
rmn30: I think `PointerConsCell` or `PointerCell` would make sense: it's a cons cell implemented using pointers. Admittedly I was exposed to Java at an early age and it may have affected me permanently.

sdk/include/ds/linked_list.h
rmn30: I don't know. It was late and I was just a bit a bit amazed at how something so simple could get so complicated...

PR #506, opened 2025-04-07
debug.hh: use 'enum struct DebugLevel'

nwf:
Prevent automagic conversion of DebugLevel to a boolean, because DebugLevel::None is nonzero (in fact, the largest of the DebugLevel values) as it is meant to be used as a threshold for reporting.

The use of an 'enum struct' means that we can't mistakenly write things like "if constexpr (DEBUG...)" and end up with debug code compiled in when DEBUG... is DebugLevel::None (or wonder why debugging might be elided when it's ::Information).

Update the sole user of DebugLevel, the allocator, to compile.

### DIFF
--- a/compile_flags.txt
+++ b/compile_flags.txt
@@ -21,7 +21,7 @@ riscv32-unknown-unknown
 -Isdk/include/platform/generic-riscv
 -Isdk/include
 -DDEBUG_LOADER=true
--DDEBUG_ALLOCATOR=true
+-DDEBUG_ALLOCATOR=DebugLevel::Information
 -DDEBUG_SCHEDULER=true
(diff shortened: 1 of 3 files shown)


davidchisnall:
Is enum struct equivalent to enum class? I have never seen this construct before.

nwf:
Apparently exactly so: https://en.cppreference.com/w/cpp/language/enum says about scoped enumerations "(the keywords class and struct are exactly equivalent)".  (Though that text is somewhat oddly placed, as if it might not be true of all forms, but AFAICT it is?)

Would you prefer `enum class`, if it's going to be more familiar to readers?

davidchisnall (approved):
Approving with `s/enum struct/enum class/`

PR #509, opened 2025-04-09
allocator-test: pass non-GLobal pointer to free

nwf:
I don't think we were previously testing that this worked.  It does, because `heap_free()` very quickly moves from using the pointer to using its cursor and internal pointers, but just to be sure...

### DIFF
--- a/tests/allocator-test.cc
+++ b/tests/allocator-test.cc
@@ -834,6 +834,23 @@ int test_allocator()
 	debug_log("Heap size is {} bytes", HeapSize);
 
 	test_preflight();
+
+	// Make sure that free() accepts delegated (non-global) pointers.
+	{
+		Timeout t{UnlimitedTimeout};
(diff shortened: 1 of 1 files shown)


rmn30 (approved):

tests/allocator-test.cc
```diff
@@ -801,6 +801,24 @@ int test_allocator()
...
+		void *volatile p = heap_allocate(&t, MALLOC_CAPABILITY, 32);
+		TEST(Capability{p}.is_valid(),
+		     "Could not perform an early allocation: {}",
+		     p);
+
+		Capability q = {p};
+		q.without_permissions(CHERI::Permission::Global);
+
+		int res = heap_free(MALLOC_CAPABILITY, q);
+		TEST(res == 0, "Could not free non-global pointer: {}", res);
```
rmn30, line 817: Could be TEST_EQUAL or TEST_SUCCESS.

nwf: Good call.  Thanks!

PR #510, opened 2025-04-09
Sunburst: tidy global defines

nwf:
These values are available to the linker under symbolic names and need not be duplicated.


davidchisnall (approved):

HU90m (approved):
Neat, thanks!

PR #514, opened 2025-04-16
Nwf/202504 test tweaks

nwf:
Very minor changes to the test suite


davidchisnall (approved):

nwf:
OK, here's a version that does the thread.priority "repacking".  It reverts the change to the test suite thread priorities so that we non-trivially exercise the machinery in question.

PR #515, opened 2025-04-16
benchmarks/compartment-call: many more points of interest

nwf:


nwf:
Sorry for taking ages to get back to this.

> ```
> diff --color -r /tmp/cheriot-rtos/benchmarks/compartment-call/caller.cc cheriot-rtos/benchmarks/compartment-call/caller.cc
> 11a12,14
> > DECLARE_AND_DEFINE_ALLOCATOR_CAPABILITY(MALLOC_CAP_TWO, 1024);
> > DECLARE_AND_DEFINE_ALLOCATOR_CAPABILITY(MALLOC_CAP_THREE, 1024);
> ```
OK, easy enough...

> ```
> 50a54,58
> > 
> > 	Timeout t{UnlimitedTimeout};
> > 	void   *heapPtr =
> > 	  heap_allocate(&t, STATIC_SEALED_VALUE(MALLOC_CAP_TWO), 100);
> >
> ```

I have no idea where this hunk is supposed to end up!  Line 54 looks to be inside `benchAround` but that surely isn't right.

> ```
> 86,96c94,141
> < 		printf("#cross call with check_pointer\n");
> < 		benchWithMiddle(
> < 		  [](int) { return check_pointer_return_metric(&nextResult); });
> < 
> < 		printf("#cross call with heap_claim_ephemeral\n");
> < 		benchWithMiddle(
> < 		  [](int) { return ephemeral_claim_return_metric(&nextResult); });
> < 
> < 		printf("#cross call with claim_release\n");
> < 		benchWithMiddle(
> < 		  [](int) { return claim_release_return_metric(&nextResult); });
> ```

Are these really not worth keeping?  (They should probably be renamed to `callee_`, tho'.)

>```
> > 		// Note that the branch is included in the cycles calculation
> > 		// (and undesired), but it's a handy way to avoid `cap` being
> > 		// optimized away.
> > 		printf("#de-privilege pointer\n");
> > 		benchWithMiddle([&](int) {
> > 			using namespace CHERI;
> > 			Capability cap{heapPtr};
> > 			cap.permissions() &= Permission::Load;
> > 			cap.bounds() = 1;
> > 			return cap.is_valid() ? 1 : 0;
> > 		});
> ```

Isn't that nearly the same thing as the existing
```
		printf("#function call pointer manipulation\n");
		benchWithMiddle(
		  [](int) { return local_attenuate_return_metric(&nextResult); });
```
The lambda isn't marked no-inline while the `local_attenuate...` function is, so I think the existing code is more robust to smarter compilers that might move code across loops &c.

> ```
> > 		printf("#check_pointer for a heap pointer\n");
> > 		benchAround([&](int) {
> > 			using namespace CHERI;
> > 			if (!check_pointer<PermissionSet{Permission::Load}>(heapPtr))
> > 			{
> > 				return;
> > 			}
> > 		});
> ```

Again, inlining and optimization concerns... do you want a `local_check_pointer` call?  Is there a reason _not_ to cross-call this one? (`callee_check_pointer_return_metric`)?

> ```
> > 		printf("#heap_claim_ephemeral for a heap pointer\n");
> > 		benchAround([&](int) {
> > 			Timeout t{UnlimitedTimeout};
> > 			heap_claim_ephemeral(&t, heapPtr);
> > 		});
> ```

This one's a little weird... we're going to hold on to those claims until we replace them or make a cross-call.  Remind me what's wrong with `callee_ephemeral_claim_return_metric`?

> ```
> > 		printf("#claim and claim_release\n");
> > 		// Get some buffers to claim/unclaim. Worried that
> > 		// claiming/unclaiming the same buffer in a loop may not be the
> > 		// same as claiming/unclaiming a buffer for the first time in a
> > 		// loop.
> > 		void *allocs[5];
> > 		for (int i = 0; i < 5; i++)
> > 		{
> > 			Timeout t{UnlimitedTimeout};
> > 			allocs[i] =
> > 			  heap_allocate(&t, STATIC_SEALED_VALUE(MALLOC_CAP_TWO), 10);
> > 		}
> > 		// Do a first claim/unclaim on an unrelated pointer, as it
> > 		// seems that the very first claim made has a weirdly higher
> > 		// latency.
> > 		heap_claim(MALLOC_CAPABILITY, heapPtr);
> > 		heap_free(MALLOC_CAPABILITY, heapPtr);
> > 		benchAround([&](int i) {
> > 			heap_claim(MALLOC_CAPABILITY, allocs[i]);
> > 			heap_free(MALLOC_CAPABILITY, allocs[i]);
> > 		});
> 158c203
> < 			  MALLOC_CAPABILITY,
> ---
> > 			  STATIC_SEALED_VALUE(MALLOC_CAP_THREE),
> ```

Oh, mm, sort of... there's a little bit of "first time" processing [per malloc capability](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/f95fc42658e9f80d86a957f02234403c94a582a6/sdk/core/allocator/main.cc#L403-L412) and [when anyone stakes their first claim on an object](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/f95fc42658e9f80d86a957f02234403c94a582a6/sdk/core/allocator/main.cc#L690-L722)... and indeed, if the claimant is the owner, as it is in the code above, we convert their ownership to a claim (and mark the object unowned).

That said, in `callee_claim_release_return_metric`, it will be a non-owner doing the claim, so that should go through the same code every time... except the first when the quota object is assigned an owner ID.  I don't _think_ there's utility to using multiple objects here, as each drop should return the claim chain to its initial, empty state.  I've added a comment.

> ```
> 167c212
> < 			Debug::Assert(res == &unsealeds[0], "Bad unseal?");
> ---
> > 			//Debug::Assert(res == &unsealeds[0], "Bad unseal?");
> ```

Assertions should completely compile away in release builds; is this OK to keep?

PR #517, opened 2025-04-22
Fix clang-tidy errors with clang-20.

resistor:


davidchisnall (approved):

sdk/lib/queue/queue.cc
```diff
@@ -403,8 +403,8 @@ int queue_destroy(AllocatorCapability  heapCapability,
 
 ssize_t queue_allocation_size(size_t elementSize, size_t elementCount)
 {
-	size_t bufferSize;
-	size_t allocSize;
+	size_t bufferSize = 0;
+	size_t allocSize = 0;
```
davidchisnall, line 407: Actually, less convinced by this.  Both of these are initialised later by being the output parameter for `__builtin_{add,mul}_overflow`.  Initialising them here will silence warnings if this isn't the case.  The analyser should probably be taught that these builtins write through their output argument and don't read it.

resistor: Switched to NOLINT instead.

davidchisnall: Can you also add a comment explaining why and indicating that these should be removed once the static analyser has learned about the semantics of Clang's own builtins?  And maybe file an upstream bug that we can reference in the comment?

resistor: Done

PR #518, opened 2025-04-23
clang-format with clang-20

resistor:


resistor:
> Approving to unblock CI, but the old formatting was the desired one. Has new clang-format separated this out somehow for lambdas and do we need a new line in our .clang-format file?

I'm not seeing anything related to this in upstream `clang-format` documentation.

PR #520, opened 2025-04-25
Softfloat implementation and smoke test.

davidchisnall:
This is divided into multiple libraries so that it's easy to pick just the operations that you need.  There are four convenience targets:

 - softfloat32 (almost everything for `float`)
 - softfloat64 (almost everything for `double`)
 - softfloat (almost everything)
 - softfloatall (everything, including complex numbers)

Most of the time, one of the first three will be adequate.

The total code comes in at a bit under 15 KiB.
Just 32-bit float support (no complex numbers, no powers) comes in at under 4 KiB.

### DIFF
--- a/sdk/include/debug.hh
+++ b/sdk/include/debug.hh
@@ -80,6 +80,20 @@ struct DebugWriter
 	 */
 	virtual void write_hex_byte(uint8_t) = 0;
 	/**
+	 * Write a single-precision floating-point value.
+	 *
+	 * If floating-point support is not compiled into the debug library, this
+	 * may print a placeholder.
+	 */
+	virtual void write(float) = 0;
+	/**
+	 * Write a double-precision floating-point value.
+	 *
+	 * If floating-point support is not compiled into the debug library, this
+	 * may print a placeholder.
+	 */
+	virtual void write(double) = 0;
+	/**
 	 * Write an integer as hex.
 	 */
 	template<typename T>
@@ -284,6 +298,29 @@ struct DebugFormatArgumentAdaptor<int64_t>
 		memcpy(&fudgedValue, &value, sizeof(fudgedValue));
 		return {fudgedValue,
 		        DebugFormatArgumentKind::DebugFormatArgumentSignedNumber64};
+	}
+};
+
+template<>
+struct DebugFormatArgumentAdaptor<float>
+{
+	__always_inline static DebugFormatArgument construct(float value)
+	{
+		uintptr_t fudgedValue = 0;
+		memcpy(&fudgedValue, &value, sizeof(value));
+		return {fudgedValue, DebugFormatArgumentKind::DebugFormatArgumentFloat};
+	}
+};
+
+template<>
+struct DebugFormatArgumentAdaptor<double>
+{
+	__always_inline static DebugFormatArgument construct(double value)
+	{
+		uintptr_t fudgedValue = 0;
+		memcpy(&fudgedValue, &value, sizeof(value));
+		return {fudgedValue,
+		        DebugFormatArgumentKind::DebugFormatArgumentDouble};
 	}
 };
 
--- a/sdk/lib/debug/debug.cc
+++ b/sdk/lib/debug/debug.cc
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: MIT
 
 #include <debug.hh>
+#include <float_to_string_helpers.hh>
 #include <thread.h>
 
 using namespace CHERI;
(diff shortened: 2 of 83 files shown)


sdk/include/debug.hh
```diff
@@ -287,6 +301,29 @@ struct DebugFormatArgumentAdaptor<int64_t>
 	}
 };
 
+template<>
+struct DebugFormatArgumentAdaptor<float>
+{
+	__always_inline static DebugFormatArgument construct(float value)
+	{
+		uintptr_t fudgedValue = 0;
+		memcpy(&fudgedValue, &value, sizeof(value));
```
rmn30, line 310: Could static assert that sizeof(value) <= sizeof(fudgedValue).

sdk/third_party/compiler-rt/README.md
```diff
@@ -0,0 +1,10 @@
+Soft-float support library
+==========================
+
+Files here are copied from compiler-rt, unmodified.
```
rmn30, line 4: We could make these file read only to protect them against careless editing.

tests/softfloat-test.cc
```diff
@@ -0,0 +1,57 @@
...
+		TEST_EQUAL(z, T(4), "Conversion result mismatch");
+	}
+
+	template<typename T>
+	void test()
+	{
+		test_op<T, std::plus>("+");
+		test_op<T, std::minus>("-");
+		test_op<T, std::multiplies>("*");
+		test_op<T, std::divides>("/");
```
rmn30, line 42: Could add comparison operators here? Do floats support any others?

sdk/lib/debug/debug.cc
```diff
@@ -101,6 +102,86 @@ namespace
...
+				write('.');
+				std::array<char, 10> buf;
+				const char           Digits[] = "0123456789";
+				auto                 s        = floatParts.decimal;
+				for (int i = static_cast<int>(buf.size() - 1); i >= 0; i--)
+				{
+					buf[static_cast<size_t>(i)] = Digits[s % 10];
+					s /= 10;
+				}
+				bool skipZero = true;
```
nwf, line 126: Set but unused?

sdk/lib/debug/debug.cc
```diff
@@ -101,6 +102,86 @@ namespace
...
+				for (size_t i = buf.size() - floatParts.decimalPlaces;
```
nwf, line 127: Hm.  This and the loop above should run the same number of iterations, but that's not obvious from their spellings... it's probably fine.

sdk/lib/stdio/xmake.lua
```diff
@@ -2,3 +2,15 @@ library("stdio")
   add_deps("string")
   set_default(false)
   add_files("printf.cc")
+  add_options("debug-print-doubles")
```
nwf, line 5: Given its use in `lib/stdio`, perhaps the option just should be `print-doubles` (and similarly for `float`)?

tests/softfloat-test.cc
```diff
@@ -0,0 +1,57 @@
+#define TEST_NAME "Softfloat"
+#include "stack_tests.h"
+#include "tests.hh"
+
+/**
+ * Tests for the softfloat integration.
+ *
+ * These tests are *not* checking that softfloat in any way conforms to IEE
```
nwf, line 8:
```suggestion
 * These tests are *not* checking that softfloat in any way conforms to IEEE
```

tests/stdio-test.cc
```diff
@@ -18,5 +18,8 @@ int test_stdio()
 	TEST_EQUAL(std::string_view(buffer), "-42", "snprintf(\"%d\", -42) failed");
 	sprintf(buffer, "%x", 6 * 9);
 	TEST_EQUAL(std::string_view(buffer), "36", "sprintf(\"%x\", 6 * 9) failed");
+	sprintf(buffer, "%f", 123.456);
+	TEST_EQUAL(
```
nwf, line 22: Make this conditional on the `debug-print-doubles` option?

sdk/third_party/compiler-rt/README.md
davidchisnall: Does that work reliably in git?  We don't do that for other vendored file from LLVM.

davidchisnall:
> Do we need to worry about the files in `compiler-rt` drifting WRT the compiler? (Could we have the compiler install them into its `$PREFIX`?)

No, these files implement a specification and the compiler relies on the specification.  The only coupling is that new functions get added to compiler-rt at around the same time clang / llvm expects them to exist.  clang works with libgcc* and compiler-rt works with gcc.  It was a stupid idea to put compiler-rt in the monorepo and I argued against it, but it worked well with Apple and Google's way of consuming LLVM.

sdk/lib/debug/debug.cc
rmn30: Removed, although it would be good to check this doesn't indicate an error. Seems to work as expected.

PR #522, opened 2025-04-26
tests: plug leaks

nwf:


davidchisnall (approved):
I wonder if each of the tests should end with a heap-free-all and if we should make the test runner do a revoke-and-empty-quarantine thing soon.

nwf:
FWIW, I made the test runner do `heap_quarantine_empty()` and `heap_render()` between each test, which made it relatively easy to figure out what was going on.

I think there are some points where a `heap_free_all` would be potentially damaging, like when we've allocated closures for `async` with the thread pool and it's caused the high-priority task to resume a test (to completion); we'll deallocate that closure only when the high priority thread sleeps again, and ripping it out from under the thread pool would be rude.  But perhaps everywhere else, yes.

PR #523, opened 2025-04-30
Switch to using riscv32cheriot as the default arch name when building.

resistor:
This will be needed to ensure that everything links properly once we
start setting EF_CHERIOT in ELF headers.


davidchisnall:
What's the oldest compiler version that supports this subarch?  13 or 17?

resistor:
> What's the oldest compiler version that supports this subarch? 13 or 17?



> What's the oldest compiler version that supports this subarch? 13 or 17?

It's present in 17 but not in 13.

davidchisnall (approved):

PR #524, opened 2025-04-30
Use -std=c23 instead of -std=c2x

resistor:


sdk/xmake.lua
```diff
@@ -163,7 +163,7 @@ toolchain("cheriot-clang")
 		toolchain:add("cxflags", default_flags, {force = true})
 		toolchain:add("cflags", default_flags)
 		toolchain:add("cxxflags", "-std=c++20")
-		toolchain:add("cflags", "-std=c2x", {force = true})
+		toolchain:add("cflags", "-std=c23", {force = true})
```
davidchisnall, line 166: We might be able to remove the `{ force = true }` bit here.  xmake knows this flag exists (sigh).

resistor: Done

PR #525, opened 2025-04-30
sdk/firmware.ldscript: introduce explicit PHDRs

nwf:
This captures what we were doing already -- the firmware images should be more or less unaltered except perhaps in choice of padding bytes -- but by forcing things into the same LOAD program header, the section ALIGN() directives ensure that the things we load are at least 4-byte aligned and sized.

This matters for OpenOCD talking to the Sonata debug unit, which can write only aligned multiples of four bytes to memories.

### DIFF
--- a/sdk/firmware.ldscript.in
+++ b/sdk/firmware.ldscript.in
@@ -3,11 +3,30 @@
 
 @mmio@
 
+PHDRS
+{
+	headers PT_PHDR PHDRS ;
+
+	# We define two loadable sections, for "low" and "high" material.  Depending
+	# on the board configuration, one of these holds "rocode" and the the other
+	# holds "rwdata".
+	loadlow PT_LOAD ;
+	loadhigh PT_LOAD ;
+}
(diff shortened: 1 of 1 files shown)


rmn30:
I think this looks reasonable, although I find linker scripts pretty hard to predict what they'll do. Have you tested uf2 loading as well as openocd?

rmn30 (approved):
Seems to work...

nwf:
Yes, UF2 loading works:
```
bootloader: Sonata system git SHA: 6cd0c7f00c4ac64f
bootloader: Selected software slot: 1
bootloader: Loading software from flash...
bootloader: Booting into program, hopefully.
Test runner: Build of d3dfb632 on May  1 2025
Test runner: Checking that rel-ro caprelocs work.  This will crash if they don't.  CHERI Permissions are:
Test runner: Global(0x0)
```

(`d3dfb632` is local but includes both this PR and #527.)

PR #527, opened 2025-05-01
Add xmake rule to define RTOS `git describe` and use it in the test suite

nwf:
There _surely must_ be a better way to do this, but being able to do it beats not at all.


tests/tests.hh
```diff
@@ -25,6 +25,8 @@ __cheri_compartment("debug_test") int test_debug_cxx();
 __cheri_compartment("debug_test") int test_debug_c();
 __cheri_compartment("unwind_cleanup_test") int test_unwind_cleanup();
 
+__cheri_compartment("test_runner") int print_version_information();
```
rmn30, line 28:
```suggestion
int print_version_information();
```

I don't think the compartment annotation is necessary here, unless you intend to export it as a compartment entry point?

rmn30 (approved):
Looks useful. I don't have any better idea for the `xmake`...

tests/tests.hh
nwf: Oh, indeed, just like `test_global_constructors()` below it.  Good catch; thanks!

PR #528, opened 2025-05-01
tests.extra/keep_ticking: a simple bringup test

nwf:


tests.extra/keep_ticking/top.cc
```diff
@@ -0,0 +1,33 @@
...
+void tick_forever(int tid, Ticks period)
+{
+	unsigned int cycles;
+
+	Debug::log("Entry {}...", tid);
+	while (true)
+	{
+		Timeout t{period};
+		int     res = thread_sleep(&t, ThreadSleepNoEarlyWake);
+		__asm__ volatile("csrr %0, mcycle" : "=r"(cycles));
```
rmn30, line 15: Maybe we should put this in a header file? We have `rdcycle64` in `sdk/include/riscvreg.h`, maybe add `rdcycle32`? There are identical inline asm statements in `benchmarks/timing.h` and `tests/test-runner.cc`.

PR #529, opened 2025-05-05
sdk/xmake: quote compartment names in generated ldscripts

nwf:


resistor:
Seems to work on my example, at least

davidchisnall (approved):

PR #531, opened 2025-05-07
sdk/xmake: conditionally link the allocator

nwf:
This is a long-standing todo, but it's helpful for shrinking some of our `tests.extra` unit tests, even if the full test suite would absolutely not be happy.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -100,7 +100,7 @@ jobs:
       run: |
         set -e
         cd $PWD/examples/03.hello_safe_compartment/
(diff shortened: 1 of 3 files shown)


nwf:
I have a follow-up on my desk that also removes the allocator hazard epoch and area shared objects, but they're generally small and it's a little invasive to (conditionally) remove them (and, in principle, someone might be disabling _our_ allocator to bring their own; that's not tested, but maybe it'd work).

davidchisnall:
Now that we’re using the merge queue, we don’t have the squash option. Please can you tidy up the commit history and then we can merge.

PR #533, opened 2025-05-12
DNM: tests: make each test family optional

nwf:
Make each test's presence configurable via xmake.  By default, all tests
will be included and some can be notched out with one or more
`--test-${test}=n` options.  Use `--all-tests=n` to flip the default,
generating a test suite binary including only those particular tests
specified with `--test-${test}=y`.  Use `xmake f --help` to see the list
of `${test}`s.

Perhaps hold off on merging until there's some better guidance from xmake upstream about `option()` handling; see https://github.com/xmake-io/xmake/issues/6419 .

Review of the general approach is, however, quite welcome.


tests/test-runner.cc
```diff
@@ -58,8 +58,37 @@ namespace
 			debug_log("{} finished in {} cycles", msg, cycles - startCycles);
 		}
 	}
+
+#ifdef ENABLED_TESTS
```
davidchisnall, line 62:
```suggestion
#ifndef ENABLED_TESTS
#define ENABLED_TESTS
#endif
```

tests/test-runner.cc
```diff
@@ -58,8 +58,37 @@ namespace
...
+		}
+
+		return false;
+	}
+#else
+	static constexpr bool test_enabled(const char *)
+	{
+		return true;
+	}
+#endif
```
davidchisnall, line 82:
```suggestion
```

nwf:
Hold merge until #236 lands.  Modularizing the test suite like this will let us be able to run the bigdata tests in CI and not just when we remember to uncomment the test.

(That said, I'd like to land this soon, since it conflicts with anything that adds new test families, like #558.)

tests/xmake.lua
```diff
@@ -39,102 +92,202 @@ freertos_compile_test("recursive-mutex-only", {"CHERIOT_EXPOSE_FREERTOS_RECURSIV
...
+    add_deps("stdio")
+
+test("bigdata", { not_most = true, conflicts = { "allocator" } })
+
 -- Test the static sealing types
-test("static_sealing")
 compartment("static_sealing_inner")
-	add_files("static_sealing_inner.cc")
+    add_files("static_sealing_inner.cc")
+    set_default(false)
```
rmn30, line 134: Why is this needed?

rmn30 (approved):
Looks OK to me. Rather a lot of xmake magic, but if it works...

PR #535, opened 2025-05-12
sdk/xmake: JSON path array traversal and append

nwf:


rmn30 (approved):
I guess lua doesn't have an equivalent of Option Base 0?!

PR #537, opened 2025-05-14
xmake: further refactoring board file management

nwf:


nwf:
See https://github.com/xmake-io/xmake/issues/6449 ; I'd rather not need all the various single-line `.in` that this approach seems to be heading towards.

nwf:
Different take that doesn't lean on `xmake` `configfiles`: those get generated for every loaded target, regardless of the dependency structure (or something like that, anyway).  Since we might want to be more choosy for some targets, generate things by hand.

PR #544, opened 2025-06-08
Use the new support for aliases in the toolchain to make memmove an alias to memcpy.

resistor:
NOTE: This needs a devcontainer release to happen containing the alias support before it can be landed.


resistor:
Updated with the macro check, ready for review now

resistor:
Ping

PR #547, opened 2025-06-25
Shared objects might want LoadGlobal permission

nwf:
Changes `sdk/core/loader/types.h` to introduce a PermitLoadGlobal bit on import entries and teaches `sdk/include/compartment-macros.h` to use it.  The rest of the changes are catching up core components and the test suite (which is expanded a bit to test the new bit).

This will not pass CI because the compiler currently mismatches the (existing) loader, in that it only masked out the top 4, not 8, bits of the length for types.  Local testing with https://github.com/CHERIoT-Platform/llvm-project/pull/180 suggests that at least that is simple to fix.

This surely conflicts with #545 and, unless @davidchisnall thinks I'm about to blow a hole in the security model, will require revisions to https://github.com/CHERIoT-Platform/llvm-project/pull/147 .

Tagging @xdoardo who, for some reason, isn't eligible as a reviewer.

### DIFF
--- a/sdk/core/allocator/alloc.h
+++ b/sdk/core/allocator/alloc.h
@@ -1034,7 +1034,8 @@ class MState
 	 */
 	[[nodiscard]] __always_inline auto hazard_list_begin()
 	{
-		auto    *lockWord{SHARED_OBJECT(uint32_t, allocator_epoch)};
+		auto    *lockWord{SHARED_OBJECT_WITH_DATA_PERMISSIONS(
+          uint32_t, allocator_epoch, true, true)};
 		uint32_t epoch = *lockWord >> 16;
 		Debug::Invariant(
 		  (epoch & 1) == 0,
@@ -1284,9 +1285,14 @@ class MState
 	bool hazard_pointer_check(Capability<void> allocation)
 	{
 		// It is now safe to walk the hazard list.
-		Capability<void *> hazards =
-		  const_cast<void **>(SHARED_OBJECT_WITH_PERMISSIONS(
-		    void *, allocator_hazard_pointers, true, false, true, false));
+		Capability<void *> hazards = const_cast<void **>(
+		  SHARED_OBJECT_WITH_PERMISSIONS(void *,
+		                                 allocator_hazard_pointers,
+		                                 true,
+		                                 false,
+		                                 true,
+		                                 false,
+		                                 false));
 		size_t pointers = hazards.length() / sizeof(void *);
 		for (size_t i = 0; i < pointers; i++)
 		{
--- a/sdk/core/allocator/main.cc
+++ b/sdk/core/allocator/main.cc
@@ -100,9 +100,13 @@ namespace
 		Capability m{tbase.cast<MState>()};
 
 		size_t hazardQuarantineSize =
-		  Capability{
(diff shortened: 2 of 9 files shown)


davidchisnall:
> I wasn't sure if the current behavior was deliberate, since the RTOS test suite carefully enumerates all the permissions expected on MMIO and shared object references

Just because I wrote some code and wrote some tests for the code doesn't mean that the code does what I meant.  I think this is a good counterexample.

> We can't express those, but SD MC !LM !LG might mitigate some of the gap?

I think that's fine as a thing you might want to pass to DMA, the question is whether that's a desirable set of initial permissions for a pre-shared object.  The existing permissions have some use cases:

 - Remove C means that we know that all information leakage must be direct.  We do this in the time shared object so that we know that the only information leak can be from the SNTP compartment *to* others, which means that there isn't a data flow (modulo timing covert channels) from time consumers to the network via NTP.
 - Removing W means that you can't write to it.  If you have C, then removing m means that it's deeply immutable.  In SNTP we make the shared object writeable only from the SNTP compartment and so we can ensure one-way dataflow. 
 - Removing R means that you have write-only access, possibly with capabilities.

I'm not sure what removing g means (it hasn't come up yet because we don't really use C for any normal uses - the allocator's hazard pointers are a special case with the write end mediated via the switcher).  You can load pointers from the global, but you can't capture them.  But that doesn't bound the time that you can hold them for in any way (a thread can load one, store it on its stack, and leave it there forever).  The closest thing I can come to a use case is something where you have a broadcast work-queue-like thing and you want everyone do to an atomic exchange to remove items and then process them.  Removing G and g after you've swapped a value out seems like a better idea there though.


All of that said:

If we make g implicit, *removing* it later is harder.  This change doesn't change the semantics of any *existing* mappings that have reduced permissions, which is nice.  And that makes me ponder if we should do it even if we don't know why yet.

(Also, sorry, I wrote the previous comment before noticing the LLVM commit)

davidchisnall:
Changing the number of arguments to the macros is an API-breaking change.  I think we can avoid that by making it variadic, and using `__VA_OPT__` to provide a default.

nwf:
To be more detailed: the C preprocessor crimes in the first commit move _all_ the permissions parameters to `__VA_ARGS__` and uses that to initialize a structure with a `bool` field for each permission.  Because we need this to work in C and C++, we have to rely on common structure initialization semantics.  Fortunately this all works out the way we want in both specifications:

- C++ permits `struct`s to be built from initializer lists with fewer elements than the structure (including empty initializer lists), list entries are interpreted as each sequential field, and elements not explicitly initialized in the list will be "copy-initialized from an empty initializer list", which amounts to zero initialization, and for the scalar `bool` type, that means the explicit conversion of  `0`, so `false`.  (At least, all of that's true for C++11 and later.)

- C permits `struct`s to be built from (short) initializer lists (C23 even permits empty initializer lists, at long last), list entries are again interpreted as each sequential field, and elements not explicitly initialized are "empty initialized"; as we `typedef` `bool` to be C99's `_Bool` in `<stdbool.h>`, this means these (integral) fields initialize to `0`, so `false`.

I decided to go down this path because it means the macros are future compatible with _some further_ additional permission flag, in ways that an alternative like `(false __VA_OPT__(|) __VA_ARGS__) ? ... : ...` would not be.

PR #554, opened 2025-07-10
compile_flags.txt: use std=c++23

nwf:
Follow-up to #551.

### DIFF
--- a/compile_flags.txt
+++ b/compile_flags.txt
@@ -1,5 +1,5 @@
 -xc++-header
--std=c++20
+-std=c++23
 -Qunused-arguments
 -target
 riscv32-unknown-unknown


nwf:
I am deeply confused... the CI `clang-tidy` failures are a consequence of having generated the `compile_commands.json` files in the prior step.  [The clang-tidy docs](https://clang.llvm.org/extra/clang-tidy/) mention `compile_commands.json` only in the context of being passed a `-p` argument, which we are not, and moreover the failures occur for files that aren't mentioned in the generated JSON.

This smells like a `clang-tidy` upstream bug.  @resistor?

resistor:
I don't think this change is needed. `clang-tidy` reads from the `compile_commands.json` by default (`-p` is just for overriding where it finds it). I tried regenerating my `compile_commands.json`, verifying that it is correctly specifying `-std=c++23`, and then verified that running the `clang-tidy` script produces no diff.

davidchisnall:
I would also expect specifying `-std=` there to break things because we run clang-tidy on some C files and it will try to apply a C++ dialect flag to them and fail.

PR #557, opened 2025-07-15
Various fixes to the allocator and the allocator tests.

hlef:
- Demonstrate a race in the allocator between claims and ephemeral claims, and how this race may lead the allocator to ignore a claim and make quota disappear into the void. Fix this by checking the number of claims on an object when removing it from the hazard quarantine (surprisingly small patch, hopefully I got it right!). This corner-case was reported by Reto Achermann (UBC) as we are working on formally verifying the allocator.
- Improve allocator tests to make them more readable and expressive.
- The tests have multiple instances of using `is_valid` to check tags when they should be using something not subject to CSE (here be dragons). I made the mistake myself when updating the tests. Address this by adding a C++ wrapper `is_valid_temporal` for `__builtin_cheri_tag_get_temporal` and using it where appropriate in the tests. Make sure the behavior w.r.t. CSE is well-documented.

Entirely artisanal, no AI involved. All hallucinations are mine.

### DIFF
--- a/sdk/core/allocator/alloc.h
+++ b/sdk/core/allocator/alloc.h
@@ -1321,11 +1321,24 @@ class MState
 					foundSkippedValue = true;
 					continue;
 				}
-				Debug::log("Found safe-to-free object in hazard list: {}", ptr);
 				// Re-derive a pointer to the chunk.
 				Capability heap{heapStart};
 				heap.address() = ptr.address();
 				auto chunk     = MChunkHeader::from_body(heap);
+				if (chunk->claims > 0)
+				{
+					/*
+					 * The chunk was freed but ended up in
+					 * the hazard quarantine due to an
+					 * ephemeral claim. Then it was
+					 * claimed, and so we cannot free it
+					 * anymore. However we can remove it
+					 * from the hazard list since
+					 * `!hazard_pointer_check(ptr)` above.
+					 */
+					continue;
+				}
+				Debug::log("Found safe-to-free object in hazard list: {}", ptr);
 				// We know this isn't in the hazard lists (we just checked!) so
 				// free it without doing any hazard pointer checks checks.
 				mspace_free(*chunk, ptr.length(), true);
--- a/tests/allocator-test.cc
+++ b/tests/allocator-test.cc
@@ -596,10 +596,15 @@ namespace
 		debug_log("Before allocating, quota left: {}",
 		          heap_quota_remaining(SECOND_HEAP));
 		Timeout longTimeout{1000};
-		void   *ptr = heap_allocate(&longTimeout, SECOND_HEAP, 16);
-		TEST(__builtin_cheri_tag_get(ptr), "Failed to allocate 16 bytes");
-		void *ptr2 = heap_allocate(&longTimeout, SECOND_HEAP, 16);
-		TEST(__builtin_cheri_tag_get(ptr2), "Failed to allocate 16 bytes");
+		size_t  allocSize = 16;
+		void   *ptr       = heap_allocate(&longTimeout, SECOND_HEAP, allocSize);
+		TEST(__builtin_cheri_tag_get(ptr),
+		     "Failed to allocate {} bytes",
+		     allocSize);
+		void *ptr2 = heap_allocate(&longTimeout, SECOND_HEAP, allocSize);
+		TEST(__builtin_cheri_tag_get(ptr2),
+		     "Failed to allocate {} bytes",
+		     allocSize);
 		debug_log("After allocating, quota left: {}",
 		          heap_quota_remaining(SECOND_HEAP));
 		static cheriot::atomic<int> state = 0;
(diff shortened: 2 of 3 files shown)


tests/allocator-test.cc
```diff
@@ -637,29 +633,34 @@ namespace
...
+		// chunks are still valid allocations) and not affect our
+		// quota.
 		TEST_EQUAL(heap_free(SECOND_HEAP, ptr),
 		           -EPERM,
-		           "Attempt to free freed but hazarded pointer not EPERM");
-		// Sleep again to make sure that the lambda from our async is gone.
-		// The logs may make it take more than one quantum in debug builds.
-		// The next test requires all memory allocated from the malloc
-		// capability to be freed before it starts.
+		           "Attempt to free freed but hazarded pointer {} not EPERM");
```
nwf: Missing argument for `{}`?

tests/allocator-test.cc
```diff
@@ -637,29 +633,34 @@ namespace
...
+		auto quotaLeft = heap_quota_remaining(SECOND_HEAP);
+		TEST(quotaLeft == SECOND_HEAP_QUOTA,
+		     "After alloc and free from {}-byte quota, {} bytes left",
+		     SECOND_HEAP_QUOTA,
+		     quotaLeft);
+		// Stop the async and sleep until it is gone. When the async is
+		// freed, the ephemeral claims are released and the allocator
```
nwf: In more detail, the ephemeral claims are released by the cross-compartment call _to `token_obj_destroy`_ inside `thread_pool.h`'s `detail::wrap_callback_lambda`.  That is the ephemeral claims are released _when_ the allocator is called.

Aside: can you stick a `yield()` up in the async's spin loop?  `yield()` isn't a cross-compartment call (but rather a RISC-V `ecall`, which "looks enough like" an interrupt that it's easy to conflate with preemptive context switching) and so won't drop the ephemeral claims.

tests/allocator-test.cc
```diff
@@ -644,11 +649,19 @@ namespace
...
 		     SECOND_HEAP_QUOTA,
 		     quotaLeft);
+		// Try to claim the buffer. This should succeed since there is
+		// still an ephemeral claim on it.
+		ssize_t claimSize = heap_claim(MALLOC_CAPABILITY, ptr2);
+		TEST((allocSize <= claimSize) &&
+		       (claimSize <= allocSize + CHERIOTHeapMinChunkSize),
+		     "{}-byte allocation claimed as {} bytes",
+		     allocSize,
+		     claimSize);
```
nwf, line 662: Hm.  I guess we don't (here, in the return of `heap_claim`) account for the size of the claim object that we (might, in general, but, in this case, surely) allocate.  Someone paying attention to their quota remaining size might notice the discrepancy.

sdk/core/allocator/alloc.h
```diff
@@ -1321,11 +1321,22 @@ class MState
 					foundSkippedValue = true;
 					continue;
 				}
-				Debug::log("Found safe-to-free object in hazard list: {}", ptr);
 				// Re-derive a pointer to the chunk.
 				Capability heap{heapStart};
 				heap.address() = ptr.address();
 				auto chunk     = MChunkHeader::from_body(heap);
+				if (chunk->claims > 0)
```
nwf, line 1328: Commenting here since I can't on the commit message itself.  This bug is doubly bad: not only do we free an object we said was claimed, we leak any associated claim record object(s), since there would, after free, be no initial pointer to the list thereof.

To ensure that bisection of commit history is easy, could you flip the order of this and the prior commit (which tests this bug)?  I believe you that it's real.

sdk/core/allocator/alloc.h
```diff
@@ -1321,11 +1321,22 @@ class MState
...
+				{
+					/*
+					 * The chunk got freed but ended up in
+					 * the hazard quarantine due to an
+					 * ephemeral claim. Then it was
+					 * claimed, and so we cannot free it
+					 * anymore.
+					 */
```
nwf, line 1338: But we can still drop it from the _hazards_ list, because `!heap_pointer_check()`, above.

nwf (changes requested):
Terrific work and great find!  Very minor nits.

I wonder if it's worth expanding the allocator _fuzz_ test to...
1. Use some async() jobs to take ephemeral claims...
2. Take full claims instead of allocations, some fraction of the time, during operation (multiply-listing claimed objects in `allocations`)
3. Ensure that any decrease in is quota is accounted for by the allocated objects in its collection (the sum of allocated sizes may appear larger than the decrease in quota, due to claims, but a quota leak would, WHP, eventually be revealed).

sdk/core/allocator/alloc.h
hlef: Right! That being said, this code is already dropping the object from the hazard list by *not* re-adding it (we don't do `hazardQuarantine[insert++] = ptr;`), doesn't it?

Would you like me to update the comment to make that explicit?

sdk/core/allocator/alloc.h
hlef: I did not realize this! Indeed we also leak the claim object. I updated the commit messages as well to clarify.

tests/allocator-test.cc
hlef: Ah, right! I took that from somewhere else in the test. Might update there as well for clarity.

hlef: Also, this test will be annoying if we want to use the test suite on the formally verified Rust implementation at some point.

hlef:
@nwf All comments should be fixed now!

nwf (approved):
LGTM!

PR #558, opened 2025-07-21
Add bitpacks.hh, a proxy-based wrapper for packed bit-fields

nwf:
Introduce a `Bitpack` class, representing a packed bit-field within an unsigned integral type.  `Bitpack`s are intended for use instead of C++ bit-fields or manual bit twiddling and should...

- not have any undefined or implementation-defined behavior or layout (besides endianness); `Field` definitions are interpreted as one might expect using `<<`, `|`, `&`, and `~` (see `Bitpack<>::Field<>::raw_view` and `raw_with` for the core of all this C++ syntax);

- readily serve as superclasses for classes fronting particular hardware control registers or other bit-packed types;

- work in `constexpr` context and in `volatile`-qualified forms;

- be reasonably ergonomic to use in as many contexts as possible.

The fundamental perspective is one of `Field::View` proxies to contiguous spans of bits within a `Bitpack<T>`'s underlying storage `T`.  Said spans are defined using `FieldInfo` structures.  Field proxies to fields of type `F` and with `FieldInfo I` may be explicitly constructed with `.view<F, I>()`, but there is also some convenience machinery for defining `Bitpack`s as _products_ of differently-typed fields, such that a proxy to a given field with type `F` can be accessed uniformly as `.view<F>()` (where, effectively, the `FieldInfo` `I` is computed from `F`).

There are also some gross C-preprocessor macros that aim to de-clutter use of `Bitpack`s, especially in combination with `enum class`es.

The raw implementation is in `sdk/include/bitpacks.hh` and a small smattering of tests (mostly checking operation in combination with `constexpr` and `volatile`) and demonstration of use is available in `tests/bitpacks-test.cc`.  It is probably sensible to review the two in parallel, as the implementation is probably confusing without examples of use, and the use is probably confusing without some understanding of the implementation.

This is still a little bit of an experiment and a request for comments / opinions, so please, fire away!

### DIFF
--- a/sdk/include/platform/sunburst/platform-uart.hh
+++ b/sdk/include/platform/sunburst/platform-uart.hh
@@ -4,6 +4,7 @@
 #	define DEFAULT_UART_BAUD_RATE 921'600
 #endif
 
+#include <bitpack.hh>
 #include <platform/concepts/uart.hh>
 #include <utils.hh>
 
@@ -18,30 +19,72 @@
  */
 struct OpenTitanUart : private utils::NoCopyNoMove
 {
-	/**
-	 * Interrupt State Register.
-	 */
-	uint32_t interruptState;
-	/**
-	 * Interrupt Enable Register.
-	 */
-	uint32_t interruptEnable;
-	/**
-	 * Interrupt Test Register.
-	 */
-	uint32_t interruptTest;
+	struct Interrupts : Bitpack<uint32_t>
+	{
+		BITPACK_USUAL_PREFIX;
+		BITPACK_MEMBER_ADD_ENUM_BOOL_CLEARED_ASSERTED(TransmitWatermark, 0);
+		BITPACK_MEMBER_ADD_ENUM_BOOL_CLEARED_ASSERTED(ReceiveWatermark, 1);
+		BITPACK_MEMBER_ADD_ENUM_BOOL_CLEARED_ASSERTED(TransmitDone, 2);
+		BITPACK_MEMBER_ADD_ENUM_BOOL_CLEARED_ASSERTED(ReceiveOverflowError, 3);
+		BITPACK_MEMBER_ADD_ENUM_BOOL_CLEARED_ASSERTED(ReceiveFrameError, 4);
+		BITPACK_MEMBER_ADD_ENUM_BOOL_CLEARED_ASSERTED(ReceiveBreakError, 5);
+		BITPACK_MEMBER_ADD_ENUM_BOOL_CLEARED_ASSERTED(ReceiveTimeout, 6);
+		BITPACK_MEMBER_ADD_ENUM_BOOL_CLEARED_ASSERTED(ReceiveParityError, 7);
+		BITPACK_MEMBER_ADD_ENUM_BOOL_CLEARED_ASSERTED(TransmitEmpty, 8);
+	};
+
+	struct InterruptState : BitpackDerived<Interrupts>
+	{
+		BITPACK_DERIVED_PREFIX;
+
+		// TransmitWatermark is a status, not an event, so RO, not RW1C.
+		BITPACK_DERIVED_FIELD_CONST_FOR_TYPE(TransmitWatermark, true);
+
+		// ReceiveWatermark is a status, not an event, so RO, not RW1C.
+		BITPACK_DERIVED_FIELD_CONST_FOR_TYPE(ReceiveWatermark, true);
+
+		// TransmitEmpty is a status, not an event, so RO, not RW1C.
+		BITPACK_DERIVED_FIELD_CONST_FOR_TYPE(TransmitEmpty, true);
+	};
+
+	InterruptState interruptState;
+	Interrupts     interruptEnable;
+	Interrupts     interruptTest;
+
 	/**
 	 * Alert Test Register (unused).
 	 */
 	uint32_t alertTest;
-	/**
-	 * Control Register.
-	 */
-	uint32_t control;
-	/**
-	 * Status Register.
-	 */
-	uint32_t status;
+
+	struct Control : Bitpack<uint32_t>
+	{
+		BITPACK_USUAL_PREFIX;
+		BITPACK_MEMBER_ADD_ENUM_BOOL_DISABLED_ENABLED(Transmit, 0);
+		BITPACK_MEMBER_ADD_ENUM_BOOL_DISABLED_ENABLED(Receive, 1);
+		BITPACK_MEMBER_ADD_ENUM_BOOL_DISABLED_ENABLED(NoiseFilter, 2);
+		BITPACK_MEMBER_ADD_ENUM_BOOL_DISABLED_ENABLED(SystemLoopback, 3);
+		BITPACK_MEMBER_ADD_ENUM_BOOL_DISABLED_ENABLED(LineLoopback, 4);
+		BITPACK_MEMBER_ADD_ENUM(Parity, uint8_t, 5, 6) {
+			None = 0b00,
+			Even = 0b01,
+			Odd  = 0b11,
+		};
+
+		BITPACK_MEMBER_ADD_NUMERIC(BreakLevel, uint8_t, 8, 9);
+		BITPACK_MEMBER_ADD_NUMERIC(Nco, uint16_t, 16, 31);
+	} control;
+
+	const struct Status : Bitpack<uint32_t>
+	{
+		BITPACK_USUAL_PREFIX;
+		BITPACK_MEMBER_ADD_ENUM_BOOL_CLEARED_ASSERTED(TransmitFull, 0);
+		BITPACK_MEMBER_ADD_ENUM_BOOL_CLEARED_ASSERTED(ReceiveFull, 1);
+		BITPACK_MEMBER_ADD_ENUM_BOOL_CLEARED_ASSERTED(TransmitEmpty, 2);
+		BITPACK_MEMBER_ADD_ENUM_BOOL_CLEARED_ASSERTED(TransmitIdle, 3);
+		BITPACK_MEMBER_ADD_ENUM_BOOL_CLEARED_ASSERTED(ReceiveIdle, 4);
+		BITPACK_MEMBER_ADD_ENUM_BOOL_CLEARED_ASSERTED(ReceiveEmpty, 5);
+	} status; // NOLINT(readability-identifier-naming)
+
 	/**
 	 * UART Read Data.
 	 */
@@ -50,14 +93,43 @@ struct OpenTitanUart : private utils::NoCopyNoMove
 	 * UART Write Data.
 	 */
 	uint32_t writeData;
-	/**
-	 * UART FIFO Control Register.
-	 */
-	uint32_t fifoCtrl;
+
+	struct FIFOControl : Bitpack<uint32_t>
+	{
+		BITPACK_USUAL_PREFIX;
+
+		BITPACK_MEMBER_ADD_ENUM_BOOL(Receive, AsIs, Reset, 0);
+		BITPACK_MEMBER_ADD_ENUM_BOOL(Transmit, AsIs, Reset, 1);
+
+		BITPACK_MEMBER_ADD_ENUM(TransmitWatermark, uint8_t, 2, 4) {
+			Level1  = 0b000,
+			Level2  = 0b001,
+			Level4  = 0b010,
+			Level8  = 0b011,
+			Level16 = 0b100,
+			Level32 = 0b101,
+			Level64 = 0b110,
+		};
+		BITPACK_MEMBER_ADD_ENUM(ReceiveWatermark, uint8_t, 5, 7) {
+			Level1  = 0b000,
+			Level2  = 0b001,
+			Level4  = 0b010,
+			Level8  = 0b011,
+			Level16 = 0b100,
+		};
+	} fifoControl;
+
 	/**
 	 * UART FIFO Status Register.
 	 */
-	uint32_t fifoStatus;
+	const struct FIFOStatus : Bitpack<uint32_t>
+	{
+		BITPACK_USUAL_PREFIX;
+
+		BITPACK_MEMBER_ADD_NUMERIC(TransmitLevel, uint8_t, 0, 7);
+		BITPACK_MEMBER_ADD_NUMERIC(ReceiveLevel, uint8_t, 16, 23);
+	} fifoStatus; // NOLINT(readability-identifier-naming)
+
 	/**
 	 * Transmit Pin Override Control.
 	 *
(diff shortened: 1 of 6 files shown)


nwf:
Some further work from use in anger.  Notably, adds `TQTYPE` helpers for using dependently-typed bitpacks, the `BITPACK_NUMERIC_..._VALUE` macros, and some additional `.raw()` accessors to elide some `static_cast<>`s in user code.

nwf:
@davidchisnall has asked that these be marked "experimental" in a way that the compiler understands -- probably with a `deprecated` attribute (or several), since there isn't an `experimental` annotation.  I agree -- this has been used in anger a little bit within SCI, but not enough that I want to promise we won't make breaking changes.  The question then is _what_ should we so mark?

Should we mark the `bitpacks.hh` header itself?  That's probably excessive, but it does mean we have at most one warning per compilation unit, which might be nice.

If we mark the `Bitpack` class itself, this will highlight all the definitions that declare things of `Bitpack` type, which probably specifically means "just the BSP headers" at the moment; the use sites are probably left un-warned-about.

Should we mark the `.read()`, `.get()`, `.with()`, `.set()`, and `.alter()` methods?  That probably gets most of the use sites (especially because most of the `Bitpack &`-s user code is likely to see are `volatile Bitpack &`-s and we're gently nudging the use of those to be through explicit methods).

Should we mark the C preprocessor macros?  That's going to be extremely chatty in some user code and quiet in others, depending on user's aesthetic decisions.

Should we have some `#if !defined()`-s around the marks to allow them to be suppressed with something like `-DNO_WARN_CHERIOT_EXPERIMENTAL`?

sdk/include/platform/sunburst/platform-uart.hh
```diff
@@ -267,22 +286,22 @@ struct OpenTitanUart : private utils::NoCopyNoMove
...
 	[[gnu::always_inline]] uint16_t receive_fifo_level() volatile
 	{
-		return ((fifoStatus >> 16) & 0xff);
+		return BITPACK_BY_QTYPE(fifoStatus.read(), ReceiveLevel).raw();
 	}
 
 	bool can_write() volatile
 	{
-		return !(status & StatusTransmitFull);
+		return BITPACK_ENUM_QTYPE_OP(status.read(), TransmitFull, ==, Cleared);
```
rmn30: I wish there was an easier way to write things like this. I can easily understand and write the prior version providing I know the definition of `StatusTransmitFull`. Now that I've seen how it's done I could probably write the bit pack version but I have to learn that and it's still not exactly pretty (sorry). I guess that's the price of type safety?

sdk/include/platform/sunburst/platform-uart.hh
```diff
@@ -18,30 +19,89 @@
...
+		BITPACK_DEFINE_TYPE_ENUM_BOOL_CLEARED_ASSERTED(TransmitDone, 2);
+		BITPACK_DEFINE_TYPE_ENUM_BOOL_CLEARED_ASSERTED(ReceiveOverflowError, 3);
+		BITPACK_DEFINE_TYPE_ENUM_BOOL_CLEARED_ASSERTED(ReceiveFrameError, 4);
+		BITPACK_DEFINE_TYPE_ENUM_BOOL_CLEARED_ASSERTED(ReceiveBreakError, 5);
+		BITPACK_DEFINE_TYPE_ENUM_BOOL_CLEARED_ASSERTED(ReceiveTimeout, 6);
+		BITPACK_DEFINE_TYPE_ENUM_BOOL_CLEARED_ASSERTED(ReceiveParityError, 7);
+		BITPACK_DEFINE_TYPE_ENUM_BOOL_CLEARED_ASSERTED(TransmitEmpty, 8);
+	};
+
+	struct InterruptState : public Interrupts
```
venngar: It is missing TransmitEmpty field to be set as read-only (reference to OpenTitan UART docs: https://opentitan.org/book/hw/ip/uart/doc/registers.html#intr_state)

sdk/include/platform/sunburst/platform-uart.hh
```diff
@@ -50,14 +110,43 @@ struct OpenTitanUart : private utils::NoCopyNoMove
...
+			Level8  = 0b011,
+			Level16 = 0b100,
+		};
+	} fifoControl;
+
 	/**
 	 * UART FIFO Status Register.
 	 */
-	uint32_t fifoStatus;
+	struct FIFOStatus : Bitpack<uint32_t>
```
venngar: I think those are supposed to be read-only too (https://opentitan.org/book/hw/ip/uart/doc/registers.html#fifo_status), therefore there should be additional `true` argument as a last one in both BITPACK macro calls.

nwf: FWIW, if the whole word is read-only, it also suffices to make it `const`.  Then the bitpack will act like every field inside it has `isConst` set to true.

nwf:
Further changes from use in anger.

- Better doxygen
- Bitpack's from-`Storage` constructor has been made `explicit`; see commentary.
- The `_ASSIGN` macros have been removed in favor of the shorter `_OP` with `=` forms.
- Macros are now more versatile and uniformly named: `BITPACK_{QUAL,WRAP,ENUM}{,_QTYPE,_TQTYPE,_UTYPE}_{OP,WITH}` and `BITPACK_{U,Q,TQ}VAL_{OP,WITH}`.

sdk/include/platform/sunburst/platform-uart.hh
nwf: I'm not sure whether `BITPACK_QVAL_OP(status.read(), ==, TransmitFull::Cleared)` is any better or if it's even more confusing.  On the one hand, shorter and more obviously "that value" is involved... and on the other "what about the others" (the compared value from the bitpack is the value _of the view projection of the same type_ as the given value).

nwf:
There was a typo in the existing `sunburst/platform-uart.hh` that suggested the clock divisor was 8, not 16, bits, but the code in fact was writing 16 bits.  I'd defined the field analogously incorrectly but that's now fixed; as of the most recent push, the bitpack-ified version generates identical code in the RTOS test suite on Sonata.

sdk/include/platform/sunburst/platform-uart.hh
```diff
@@ -18,30 +19,98 @@
...
+	{
+		using Interrupts::Interrupts;
+		using Interrupts::operator=;
+
+		// Inherit field info for most types from parent bitpack
+		template<typename FieldType>
+		static constexpr FieldInfo field_info_for_type()
+		{
+			return Interrupts::field_info_for_type<FieldType>();
+		}
```
rmn30: We could consider adding  macro for this?

sdk/include/platform/sunburst/platform-uart.hh
```diff
@@ -50,14 +119,43 @@ struct OpenTitanUart : private utils::NoCopyNoMove
...
+			Level32 = 0b101,
+			Level64 = 0b110,
+		};
+		BITPACK_DEFINE_TYPE_ENUM_CLASS(ReceiveWatermark, uint8_t, 5, 7) {
+			Level1  = 0b000,
+			Level2  = 0b001,
+			Level4  = 0b010,
+			Level8  = 0b011,
+			Level16 = 0b100,
+		};
```
rmn30, line 119: Picky, but I'd prefer a log2 field type to an enum here...

sdk/include/platform/sunburst/platform-uart.hh
```diff
@@ -194,16 +193,21 @@ struct OpenTitanUart : private utils::NoCopyNoMove
...
 	/// Clears the contents of the receive and transmit FIFOs.
 	void fifos_clear() volatile
 	{
-		fifoCtrl =
-		  fifoCtrl | FifoControlTransmitReset | FifoControlReceiveReset;
+		fifoControl.alter([](auto v) {
+			BITPACK_QVAL_OP(v, =, Transmit::Reset);
+			BITPACK_QVAL_OP(v, =, Receive::Reset);
+			return v;
+		});
```
rmn30, line 184: Took me a moment to figure out why the first lambda uses `[=]` here and the second just `[]`, but it's obvious now. Also why does the first need `TQVAL_OP` and the second just `QVAL_OP`? I understand that TQVAL is for templated types but understanding stuff like this is the biggest hurdle when using bitpacks. The compiler is not very helpful for stuff like this so we need lots of good examples and clear documentation.

sdk/include/platform/sunburst/platform-uart.hh
```diff
@@ -262,27 +292,27 @@ struct OpenTitanUart : private utils::NoCopyNoMove
 	 */
 	void reset() volatile
 	{
-		control = 0;
+		control = {};
```
rmn30: Could this be `{0}`? That would be more explicit as it's not obvious what this does otherwise.

sdk/include/bitpacks.hh
```diff
@@ -0,0 +1,1106 @@
+#pragma once
+
+/**
+ * \defgroup bitpacks Bitpacks
+ *
+ * `Bitpack`-s are bitfield-esque composite structures within numeric words.
+ * They are, roughly, meant to serve as lenses through which one views the bits
+ * of an underlying numeric word as as a product of (typed) fields, each
+ * occuping a contiguous span of bits within that word.
```
rmn30, line 9: This is a very dense opening paragraph! I think we need to explain this in simpler terms with a concrete motivating use case (e.g. MMIO registers). How about:

`Bitpack`-s are designed to provide an easy-to-use and type-safe way to handle situations where many fields and flags are packed into an underlying integer (e.g. a `uint32_t`) at arbitrary bit alignments. This is similar to the `bitfield` type in `C` except more portable and type safe. They are particularly useful for handling memory mapped I/O registers for hardware drivers. Supported features include:

 - arbitrary bit width numeric fields within an underlying integer
 - type safe enumeration and flags within an underlying integer
 - `const` fields e.g. read-only by software but writable by hardware
 - making handling of `volatile` explicit but easy e.g. for read-modify write on an MMIO register
 - possible to reuse field types between bit pack types.

Following up with some concrete examples would then be really useful.

rmn30 (approved):
Having finally actually used these a little I think they're really useful and we but we should add some good documentation mainly in the form of motivated examples, especially for when to use the different helper macros.

sdk/include/platform/sunburst/platform-uart.hh
nwf: Yeah.  I did it this way because [the official docs](https://opentitan.org/book/hw/ip/uart/doc/registers.html) don't promise that it's actually log2.

sdk/include/platform/sunburst/platform-uart.hh
nwf: For reasons beyond my understanding, `decltype(v)::Transmit::Reset` works but `decltype(v)::SystemLoopback{systemLoopback}` doesn't unless it's qualified with `typename`. :(

sdk/include/platform/sunburst/platform-uart.hh
nwf: That tries to use the `Bitpack(Storage)` constructor, which we've now marked `explicit`, so fails, while `{}` uses the `Bitpack()` constructor.  AFAIK there isn't a `zero_t` analog to `nullptr_t`, alas.

We could do `control = decltype(control){0};`, tho'?

sdk/include/platform/sunburst/platform-uart.hh
nwf: https://github.com/CHERIoT-Platform/cheriot-rtos/pull/558/commits/4de369aec616583706faa9b400d9075fab5156a0 adds some machinery to make this easier.  I will squash that into the intro commit before merging, but wanted it to be separate for ease of review.

sdk/include/bitpacks.hh
```diff
@@ -0,0 +1,1359 @@
...
+ *
+ * \section bitpacks_intro Introduction
+ *
+ * `Bitpack`-s are a way to handle situations where many flags and/or fields of
+ * arbitrary widths are packed into an underlying integer (such as a `uint32_t`
+ * and referred to as a `Bitpack`'s `Storage` type) at arbitrary bit alignments.
+ * This is similar to C and C++'s bitfields, but more portable, more versatile,
+ * and with a type-centric interface.  They encapsulate the bitwise AND, OR, and
+ * shift operations required to get at spans of bits within a word.  and
+ * bitwise-or They are meant to be particularly useful for handling memory
```
rmn30:
```suggestion
 * shift operations required to get at spans of bits within a word.
 * They are meant to be particularly useful for handling memory
```

sdk/include/bitpacks.hh
```diff
@@ -0,0 +1,1359 @@
...
+ * * `const` fields to represent, for example, fields mutable only by hardware.
+ *
+ * * `volatile` `Bitpacks` require overt reads and writes, making it easier to
+ *   map program source to memory or register updates.  Individual methods'
+ *   documentation will call out special handling of `volatile` forms.
+ *
+ * * a type-centric view of fields, with type-based accessors (getters, setters,
+ *   read-modify-write helpers, &c).
+ *
+ * * use in `constexpr` contexts.
```
rmn30:
```suggestion
 * * useable in `constexpr` contexts.
```

sdk/include/bitpacks.hh
```diff
@@ -0,0 +1,1359 @@
...
+ *
+ * \subsection bitpacks_intro_defining Defining Bitpacks and Fields
+ *
+ * To define a bitpack, begin by defining a `class` (or `struct`) inherits from
```
rmn30:
```suggestion
 * To define a bitpack, begin by defining a `class` (or `struct`) that inherits from
```

sdk/include/bitpacks.hh
```diff
@@ -0,0 +1,1359 @@
...
+ *     BITPACK_DEFINE_TYPE_ENUM_CLASS(Drivers, uint8_t, 12, 13)
+ *     {
+ *       None     = 0,
+ *       UpOnly   = 1,
+ *       DownOnly = 2,
+ *       Both     = 3,
+ *     };
+ *
+ * defines an `enum class Drivers` within the bitpack containing it (multiple
+ * bitpack definitions may each have their own `Driver` type within, but each
```
rmn30:
```suggestion
 * bitpack definitions may each have their own `Drivers` type within, but each
```

sdk/include/bitpacks.hh
```diff
@@ -0,0 +1,1359 @@
...
+ * `true` at the end sets the `FieldInfo` `isConst` flag and so will inhibit
+ * attempts to use the `view()` of the field to change its value.
+ *
+ * \endparblock
+ *
+ * \subsection bitpacks_intro_views Using Bitpacks
+ *
+ * Having defined a bitpack and its fields and internal machinery, one can now
+ * make use of it!  Most often, bitpacks are used directly by external code;
+ * that is. the views of its fields are public rather than being used by methods
```
rmn30: Grammar/ punctuation? I think perhaps "That is, ..."

rmn30 (approved):
A few typos in the comments. The new documentation is much easier to read. Could perhaps include some examples of accessing / writing fields?

nwf:
Grand renaming:

- `bitpacks.hh` -> `bitpack.hh`
- `Field::View` -> `Field::Proxy`
- `.view<>()` -> `.member<>()`
- `BITPACK_NAMED_VIEW` -> `BITPACK_MEMBER_ADD`
- `BITPACK_DEFINE_TYPE_` -> `BITPACK_MEMBER_ADD_...`
- `_{Q,TQ,U}TYPE` -> `_{DECLTYPE,DEPENDENT,TYPE}`
- `_{Q,TQ,U}VAL_` -> `_VALUE_{DECLTYPE,DEPENDENT,TYPE}`
- `_QUAL_` -> `_QUALIFY_`
- `BITPACK_..._OP` -> `BITPACK_OPERATE_...`
- `MASKED_REL` -> `RELATE_MASKED`

In-tree tests and users are caught up.  Out of tree users are going to experience some pain... thankfully fixes are at least somewhat compiler-directed as opposed to subtle runtime issues.

PR #561, opened 2025-07-25
Fix mistake in benchmarks

hlef:
This benchmark is supposed to measure the compartment latency but somehow we measure the function call latency instead. (We didn't have this bug when we made the paper's measurements.)

### DIFF
--- a/benchmarks/compartment-call/caller.cc
+++ b/benchmarks/compartment-call/caller.cc
@@ -72,7 +72,7 @@ int __cheri_compartment("caller") run()
 		auto [full, callPath, returnPath] =
 		  CHERI::with_interrupts_disabled([&]() {
 			  auto start  = METRIC();
-			  auto middle = local_noop_return_metric();
+			  auto middle = callee_noop_return_metric();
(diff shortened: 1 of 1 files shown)


davidchisnall:
It was measuring both so that we had a baseline. Did it somehow get copied across in a refactoring?

hlef:
> It was measuring both so that we had a baseline. Did it somehow get copied across in a refactoring?

The baseline is already measured here: https://github.com/CHERIoT-Platform/cheriot-rtos/blob/cdfb5b8977fbb6db12ddc94a70ee313a7b22787f/benchmarks/compartment-call/caller.cc#L157

But somehow the compartment latency with noop is not measured at all.

Not sure how this happened.

nwf (approved):

nwf:
Looking at commit history, this is probably my fault in https://github.com/CHERIoT-Platform/cheriot-rtos/commit/ba9a997059ee050c3812148e2dbd47d8daa9a8d4, a mis-merge of Hugo's local version used in generating the paper's numbers.  I think we're good to land it.

PR #566, opened 2025-07-26
Update a small number of places to use the most current target triple.

resistor:
These have been supported for ~6 months, so it seems safe to adopt them at this point.

### DIFF
--- a/sdk/xmake.lua
+++ b/sdk/xmake.lua
@@ -1178,6 +1178,7 @@ rule("cheriot.define-rtos-git-description")
 -- this to create threads.
 target("cheriot.loader")
 	add_rules("cheriot.component-debug")
+	add_rules("cheriot.baremetal-abi")
 	set_kind("object")
 	-- FIXME: We should be setting this based on a board config file.
 	add_files(path.join(coredir, "loader/boot.S"), path.join(coredir, "loader/boot.cc"),  {force = {cxflags = "-O1"}})


scripts/run_clang_tidy_format.sh
```diff
@@ -4,8 +4,8 @@ if [ ! -d sdk ] ; then
 	echo Please run this script from the root of the cheriot-rtos repository.
 	exit 1
 fi
-CLANG_TIDY=/cheriot-tools/bin/clang-tidy
-CLANG_FORMAT=/cheriot-tools/bin/clang-format
+CLANG_TIDY=/llvm/build-linux/bin/clang-tidy
```
davidchisnall: I think this was committed accidentally?

scripts/run_clang_tidy_format.sh
```diff
@@ -41,12 +41,6 @@ SOURCES=$(find ${DIRECTORIES} -name '*.cc' | grep -v -f scripts/run_clang_tidy_f
 echo Headers: ${HEADERS}
 echo Sources: ${SOURCES}
 
-${CLANG_FORMAT} -i ${HEADERS} ${SOURCES}
```
davidchisnall: And this.

davidchisnall (commented):
Now that we have the bare metal ABI, we should probably use that for the loader (unless @nwf did that already?).

scripts/run_clang_tidy_format.sh
resistor: Oops, fixed.

nwf:
> Now that we have the bare metal ABI, we should probably use that for the loader (unless @nwf did that already?).

I had not exactly, no, but I did have reason to add the baremetal ABI toolchain in my ~infinite stack of local patches, so I can pull that forward and then add the one more line to `xmake`.  If the two additional commits on https://github.com/CHERIoT-Platform/cheriot-rtos/compare/nwf/202507-566_loader_baremetal look right, @resistor can add them to this PR, or someone can raise a follow-up after this one lands.

nwf:
@resistor Ping re: that branch's commits.  I made the change you asked for, but there might be more that need to happen.

nwf:
Got merged as part of #571.

Issue #560, opened 2025-07-25
Side effect in switcher's compartment-call exception path


stellamplau:
@vmurali and I noticed that an exception in the [switcher's compartment call path](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/cdfb5b8977fbb6db12ddc94a70ee313a7b22787f/sdk/core/switcher/entry.S#L325) can lead to the side effect of storing registers to non-stack memory if the caller's `csp` is non-local. In this case, the switcher will first spill registers to non-local memory, then throw an exception and unwind the stack. The writes to non-local memory along this path seem like unintended side effects. 

(For context: this is motivated similarly to [#559](https://github.com/CHERIoT-Platform/cheriot-rtos/issues/559), where reasoning about the switcher would be simplified if it only ever accessed registers, the stack, and read-only memory).

A concrete proposal would be to check the integrity of csp first, before attempting to spill registers (in other words, swap the order of [.Lswitch_entry_first_spill](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/cdfb5b8977fbb6db12ddc94a70ee313a7b22787f/sdk/core/switcher/entry.S#L325C1-L325C27) with [.Lswitch_csp_check](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/cdfb5b8977fbb6db12ddc94a70ee313a7b22787f/sdk/core/switcher/entry.S#L356)).

davidchisnall:
I don’t think this is a problem. Anything that happens in this context is logically in the context of the caller.

This is a code-size optimisation to avoid every caller needing to spill these registers. The switcher can’t overwrite store-local, these spills occur exactly as if the caller had done them.

Adding additional checks here would slow down the non-malicious path. The current design is optimised for the non-malicious case and makes the malicious case. If the caller does something that makes this trap, we assume it’s malicious and force unwind out (you attack the switcher, you don’t get to retry immediately).

nwf:
> The fix is actually simple and not expensive. Just check if CSP points to a stack (CSP canStore Local i.e. has StoreLocal permission) before trying to check if stack has enough size for the callee.

Depending on what you mean... we already do that? `.Lswitch_csp_check` _is_ before `.Lswitch_stack_check_length` (which has to be after `.Lswitch_unseal_entry`, because we need to know the callee's minimum stack length, and so is actually fairly late in the cross-call path).  

[...]

PR #567, opened 2025-07-28
switcher: Check stack perms before spilling

nwf:
Full credit to @stellamplau; I just push buttons when told which buttons to push.

@stellamplau and @vmurali point out that spilling caller registers before doing permission checks means that those registers' contents could end up shared memory.  While the caller could always do this to itself, defense in depth suggests that we not let the switcher exacerbate the situation if the caller's csp has somehow been corrupted. Conveniently, in this case, it is exactly zero more instructions to perform the permission and alignment checks before the spill.  While we could spilt those, I don't think it hurts to do both earlier, since we're going to do both eventually anyway.

Tweak the test for untrusted-stack exhaustion during switcher spill so that it passes the now earlier alignment check.

Add a test explicitly for the motivating case, by copying cgp over csp before the call, which should induce a force-unwind (and does, looking at the Sail simulator's trace).

FIXES https://github.com/CHERIoT-Platform/cheriot-rtos/issues/560

### DIFF
--- a/tests/stack_integrity_thread.cc
+++ b/tests/stack_integrity_thread.cc
@@ -106,6 +106,14 @@ int exhaust_thread_stack()
  */
 int exhaust_thread_stack_spill(__cheri_callback int (*fn)())
 {
+	/*
+	 * The switcher spills five callee-save registers, but only after it checks
+	 * for 16-byte alignment of the stack pointer (and its base).  Leave
+	 * insufficient space for the whole spill while preserving alignment.
+	 */
+	constexpr size_t Stackleft = 2 * sizeof(void *);
+	static_assert((Stackleft & 0xf) == 0);
+
 	register auto      rfn asm("ct1") = fn;
 	register uintptr_t res asm("ca0") = 0;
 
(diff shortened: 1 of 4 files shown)


tests/stack_integrity_thread.cc
```diff
@@ -106,6 +106,14 @@ int exhaust_thread_stack()
  */
 int exhaust_thread_stack_spill(__cheri_callback int (*fn)())
 {
+	/*
+	 * The switcher spills five callee-save registers, but only after it checks
+	 * for 16-byte alignment of the stack pointer (and its base).  Leave
+	 * insufficient space for the whole spill while preserving alignment.
+	 */
+	constexpr size_t stackleft = 2 * sizeof(void *);
```
vmurali: /__w/cheriot-rtos/cheriot-rtos/tests/stack_integrity_thread.cc:114:19: warning: invalid case style for constant 'stackleft' [readability-identifier-naming]

Maybe do what it asks: Stackleft instead of stackleft?

vmurali:
This seems to be fine. If this ends up being in the critical path for buggy code, then maybe consider having just permission check here in the future?

PR #569, opened 2025-07-29
Tweaks and fixes to the core API benchmarks.

hlef:
Various fixes and tweaks to the core API benchmarks. Described in commit messages.

Note that none of these issues affect the paper's measurements, these happened during the merge of the patch we used for the paper's numbers (https://github.com/CHERIoT-Platform/cheriot-rtos/pull/515).


nwf (approved):
LGTM

PR #577, opened 2025-08-12
Enable sub-object bounds for the core components.

davidchisnall:
This adds a `cheriot.subobject-bounds` rule in xmake that enables sub-object bounds and opts out two data structures that do not work with sub-object bounds.

### DIFF
--- a/sdk/core/scheduler/main.cc
+++ b/sdk/core/scheduler/main.cc
@@ -541,9 +541,9 @@ __cheriot_minimum_stack(0xb0) int futex_timed_wait(Timeout        *timeout,
 	return 0;
 }
 
-__cheriot_minimum_stack(0xc0) int futex_wake(uint32_t *address, uint32_t count)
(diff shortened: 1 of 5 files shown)


sdk/core/scheduler/main.cc
```diff
@@ -541,9 +541,9 @@ __cheriot_minimum_stack(0xb0) int futex_timed_wait(Timeout        *timeout,
 	return 0;
 }
 
-__cheriot_minimum_stack(0xc0) int futex_wake(uint32_t *address, uint32_t count)
+__cheriot_minimum_stack(0xd0) int futex_wake(uint32_t *address, uint32_t count)
 {
-	STACK_CHECK(0xc0);
+	STACK_CHECK(0xd0);
```
nwf, line 546: Did you chase down why?

nwf (approved):

sdk/core/scheduler/main.cc
davidchisnall: No, but it's only one more spill.  Setting sub-object bounds on variable-sized structures can require an extra register.

PR #585, opened 2025-08-20
Fix adding the bare-metal flags in xmake.

davidchisnall:
xmake does a *monumentally* stupid and dangerous thing of assuming that it knows what flags mean and which ones are safe to deduplicate.  It sees the second `-target` and removes it, leaving the triple as an argument that clang then interprets as a file name.

See xmake discussion 4762.


nwf (approved):
I was sure I tested this; sorry.  But if this spelling works for you and for CI, great.

PR #588, opened 2025-08-22
WIP: Kudu SAFE board file & support

nwf:


scripts/msft-safe-build-firmware.sh
```diff
@@ -16,3 +16,7 @@ fi
 ${OBJCOPY} -O binary $1 - | hexdump -v -e '"%08X" "\n"' > firmware/cpu0_iram.vhx
 # Add a newline at the end of the vhx file
 echo >> firmware/cpu0_iram.vhx
+
```
davidchisnall, line 19: Possibly drop 'ibex' from this and the next file's name, since now they're generic SAFE platform things, not just for Ibex..

sdk/boards/kudu-safe-simulator.json
```diff
@@ -0,0 +1,53 @@
...
+        {
+            "name": "DummyInterrupt",
+            "number": 1,
+            "priority": 2,
+            "edge_triggered": true
+        }
+    ],
+    "defines" : [
+        "IBEX",
+        "IBEX_SAFE"
```
davidchisnall: We probably shouldn't define `IBEX`.  And maybe change `IBEX_SAFE` to `CHERIOT_SAFE` here and in the place(s) that use it?

sdk/boards/kudu-safe-simulator.json
```diff
@@ -0,0 +1,53 @@
...
+    "devices": {
+        "clint": {
+            "start": 0x14001000,
+            "length": 0x1000
+        },
+        "plic": {
+            "start": 0x10000000,
+            "end": 0x10400000
+        },
+        "revoker": {
```
davidchisnall: Does Kudu have a revoker now?

sdk/core/allocator/revoker.h
```diff
@@ -11,9 +11,10 @@
 
 #include <platform/concepts/hardware_revoker.hh>
 
-#if __has_include(<platform-hardware_revoker.hh>)
+#if defined(TEMPORAL_SAFETY) && defined(SOFTWARE_REVOKER)
+#elif __has_include(<platform-hardware_revoker.hh>)
```
davidchisnall: I find the empty `#if` followed by `#elif` confusing, it could benefit from a comment.

sdk/boards/kudu-safe-simulator.json
nwf: Whoops, no, it does not.

sdk/boards/kudu-safe-simulator.json
nwf: `CHERIOT_SAFE` seems a bit generic ("as opposed to Unsafe CHERIoT"?) so I've gone with `MICROSOFT_CHERIOT_SAFE` for the moment.

sdk/core/allocator/revoker.h
nwf: Probably clearer to restructure the conditions.  I don't recall why I wrote it like this.

PR #591, opened 2025-08-27
Reformat with clang-format-21

resistor:


nwf (approved):
Forgive me for the sin of self-approval...

PR #592, opened 2025-08-27
Apply NOLINT to an intention OOB access in a test so that clang-tidy doesn't complain about it.

resistor:


nwf:
This shouldn't be necessary after #593?

PR #595, opened 2025-09-08
Import libdivide and use it to implement fixed-cycle-count division by a constant in a sensitive code path.

resistor:


sdk/include/libdivide/libdivide.h
```diff
@@ -0,0 +1,3379 @@
...
+
+#endif  // __cplusplus
+
+#if defined(_MSC_VER) && !defined(__clang__)
+#pragma warning(pop)
+#endif
+
+#endif  // LIBDIVIDE_H
+
+// clang-format on
```
rmn30: Add trailing new line?

resistor: Done

PR #600, opened 2025-09-19
Slightly improve debug log code generation.

davidchisnall:
When there are no arguments, pass a null pointer instead of materialising a zero-length array on the stack.

This reduces the test suite size by 0.1%, but also makes evaluation of the frequency of csetbounds instructions with immediate offsets less silly.


nwf (approved):
LGTM

PR #601, opened 2025-09-22
Sealed objects: make sealed address the payload's

nwf:
This means that it is now never possible to confuse a sealed handle for a pointer to the entire allocation, as its address is always past the sealed object header, and thus interior to the allocation.

While here, shuffle some type definitions out to their own file for reuse in the loader so we don't need mysterious constants.


sdk/core/allocator/main.cc
```diff
@@ -1279,7 +1284,7 @@ __cheriot_minimum_stack(0x260) int token_obj_destroy(
   CHERI_SEALED(void *) object)
 {
 	STACK_CHECK(0x260);
-	void *unsealed;
+	SealedAllocation unsealed;
```
davidchisnall: This is now a confusing name: the type of a thing called `unsealed` is sealed thing.

sdk/core/allocator/token_types.h
```diff
@@ -0,0 +1,42 @@
+// Copyright Microsoft and CHERIoT Contributors.
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include <stdint.h>
+
+/// Opaque type.  Sealing keys don't really point to anything.
+struct SKeyStruct;
```
davidchisnall: If you're moving this around, we could give it a sensible name...

sdk/core/allocator/token_types.h
```diff
@@ -0,0 +1,42 @@
...
+ * size increases.
+ */
+struct SObjStruct
+{
+	/// The sealing type for this object.
+	uint32_t type;
+	/// Padding for alignment
+	uint32_t padding;
+	/// The real data for this.
+	char data[];
```
davidchisnall: If the pointer always points here now, we maybe don't need this element and can just have a `SealedObjectHeader` type?

sdk/core/token_library/token_unseal.S
```diff
@@ -80,18 +80,18 @@ __sealingkey_dynamic:
 
   /*
    * Load software type tag.  This will not trap, thanks to above tag check and
-   * because IRQs are deferred (see our export entry below)
+   * because IRQs are deferred (see our export entry below), unless the
```
davidchisnall, line 88: Do we actually need interrupts disabled here?

nwf: Yes, otherwise we TOCTTOU our tag check against `free()` (and revocation).  I'd prefer a story in which we didn't need IRQs deferred, but that probably means changes to the synchronous trap handler logic to clear the sensitive state from the GPRs before invoking the handler.

davidchisnall: I thought Tita proved that this was safe.  Line 75 we've lost the permit-unseal cap, so now the only sensitive thing is the unsealed capability.  If this traps, then we leak the unsealed version, but for that to happen then either:

 - We must have done a racy free, in which case we are not leaking a tagged thing, or
 - We must have created a valid sealed thing with the wrong layout.

Running with interrupts disabled protects us only against the first of these, and we leak only an unsealed thing in that case.

That said, I think it is a nicer API if it doesn't trap if it races on free.  We may want to add an explicit unwind thing to handle those cases when we start thinking about multicore.  It would be nice if that code had a special case for leaf functions so that we could avoid spilling s0 and s1 (which we know won't be clobbered), but I guess it's only two more instructions.  It's going to make formal verification much more painful unless we simply assume the setjmp bit does the right thing...

nwf:
Here's a spin, best reviewed commit by commit, that follows up the original shift of address with some tree-wide renaming and (I hope) type-level clarification within the allocator itself.

sdk/core/token_library/token_unseal.S
nwf: Oh, indeed, yes, if this load fails due to a clear tag then we have no secrets to leak.  Great.

I'd like to offer some limited unwind functionality in privileged libraries in general.  Bonus points if we can make it sufficiently expressive to implement the current "trap in cross-call is forces unwind of trusted stack" policy. :)

sdk/include/token.h
```diff
@@ -8,8 +8,8 @@
 #include <stdlib.h>
 #include <timeout.h>
 
-struct SKeyStruct;
-typedef struct SKeyStruct *SKey;
+struct TokenKeyType;
+typedef struct TokenKeyType *TokenKey;
```
davidchisnall, line 12: For backwards compatibility, can we also add a `typedef TokenKey SKey`.  Not sure if you can put `__attribute__((deprecated("SKey has been renamed TokenKey"))` on a type.  If you can, that would be nice as well.

sdk/core/allocator/main.cc
davidchisnall: I think this does make sense, but it's still somewhat confusing.  It would be good to have a comment that explains the name mismatch.

sdk/core/token_library/token_unseal.S
davidchisnall: The other alternative is to make this the internal API and add something like:

```c
__always_inline
void *token_unseal(struct TokenKey * tokenKey, struct TokenObject __sealed_capability * objectPointer)
{
    CHERIOT_DURING
        return token_unseal_libcall(tokenKey, objectPointer);
    CHERIOT_HANDLER
        return NULL;
    CHERIOT_END_HANDLER
}
```

Which would mean that Tita's proof remain simple (we're just verifying that the privileged bit doesn't leak secrets, we're not verifying that it doesn't trap, and then we have a wrapper that says 'any trap will be handled gracefully'.  We could even make this wrapper an unprivileged library.

davidchisnall: (Note: None of this is necessary for this PR)

davidchisnall:
Two small nits, then looks good.

sdk/include/token.h
rmn30: There are downstream users of these in at least the demo repo  and network stack so backwards compatibility  would be good.

sdk/core/allocator/main.cc
nwf: The following big renaming commit changes this (to `TokenHandle<true> unsealed;`) to be hopefully clearer?

nwf: I re-shuffled a bit between the two commits so there's no longer an awkward transition through "the type of unsealed is sealed".

Issue #599, opened 2025-09-17
core/allocator: heap_free_all should periodically yield the allocator lock


nwf:
At the moment, the entirety of `heap_free_all` runs under a single acquire/release of the allocator lock: https://github.com/CHERIoT-Platform/cheriot-rtos/blob/ba5cde496e359e6e7c0e65208f30f45be2732978/sdk/core/allocator/main.cc#L999-L1038
While it's true that our heaps will never have gazillions of objects in it, this nevertheless is a source of long-tail latency, which we make some efforts to avoid elsewhere (for example, our O(1) de-quarantining per heap operation).  I _believe_ `heap_free_all` can safely drop and reacquire the heap lock if it first takes an _ephemeral_ claim on the `chunk` it is currently processing.  (The thread currently running `heap_free_all` by definition will have no ephemeral claims that this risks clobbering: `heap_free_all` is a cross-compartment call, and so any ephemeral claims have been shed.)  Because the allocator lock is held, the hazard epoch is necessarily both stable and even, and so there's no need for the full `heap_claim_ephemeral` machinery: just store the pointer to the chunk _body_ into one of the slots returned by `switcher_thread_hazard_slots`.

Whether or not we should wake sleeping threads as part of this lock yield is a good question, but can be a follow-up issue.  (Maybe we should be recording something about how much progress needs to be made before we wake the thundering herd?)

Issue #140, opened 2023-11-29
Create sub-quotas for allocation capabilities


davidchisnall:
The allocator currently deals with static heap quotas.  Sometimes it's useful to allow another compartment to allocate memory on your behalf, but not to consume your entire quota.  We should allow a mechanism for creating a new quota object that contains part of an existing quota.  This should have the same ID as the original, so that objects can be freed in both.

The main complication in this is that, currently, quotas are checked without the allocator lock being held (I think).  We need to make sure that the mechanism is robust in the presence of a quota being deallocated during an allocation or deallocation operation.

hlef:
We are encountering an issue in the network stack which would likely benefit from such a feature.

- Sockets are allocated with the quota of the TCP/IP compartment. See [here](). This would be hard to change since this allocation and the corresponding free are done deep in the internals of FreeRTOS+TCP.
- To prevent the caller from DoS-ing the TCP/IP compartment by exhausting its quota through creating many sockets without freeing them, we claimed the allocation with the quota of the caller. See [here](). Thus the caller is accountable for the socket allocation, and would exhaust its own quota in the process.
- Unfortunately this approach does not work, since the caller can always call `heap_free_all` to remove this claim. The problem would also be there if the caller simply had a much larger quota than the TCP/IP stack.

To address this issue, it would be nice if we could have a mechanism to loan quota: the caller would give enough quota to allocate the socket to the TCP/IP stack, and the loan would be returned when the socket is freed.

This does raise questions on how to refund the quota though:

- The free of the socket happens asynchronously, deep in FreeRTOS+TCP. There is no callback to our code to say that the socket was freed.
- We could refund the quota when closing the socket, and claim it with the TCP/IP stack compartment's quota. If the TCP/IP stack's quota is insufficient we would simply not close the socket. We would thus allow callers to consume some of the stack's quota, but only for a short and rather well-defined amount of time.

tolbooth:
So, I've taken a look at this issue, and unfortunately it doesn't seem like there's a simple solution.

First off, it's not entirely clear that these quotas *should* share an ID. The issue with this model is that `heap_free_all` depends directly on said ID, meaning:

1. We cannot support untrusted delegatees. Compartment A may want to hand out quota to compartment B to make allocations, but not allow B to arbitrarily free A's own allocations. If we assume a fully trusted recipient, we should have no trouble handing them the original quota anyway.
[...]

PR #604, opened 2025-10-08
Add heap_free_all to the allocator fuzz tests.

tolbooth:
This PR comes as a result of some discussion with regards to #599. 

It was mentioned that fuzzing for `heap_free_all` would be a helpful addition, and so this PR addresses that. Since we can't invoke `heap_free_all` on the default allocator capability (without messing up allocations we're using to test), I took this as an opportunity to add a layer of abstraction to allow for fuzzing multiple heap quotas. This also serves as preparation for testing sub-quotas (#140) when we want to pursue that.

### DIFF
--- a/tests/allocator-test.cc
+++ b/tests/allocator-test.cc
@@ -441,7 +441,7 @@ namespace
 			{
 				for (auto eachAllocation : allocations)
 				{
-					heap_free(quota, eachAllocation);
+					TEST_SUCCESS(heap_free(quota, eachAllocation));
 				}
 				allocations.clear();
 			}
@@ -485,7 +485,7 @@ namespace
 
 			cachedFrees[rand.next() % NCachedFrees] = p;
 
-			heap_free(heap.quota, p);
+			TEST_SUCCESS(heap_free(heap.quota, p));
 		};
 
 		auto doFreeAll = [&](HeapTestState &heap) {


tests/allocator-test.cc
```diff
@@ -410,38 +410,100 @@ namespace
 		ds::xoroshiro::P32R16 rand = {};
 		auto                  t    = Timeout(0); /* don't sleep */
 
-		std::vector<void *> cachedFrees;
-		cachedFrees.resize(NCachedFrees);
+		/*
+		 * A struct that holds per-heap state for fuzzing. Will come in handy
+		 * when sub-quotas are added.
+		 */
```
davidchisnall: Please use doc comments (`/**`) and complete sentences.  The fields should also have a doc comment explaining what they're for.

tests/allocator-test.cc
```diff
@@ -410,38 +410,100 @@ namespace
...
+		struct HeapTestState
+		{
+			AllocatorCapability quota;
+			std::vector<void *> allocations;
+			std::vector<void *> cachedFrees;
+
+			HeapTestState(AllocatorCapability quota) : quota(quota)
+			{
+				allocations.reserve(MaxAllocs);
+				cachedFrees.resize(NCachedFrees);
```
davidchisnall: Note that these will use the default malloc capability to allocate the required space and will crash if we're out of memory.  That should probably be documented.

tests/allocator-test.cc
```diff
@@ -410,38 +410,100 @@ namespace
...
+			}
+
+			void clear()
```
davidchisnall: This leaves the object in an undefined state, so feels more like a destructor than a method.

tests/allocator-test.cc
```diff
@@ -410,38 +410,100 @@ namespace
...
+			}
+
+			ssize_t freed = heap_free_all(heap.quota);
 
+			TEST(freed > 0,
+			     "heap_free_all on {} returned {} but had {} allocations",
+			     &heap,
+			     freed,
+			     allocCount);
+
```
nwf, line 511: Might be good to check that all pointers in `heap.allocations` are now freed?

tests/allocator-test.cc
tolbooth: Along those lines, I'm wondering if it makes sense to also move the `doAlloc`, `doFree`, `doFreeAll` lambdas from the test function scope into the struct (in this case, maybe make it a class)?

<del>I could also add the</del>
```c
for (void* p : heap.cachedFrees)
{
    TEST(...)
}
```
<del>loop to this class as a `checkForNecromancy` method.</del>

tests/allocator-test.cc
```diff
@@ -410,38 +410,100 @@ namespace
...
-		cachedFrees.resize(NCachedFrees);
+		/*
+		 * A struct that holds per-heap state for fuzzing. Will come in handy
+		 * when sub-quotas are added.
+		 */
+		struct HeapTestState
+		{
+			AllocatorCapability quota;
+			std::vector<void *> allocations;
+			std::vector<void *> cachedFrees;
```
nwf: I think this doesn't need to be per-quota (in that I think there's no particular reason to believe that any kind of necromancy bug would correlate with quota objects' handling).

tests/allocator-test.cc
nwf: `.clear()` should leave the `std::vector`-s in defined state?

tests/allocator-test.cc
```diff
@@ -410,38 +410,100 @@ namespace
...
+			     "heap_free_all on {} returned {} but had {} allocations",
+			     &heap,
+			     freed,
+			     allocCount);
+
+			heap.allocations.clear();
+		};
+
+		std::vector<HeapTestState> heapList;
+		heapList.emplace_back(SECOND_HEAP);
```
nwf, line 528: Also add `MALLOC_CAPABILITY`?  (If `SECOND_HEAP` is the only quota used, then it's going to pretty strongly limit the heap shapes we build, since its quota is 1024 bytes and that limits us to 64 objects at most, a far cry from `MaxAllocs`'s current value of 256.)

tests/allocator-test.cc
```diff
@@ -453,37 +528,54 @@ namespace
...
 					size_t szix = rand.next() % NAllocSizes;
 					size_t sz   = AllocSizes[szix];
-					doAlloc(sz);
+					doAlloc(heap, sz);
 				}
 			}
 
-			for (size_t j = rand.next() & 0xF; j > 0; j--)
+			/**
+			 * We want to free all 12.5% of the time. This value is chosen to be
```
nwf: That seems high.  We perform at most 15 allocations every round, so we should expect something like 7 each time, and if we're dumping everything in expectation in every 8 rounds, I think that means we expect to get something like 50 objects allocated before we free everything, and that doesn't give a lot of opportunity to churn heap shapes.

tests/allocator-test.cc
tolbooth: My one thought with this is that if we want to add sealed allocations to the fuzz test, this wouldn't necessarily be true? I can add that test and leave a comment there along those lines?

tests/allocator-test.cc
tolbooth: Regarding this, would you prefer that I 
1. Just reduce this percentage, and make no other changes?
2. Reduce the percentage, and add more iterations to the fuzz test?
3. Have a separate loop for testing `heap_free_all`?

tests/allocator-test.cc
nwf: That's true.  (And similarly, once we start having claims, we shouldn't necessarily consider the pointers passed to `free()` now free, because they might be claimed.)

tests/allocator-test.cc
nwf: I think just reducing the percentage is fine.

tolbooth: Tested with 1/32 and 1/64. 1/64 meant in practice that we only invoke `heap_free_all` once per run of the test suite (when `NDEBUG` is not set), which seems low. I've added justification to the comment.

tests/allocator-test.cc
```diff
@@ -410,38 +413,125 @@ namespace
...
+				 * capability, and will crash if we are out of memory.
+				 */
+				allocations.reserve(MaxAllocs);
+			}
+
+			~HeapTestState()
+			{
+				for (auto eachAllocation : allocations)
+				{
+					heap_free(quota, eachAllocation);
```
rmn30:
```suggestion
					TEST_SUCCESS(heap_free(quota, eachAllocation));
```

rmn30: Might need tweaking if `eachAllocation` may have been freed already?

tests/allocator-test.cc
```diff
@@ -410,38 +413,125 @@ namespace
...
+			void  *p             = heap.allocations[ix];
+			heap.allocations[ix] = heap.allocations.back();
+			heap.allocations.pop_back();
 
 			TEST(CHERI::Capability{p}.is_valid(), "Double free {}", p);
 
 			cachedFrees[rand.next() % NCachedFrees] = p;
 
-			free(p);
+			heap_free(heap.quota, p);
```
rmn30: Again might want to check for success here?

tests/allocator-test.cc
tolbooth: Will add the check, as it should be fine--anytime we free in this test we remove from allocations (to avoid accidental double freeing). This would fail only if something else has gone wrong. Maybe be more correct, though, to check the tag for all allocations to ensure they're all invalid? Perhaps much of a muchness. Will implement with TEST_SUCCESS.

rmn30 (approved):

Issue #452, opened 2025-02-14
heap_claim should return negative on failure


rmn30:
Currently `heap_claim` returns zero on failure which is inconsistent with other compartment calls.

hlef:
This was addressed by #605. @davidchisnall can you close this?

PR #605, opened 2025-10-08
Allocator: Standardize heap_* function return values in main.cc

tolbooth:
This addresses #452. Return values added are modeled on those given by `heap_free_internal`.

Also changed is the return value for `heap_quota_remaining`, similarly to bring it in line with what other functions are returning. Along with updates to the documentation to these two, I've expanded the documentation for `heap_can_free`.

I'm wondering if -EPERM is the right value to return if the claim fails due to the quota being too small? It seems like there are two ways to view a too-small quota: either as "not enough memory to perform this operation" or as "you do not have permission to allocate this much".

### DIFF
--- a/tests/allocator-test.cc
+++ b/tests/allocator-test.cc
@@ -1115,7 +1115,10 @@ namespace
 		state = 2;
 		while (state.load() == 2)
 		{
-			sleep(1);
+			/*
+			 * Temporary fix, See Issue #509.
+			 */
+			sleep(2);
 		}
 
 		debug_log("End of token hazard test");


davidchisnall:
`-ENOMEM` seems perhaps a better choice, but it might be worth adding an error code for quota exhausted.  We should also probably make `heap_allocate` return negative error codes on failure, since the `malloc` wrapper turns all untagged values into null.

sdk/include/stdlib.h
```diff
@@ -276,16 +281,18 @@ ssize_t __cheri_compartment("allocator")
   heap_free_all(AllocatorCapability heapCapability);
 
 /**
- * Returns 0 if the allocation can be freed with the given capability, a
- * negated errno value otherwise.
+ * Returns 0 if the allocation can be freed with the given capability,
+ * `-EPERM` if `heapCapability` is not valid or it cannot free the provided
+ * allocation (such as when there is an active claim), and `-EINVAL` if
```
rmn30: Does this really return negative if there is an active claim? That doesn't sound right as `heap_free` would succeed in that case and defer the deallocation until the claim is released.

tolbooth: Thanks for the catch--I'll make the example here something along the lines of  "when you are not the owner, and you have no claim", as I think that's not revealing too many internal details, but gives a little more description than just "can't free".

tests/allocator-test.cc
```diff
@@ -953,6 +1001,14 @@ int test_allocator()
 
 	test_preflight();
 
+	// Test `heap_quota_remaining` failure value. This is used throughout the
+	// allocator tests, so ensure that we fail correctly.
+	int failedQuotaRemaining = heap_quota_remaining(nullptr);
+	TEST(failedQuotaRemaining == -EPERM,
```
rmn30, line 1147: We have a macro `TEST_EQUAL` that I think could be used here. It has the advantage that it will print both expected and actual value on failure.

sdk/core/allocator/main.cc
```diff
@@ -288,7 +288,8 @@ namespace
 			// `AllocateWaitNone`, or `timeout` is 0), fail now.
 			if (flags == AllocateWaitNone || !may_block(timeout))
 			{
-				return nullptr;
+				// Transient failure, may succeed if retried
+				return reinterpret_cast<void *>(-EAGAIN);
```
rmn30: I don't think `EAGAIN` is always appropriate here. For example `mspace_dispatch` can return `AllocationFailurePermanent`, `AllocationFailureQuotaExceeded` or `AllocationFailureHeapFull` in some cases. These should probably be mapped to `-EINVAL` (for the first) or `-ENOMEM` (for the latter two). Additionally this may be a bit pedantic but I'm wondering if `-ETIMEDOUT` might be more appropriate for the case `ret == AllocationRevocationNeeded && flags != AllocateWaitNone && !may_block(timeout)`? 
 In fact I feel like this is really just an optimisation for the (hopefully) infrequent error cases so only AllocationFailurePermant should be handled here we should just fall through to below in other cases? Tagging @nwf and @davidchisnall for an opinion.

sdk/core/allocator/main.cc
```diff
@@ -370,16 +373,18 @@ namespace
...
-					return nullptr;
+					// Failed to acquire lock within allowed timeout.
+					return reinterpret_cast<void *>(-ETIMEDOUT);
 				}
 				continue;
 			}
 			if (std::holds_alternative<MState::AllocationFailurePermanent>(ret))
 			{
-				return nullptr;
+				return reinterpret_cast<void *>(-ENOMEM);
```
rmn30: Maybe this should be moved further up as per comment above? Also perhaps `-EINVAL` instead?

nwf: `AllocationFailurePermanent` is returned when...
- the requested size is 0
- CRRL of the requested size has overflowed
- the requested size is larger than the heap

I think `-EINVAL` is a fine return for these cases, yes.

sdk/core/allocator/main.cc
nwf: Agreed.  That is, I am convinced that we should just test for `Permanent` failures here at the top of the loop (and return `-EINVAL`, as you say, below) and let the `-ETIMEDOUT`/`-EAGAIN` cases sort themselves out below.

tolbooth: I've updated the PR accordingly. This is documented in stdlib and in the commit message, but now:
- `-EAGAIN` is returned to signal a revocation failure only if `AllocateWaitRevocationNeeded` is false
- `-ENOMEM` is returned to signal OOM only if the respective wait flags are set to false.
- `-ETIMEDOUT` is the default return value for any non-permanent allocation failure for a blocking call.
- `-EINVAL` is returned for permanent failures in blocking and non-blocking modes.

sdk/include/stdlib.h
davidchisnall: 'You are not the owner and do not have a claim' is the API contract, not an implementation detail.  More generally: if this object is not counted against your quota, then it cannot be refunded to your quota.

rmn30:
Sorry but there is another problem with this PR. In `allocator/main.cc:1159` we have:

```C++
SealedAllocation obj{static_cast<SObjStruct *>(malloc_internal(
		  sealedSize, std::move(g), capability, timeout, true))};
		if (obj == nullptr)
		{
			Debug::log<DebugLevel::Warning>(
			  "Underlying allocation failed for sealed object");
			return {nullptr, nullptr};
		}
```

This is now broken as the null check only checks whether the address of the capability is zero, which it no longer is in the error cases.

The best solution to this is to change the return type of `malloc_internal` form `void*` to a union type of pointer and integer (error). @davidchisnall has a plan to write such a type but for now we could make it return `CHERI::Capability<void>` and test for `is_valid` manually.

rmn30:
> Thanks for another catch! Even before these changes, is there a reason we would'nt've seen `-ENOTENOUGHSTACK` for this call? I'll also update the documentation in allocator.md

That's true and one reason that we ought to be more strict about the return type of compartment calls. We have a [compiler  warning](https://github.com/CHERIoT-Platform/llvm-project/issues/34) about voids, but it is not turned on and probably not strict enough. I thin the only type safe return values are `int` and `ErrorOr` (once merged), but most of our public APIs do not conform to this at the moment. Something for us to think about...
 
> Absolutely a good idea to have real enforcement for handling the errors!
> 
> I pushed up the change to this check after testing locally with both debug and release build test runs passing, but it seems like there's failing tests in the `CHERIoT RTOS Tests / run-tests (debug, ibex-safe-simulator)` test. Will investigate in the morning.

I'll have a look.

tolbooth:
> > Thanks for another catch! Even before these changes, is there a reason we would'nt've seen `-ENOTENOUGHSTACK` for this call? I'll also update the documentation in allocator.md
> 
> That's true and one reason that we ought to be more strict about the return type of compartment calls. We have a [compiler warning](https://github.com/CHERIoT-Platform/llvm-project/issues/34) about voids, but it is not turned on and probably not strict enough. I think the only type safe return values are `int` and `ErrorOr` (once merged), but most of our public APIs do not conform to this at the moment. Something for us to think about...
> 
> > Absolutely a good idea to have real enforcement for handling the errors!
> > I pushed up the change to this check after testing locally with both debug and release build test runs passing, but it seems like there's failing tests in the `CHERIoT RTOS Tests / run-tests (debug, ibex-safe-simulator)` test. Will investigate in the morning.
> 
> I'll have a look.

I've taken a look, and have narrowed things down somewhat. The failure is linked to my `fuzz_free_all` PR (which makes sense, as this PR did not fail until that was merged into main).

The exact failure is our `MALLOC_QUOTA` is 40 bytes short right when `test_claims()` starts execution, i.e. it fails the first test in that function. This occurs just after the hazard token tests.

Runs built with `xmake f --board=ibex-safe-simulator --sdk=/cheriot-tools/ --debug-loader=y --debug-scheduler=y --debug-allocator=information --allocator-rendering=y -m debug  --print-doubles=y --print-floats=n` will fail on / after `f1e0d4428bcc27a5bbba2e690e392777b87e7162`.

This is a seemingly innocuous commit, simply appending some additional tests to the end of `test_claims`, and is only the second commit in this PR.

`main` still passes the test run, as does `7d958e13a3cc80e1c26a27e6b64ae1479edd14a2` on top of main, the first commit of this PR. It's only after we increase the length of the `test_claims` function that we encounter this error. Deleting the lines added by this commit causes the tests to pass on everything *except* the latest commit in this PR.

Notably, by commenting out the line in which we call `heapList.emplace_back(FUZZ_HEAP);`, the tests pass on ALL commits, including up to most recent. However, `test_claims` runs *well before* `test_fuzz`, so we never actually run the fuzz tests (i.e. should never touch that line) in our failing cases.

Removing the declaration of `FUZZ_HEAP` and replacing its use in the heapList with an existing quota like `EMPTY_HEAP` has no effect, and the tests still fail.

rmn30:
Sounds like we might be leaking some memory? Worth printing the result of `heap_quota_remaining` in a few places to see if it is unexpectedly consumed? The fact it is only failing on ibex-safe could be because the heap is too small as it has less memory than other platforms.

rmn30:
Answering my own question: the existing `sleep(1)`s in the main thread are non-deterministic and could end up pre-empting the background thread at an unlucky point, which will depend on timing, especially when `--debug-scheduler=y` as this reduces the quantum for which the background thread actually runs to quite small due to the time spent debugging printing.

tolbooth:
> Answering my own question: the existing `sleep(1)`s in the main thread are non-deterministic and could end up pre-empting the background thread at an unlucky point, which will depend on timing, especially when `--debug-scheduler=y` as this reduces the quantum for which the background thread actually runs to quite small due to the time spent debugging printing.

This makes sense--so adding another `sleep(1)` call or a `TEST` statement anywhere in the file could shift around the generated asm, which changes how many cycles it takes for the main thread to reach its sleep(1) calls? This would shift the timing of scheduler invocations / interrupts, which determines whether those interrupts happen to preempt the background thread before or after it completes token_obj_destroy()? And that's why making those changes effects the pass/fail behavior. Am I understanding this correctly? 

Then the ibex debug build is just particularly susceptible to running into this issue due to logging eating up so many cycles, meaning the background thread has fewer cycles to complete cleanup before the next preemption?

And just to clarify: these calls are "non-deterministic" in that the preemption behavior is highly dependent on timing, but "deterministic" in that for a given build, they will always result in the same sequence? I.e., a failing build will consistently fail every time, and vice-versa for passing, but changing the code (even "unrelated" parts) could flip which case we're in?

sdk/core/allocator/main.cc
```diff
@@ -1048,26 +1052,23 @@ __cheriot_minimum_stack(0x230) void *heap_allocate_array(
   uint32_t            flags)
 {
 	STACK_CHECK(0x230);
-	if (!check_timeout_pointer(timeout))
+	if (!check_timeout_pointer(timeout) ||
+	    !check_pointer<PermissionSet{Permission::Load, Permission::Store}>(
```
rmn30: `check_timeout_pointer` already checks permissions so the second half of this condition is redundant. ~~Looks like this came from `futex_timed_wait` (perhaps copied by AI?). We should fix that in another PR~~ (`futex_timed_wait` is actually checking two different pointers).

tolbooth: good spot, I should've double checked that. looking at the blame it seems like this check just wasn't removed when  the timeout check helper was added. addressed.

PR #607, opened 2025-10-10
Get rid of out-of-line CTZ/CLZ expansions when the RV Zbb extension is present.

resistor:


sdk/lib/crt/cz.c
```diff
@@ -1,17 +0,0 @@
-// Copyright Microsoft and CHERIoT Contributors.
```
davidchisnall, line 1: These APIs are public, we probably need to at least replace them with a static inline wrapper around the builtin.

davidchisnall: Delete button isn’t working. Rereading more carefully, I see this is what you did.

sdk/lib/crt/arith64.c
```diff
@@ -208,8 +216,9 @@ typedef union
     }
 
     // let's do long division
-    char bits = __clzdi2(b) - __clzdi2(a) + 1;  // number of bits to iterate (a and b are non-zero)
-    arith64_u64 rem = a >> bits;                   // init remainder
+	char bits = __builtin_clzg(b) - __builtin_clzg(a) +
```
resistor, line 213: @rmn30 Should we #ifdef this part? I think it's still better for code size to call __clzdi2 when the B extension is not present.

rmn30: Yes, maybe. I guess we could measure? I also want to compare existing implementations with the equivalent builtins (without B). Are you sure that clang will never emit a call to `__clzdi2` etc. when B is enabled?

resistor: I do not believe it will, but I don't have a way to positively prove that it won't.

sdk/lib/crt/arith64.c
```diff
@@ -42,10 +42,14 @@ arith64_s64 __cheri_libcall __ashldi3(arith64_s64 a, int b) __asm__("__ashldi3")
 arith64_s64 __cheri_libcall __absvdi2(arith64_s64 a) __asm__("__absvdi2");
 arith64_s64 __cheri_libcall __ashldi3(arith64_s64 a, int b) __asm__("__ashldi3");
 arith64_s64 __cheri_libcall __ashrdi3(arith64_s64 a, int b) __asm__("__ashrdi3");
+
+#if !defined(__riscv_zbb)
```
davidchisnall, line 49: This file has a list of changes from the upstream version at the top.  Please can you add this to the list?

sdk/lib/crt/arith64.c
davidchisnall: I think we should get the libcall version from the builtin unless the expansion is small?

rmn30: I tried calling `__clzdi2` and clang inlined it so it didn't help code size! Adding `no inline` helped but actually I'm not sure we care so much about code size if we don't have B: we should make sure all real CPUs have it as it's not that hard and can help a lot.

sdk/lib/crt/arith64.c
resistor: Done

sdk/lib/crt/arith64.c
```diff
@@ -132,52 +140,46 @@ typedef union
...
+// Return the number of bits set in a.
+[[clang::no_builtin]] int __popcountsi2(arith64_u32 a)
+{
+    return __builtin_popcount(a);
+}
+
+// Return the number of bits set in a.
+[[clang::no_builtin]] int __popcountdi2(arith64_u64 a)
+{
+    return __builtin_popcountll(a);
```
resistor: Maybe use __builtin_popcountg() so it's generic on the argument's type? Same for __builtin_ctz / __builtin_clz above?

PR #613, opened 2025-10-23
Add an ErrorOr<T> class.

davidchisnall:
This provides some convenience wrappers around tagged unions of pointers and error codes.

### DIFF
--- a/sdk/include/cheri.hh
+++ b/sdk/include/cheri.hh
@@ -1582,4 +1582,120 @@ namespace CHERI
 		    .value_or(RegisterNumber::Invalid);
 		return {causeCode, registerNumber};
 	}
+
+	/**
+	 * Helper for a value that encodes either a pointer or an error value in a
+	 * pointer-sized value, discriminated by the tag bit.
+	 *
+	 * This is a *non-owning* pointer type.
+	 */
+	template<typename T>
+	class ErrorOr
+	{
+		/// The underlying value
+		Capability<T> pointer;
+		public:
+		/**
+		 * Construct from a pointer.  This may be a C API that uses the
+		 * convention of holding a pointer or an error value.  You may call
+		 * this constructor with an invalid pointer.
+		 */
+		ErrorOr(T* pointer) : pointer(pointer) {}
+
+		/**
+		 * Initialise the value with an error value.  By convention, error
+		 * numbers are negated in CHERIoT RTOS.
+		 */
+		ErrorOr(int error) : pointer(reinterpret_cast<T*>(static_cast<intptr_t>(error))) {}
+
+		/**
+		 * Returns `true` if this holds an error value, false otherwise.
+		 */
+		bool is_error()
+		{
+			return !pointer.is_valid();
+		}
+
+		/**
+		 * If this holds an error value, return it, otherwise return 0.
+		 *
+		 * CHERIoT RTOS follows the C convention of using 0 to indicate
+		 * success, but this may result in ambiguity.  Use `is_error` if this
+		 * ambiguity matters.
+		 */
+		int as_error()
+		{
+			if (!is_error())
+			{
+				return 0;
+			}
+			return reinterpret_cast<intptr_t>(pointer.get());
+		}
+
+		/**
+		 * If this holds a pointer, return it, otherwise return `nullptr`.
+		 * This ensures that all values that are *not* valid pointers are
+		 * mapped to null.
+		 */
+		T* as_pointer()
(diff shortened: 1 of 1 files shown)


rmn30:
I tried using this [here](https://github.com/CHERIoT-Platform/cheriot-rtos/commit/810e588250177f8a36b61a8a178fcff0bf227336) motivated by https://github.com/CHERIoT-Platform/cheriot-rtos/pull/605#issuecomment-3436313532 . It worked fine but I had to add `as_raw` to convert to the existing raw pointer or integer API.

sdk/include/cheri.hh
```diff
@@ -1582,4 +1582,120 @@ namespace CHERI
...
+	 * Helper for a value that encodes either a pointer or an error value in a
+	 * pointer-sized value, discriminated by the tag bit.
+	 *
+	 * This is a *non-owning* pointer type.
+	 */
+	template<typename T>
+	class ErrorOr
+	{
+		/// The underlying value
+		Capability<T> pointer;
```
nwf, line 1596: Perhaps

```suggestion
	template<typename T, bool IsSealed = false>
	class ErrorOr
	{
		/// The underlying value
		Capability<T, IsSealed> pointer;
```

sdk/include/cheri.hh
```diff
@@ -1582,4 +1582,120 @@ namespace CHERI
...
+		 */
+		ErrorOr(int error) : pointer(reinterpret_cast<T*>(static_cast<intptr_t>(error))) {}
+
+		/**
+		 * Returns `true` if this holds an error value, false otherwise.
+		 */
+		bool is_error()
+		{
+			return !pointer.is_valid();
+		}
```
nwf, line 1617: FWIW, `std::optional` also exposes an implicit conversion to `bool` to signal non-empty.  Not sure if that'd be useful here, but maybe?

sdk/include/cheri.hh
```diff
@@ -1582,4 +1582,120 @@ namespace CHERI
...
+		/**
+		 * Monadic helper modelled on `std::optional`.  Takes a reference to a
+		 * callable object that accepts a `T*` or something that can be
+		 * implicitly cast to a `T*`.  The return value for the argument should
+		 * be either `void` or something that can implicitly cast to an
+		 * `ErrorOr<T>`.
+		 *
+		 * The callback is invoked if and only if this holds a valid pointer.
+		 */
+		ErrorOr and_then(auto &&function)
```
nwf, line 1658: This doesn't allow changing the type in the returned `ErrorOr`.

sdk/include/cheri.hh
```diff
@@ -1582,4 +1582,120 @@ namespace CHERI
...
+				return nullptr;
+			}
+			return pointer;
+		}
+
+		/**
+		 * Monadic helper modelled on `std::optional`.  Takes a reference to a
+		 * callable object that accepts a `T*` or something that can be
+		 * implicitly cast to a `T*`.  The return value for the argument should
+		 * be either `void` or something that can implicitly cast to an
```
nwf, line 1653: The "either `void` or" bit is new relative to `std::optional`.  Why?

sdk/include/cheri.hh
```diff
@@ -1582,4 +1582,120 @@ namespace CHERI
...
+		 * `ErrorOr<T>`.
+		 *
+		 * The callback is invoked if and only if this holds a valid pointer.
+		 */
+		ErrorOr and_then(auto &&function)
+		{
+			if (is_error())
+			{
+				return this;
```
nwf, line 1662: Prrrrobably that should be... ?
```suggestion
				return *this;
```

sdk/include/cheri.hh
```diff
@@ -1582,4 +1582,120 @@ namespace CHERI
...
+			}
+			auto result = function(pointer);
+			if constexpr (std::is_same_v<decltype(result), void>)
+			{
+				return this;
+			}
+			else
+			{
+				return result;
+			}
```
nwf, line 1672: This seems to not work as intended with `void`-returning `function`-s.
```
/cheri/source/cheriot/cheriot-rtos/sdk/include/cheri.hh:1703:9: error: variable has incomplete type 'void'
 1703 |                         auto result = function(pointer);
```
Perhaps...
```suggestion
			if constexpr (std::is_same_v<std::invoke_result_t<decltype(result), T *>, void>)
			{
				function(pointer);
				return *this;
			}
			else
			{
				return function(pointer);
			}
```

PR #614, opened 2025-10-24
Add initial support for user permissions on sealed objects.

davidchisnall:
All sealed object handles have all three user permissions set when they are created.  They currently can have all three bits set.

### DIFF
--- a/sdk/core/allocator/main.cc
+++ b/sdk/core/allocator/main.cc
@@ -1179,8 +1179,13 @@ namespace
 		  reinterpret_cast<TokenObjectType *>(headerThenPayload.get())};
 		payloadAfterHeader.address() += sizeof(TokenObjectHeader);
 
-		// Sealed points at payload but can access header, at negative offset
-		auto sealed = payloadAfterHeader.seal(allocatorSealingKey);
+		// Sealed points at payload but can access header, at negative offset.
+		auto sealed = [=]() {
+			auto res{payloadAfterHeader};
+			// Set all of the user permissions
+			res.address() += 7;
+			return res.seal(allocatorSealingKey);
+		}();
 
 		// Unsealed points at payload and can access only payload
 		TokenHandle<false> payload{payloadAfterHeader.get()};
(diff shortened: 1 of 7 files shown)


sdk/core/allocator/main.cc
```diff
@@ -1179,10 +1179,15 @@ namespace
...
 		payloadAfterHeader.address() += sizeof(TokenObjectHeader);
+		// Set all of the user permissions
+		payloadAfterHeader.address() += 7;
 
 		// Sealed points at payload but can access header, at negative offset
 		auto sealed = payloadAfterHeader.seal(allocatorSealingKey);
 
+		// Undo the displacement for the user permissions
+		payloadAfterHeader.address() -= 7;
+
```
nwf, line 1189: I know it's a completely trivial thing to object to add-copy-and-subtract, but maybe instead...
```
		// Sealed points at payload but can access header, at negative offset.
		auto sealed = [=](){
			auto res {payloadAfterHeader};
			// Set all of the user permissions
			res.address() += 7;
			return res.seal(allocatorSealingKey);
		}();
```

tests/allocator-test.cc
```diff
@@ -1046,7 +1048,7 @@ namespace
 		state = 2;
 		while (state.load() == 2)
 		{
-			sleep(1);
+			sleep(2);
```
nwf, line 1123: This really bothers me and I'd like to spend some time trying to understand why it's needed.

davidchisnall: It's needed only in debug builds, so I *believe* it's that the claims are not dropped quite fast enough.  I'd expect them to be dropped on the call to the allocator though.

davidchisnall: The failing case has allocator and *scheduler* debug enabled, so I wonder if a one-tick sleep isn't actually enough to run the other thread at all, after the debug log messages from the scheduler.

tolbooth: Is this in response to the same failure we're seeing in #605?
```
allocator-test.cc:609 Invariant failure in test_claims
After claim and free from 1048576-byte quota, 1048536 bytes left before running claims tests
```

PR #615, opened 2025-10-28
cheri.hh: remove Capability move operators

nwf:
This should make CHERI::Capability<>s eligible to be returned in registers.  Attempt a static_assert that will at least catch some future changes that would break this.

This makes the test suite slower, but almost exclusively because it makes the code smaller and the heap larger, so we do more work in the allocator test.  The other tests look to wibble in cycle counts a bit.

### DIFF
--- a/sdk/include/cheri.hh
+++ b/sdk/include/cheri.hh
@@ -830,27 +830,12 @@ namespace CHERI
 		/// Constructor, takes an existing pointer to wrap
 		constexpr Capability(decltype(ptr) p) : ptr(p) {}
 		/// Copy constructor, aliases the object that is pointed to by `ptr`.
-		constexpr Capability(const Capability &other) : ptr(other.ptr) {}
-		/// Move constructor.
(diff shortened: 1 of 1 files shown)


tests/misc-test.cc
```diff
@@ -11,6 +11,15 @@
...
 
+/*
+ * Partially ensure that CHERI::Capability<>s can be housed in registers rather
+ * than needing to go via memory.  This is an attempt to capture some of C++'s
+ * [class.temporary]p3 requirements, but I don't know how to capture all of
+ * them in the existing metaprogramming library.
+ */
+static_assert(std::is_trivially_copy_constructible_v<Capability<void>> &&
+              std::is_trivially_destructible_v<Capability<void>>);
+
```
davidchisnall: Is there a reason to not put this in cheri.hh, where it would fire as soon as you (okay, I) break it?

nwf: I don't have strong opinions between the two.  I happened to put it over here so it wasn't computation every compilation unit including cheri.hh would have to do.

Issue #141, opened 2023-11-29
Separate permissions in quotas


davidchisnall:
It would be nice to have allocation capabilities that allow claiming but not allocation.  This provides some defence in depth against compromised compartments.  Once #140 is created, we should also separate allocation and deallocation rights so that you can hand out a sub-quota that allows allocation but not deallocation, or one that allows cleanup and returning memory to the original quota, but not allocation.

PR #617, opened 2025-10-30
Allocator: Fine-grained allocator capability permissions

tolbooth:
We add permissions to allocator capabilities, building on top of @davidchisnall's work adding user permissions to sealed objects.

An allocator capability can now have its permission to be used in allocation/claiming and freeing/releasing claims restricted independently, addressing #141.

Since the user permissions work gives us 3 bits to work with, I had floated the idea of using our third bit to describe the ability of an allocator capability to be used to make _sealed_ allocations. The thought here is that since sealed allocations cannot be freed via `heap_free_all`, it makes sense to draw a line between trusting a compartment to allocate with your quota and trusting it to ensure its own cleanup. I'm opening this up for discussion, but should be a very quick addition if we decide that's the right approach.

EDIT: We've decide for now to keep this last bit free. This may come in handy in dealing with some issues when it comes to implementing sub-quotas.

### DIFF
--- a/tests/allocator-test.cc
+++ b/tests/allocator-test.cc
@@ -5,6 +5,7 @@
 #define TEST_NAME "Allocator"
 
 #include "tests.hh"
+#include <allocator.h>
 #include <cheri.hh>
 #include <cheriot-atomic.hh>
 #include <cstdlib>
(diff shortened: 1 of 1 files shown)


davidchisnall:
A permission for sealed allocations is an interesting proposition, but I think most use cases for allowing someone else to allocate from your quota include sealed allocations.  I'd be somewhat more inclined to keep that permission for future use.

sdk/core/allocator/main.cc
```diff
@@ -387,18 +387,25 @@ namespace
...
+	 * Unseal an allocator capability and return it.
+	 *
+	 * Returns a pair, with `nullptr` as the first element if this is not a
+	 * heap capability.
+	 *
+	 * The permission bits from the sealed capability are extracted and returned
+	 * in the second element of the pair.
 	 */
-	PrivateAllocatorCapabilityState *
+	std::pair<PrivateAllocatorCapabilityState *, uint8_t>
```
davidchisnall: Minor nit, but making this a `uint8_t` will generate some explicit masking instructions, whereas making it an `int` will likely result in very slightly smaller code.

sdk/core/allocator/main.cc
```diff
@@ -823,13 +830,18 @@ namespace
...
 	{
-		auto *capability = malloc_capability_unseal(heapCapability);
+		auto [capability, quotaPermissions] =
+		  malloc_capability_unseal(heapCapability);
 		if (capability == nullptr)
 		{
 			Debug::log<DebugLevel::Warning>("Invalid heap capability {}",
 			                                heapCapability);
 			return -EPERM;
 		}
```
davidchisnall: Is this check necessary?  I believe all of the code paths that fail to unseal in `malloc_capability_unseal` will return 0 for permissions so the following check will fail.

It might be cleaner for `malloc_capability_unseal` to take the required permissions and return either null or an unsealed capability that has the required permissions, rather than a pair.

PR #618, opened 2025-10-31
Use user permissions on message queues.

davidchisnall:
This deprecates the old APIs for restricted endpoints, which added an indirection layer.  Now, sealed queue endpoint handles are capabilities with permissions for send, receive, and deallocated.

### DIFF
--- a/sdk/include/queue.h
+++ b/sdk/include/queue.h
@@ -24,11 +24,11 @@
 
 #include "cdefs.h"
 #include <multiwaiter.h>
-#include <stdatomic.h>
 #include <stddef.h>
 #include <stdint.h>
 #include <stdlib.h>
 #include <timeout.h>
+#include <token.h>
 
 /**
  * Structure representing a queue.  This structure represents the queue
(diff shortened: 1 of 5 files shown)


sdk/include/queue.h
```diff
@@ -207,11 +237,11 @@ int __cheri_compartment("message_queue")
...
- * Destroy a queue handle.  If this is called on a restricted endpoint
- * (returned either from `queue_receive_handle_create_sealed` or from
- * `queue_send_handle_create_sealed`), this frees only the handle.  If called
- * with the queue handle returned from `queue_create_sealed`, this will destroy
- * the queue.
+ * Destroy a queue.  This requires a handle with `MessageQueuePermitDestroy`
+ * permission.
+ *
+ * Returns 0 on success, `-EINVAL` if `queueHandle` is not a valid queue handle
+ * or lacks the relevant permissions.
```
rmn30, line 244: Why not `-EPERM` if lacking permissions?

rmn30: OK, having looked at the implementation I can understand why, but it is a bit odd.

rmn30 (approved):
I think this is fine. My only observation is that the specific error number returned by various functions is a bit unintuitive: often `-EINVAL` instead of `-EPERM`. I think this is because of the way `unseal` is implemented. It's not a new thing caused by this PR but perhaps we could think about fixing it,.

PR #619, opened 2025-11-02
Add cpp version of heap_allocate* that return ErrorOr and deprecate C versions for CPP

rmn30:


sdk/lib/event_group/event_group.cc
```diff
@@ -42,16 +43,16 @@ int eventgroup_create(Timeout            *timeout,
...
+	size_t size  = sizeof(EventGroup) + (threads * sizeof(EventWaiter));
+	auto   alloc = heap_allocate_cpp(timeout, heapCapability, size);
+	return alloc.either(
+	  [&](CHERI::Capability<void> pointer) {
+		  auto group         = new (pointer) EventGroup();
+		  *outGroup          = group;
+		  group->waiterCount = threads;
+		  return 0;
+	  },
+	  [&](int error) { return error; });
```
nwf: That lambda shouldn't need to capture anything, right?

```suggestion
	  [](int error) { return error; });
```

sdk/lib/event_group/event_group.cc
```diff
@@ -42,16 +43,16 @@ int eventgroup_create(Timeout            *timeout,
...
-	*outGroup = group;
-	if (!__builtin_cheri_tag_get(group))
-	{
-		return -ENOMEM;
-	}
-	group->waiterCount = threads;
-	return 0;
+	size_t size  = sizeof(EventGroup) + (threads * sizeof(EventWaiter));
+	auto   alloc = heap_allocate_cpp(timeout, heapCapability, size);
+	return alloc.either(
```
nwf, line 48: I wonder if this (`.either([...](auto c){ ...; return 0; }, [](int e){ return e; })`) is enough of a pattern that we should give it a name?  (Basically a twist on `.and_then(f)` with `f` returning `void`?)

sdk/lib/queue/queue.cc
```diff
@@ -448,18 +449,17 @@ int queue_create(Timeout              *timeout,
...
-
-	*outQueue = new (buffer.get()) MessageQueue(elementSize, elementCount);
-	return 0;
+	auto buffer = heap_allocate_cpp(timeout, heapCapability, allocSize);
+	return buffer.either(
+	  [&](Capability<void> pointer) {
+		  *outQueue = new (pointer) MessageQueue(elementSize, elementCount);
+		  return 0;
+	  },
+	  [&](int error) { return error; });
```
nwf: As above

tests/allocator-test.cc
```diff
@@ -332,56 +333,52 @@ namespace
...
 		TEST_EQUAL(heap_render(), 0, "heap_render returned non-zero");
 		debug_log("Trying a non-blocking allocation");
-		TEST_EQUAL(
-		  ErrorOr{heap_allocate(&noWait, MALLOC_CAPABILITY, BigAllocSize)}
-		    .as_error(),
-		  -ETIMEDOUT,
-		  "Non-blocking heap allocation did not fail with memory "
-		  "exhausted");
+		TEST_EQUAL(heap_allocate_cpp(&noWait, MALLOC_CAPABILITY, BigAllocSize)
+		             .as_error(),
```
nwf, line 337: Not needing all the `ErrorOr{}`s is really nice.

tests/allocator-test.cc
```diff
@@ -491,8 +487,10 @@ namespace
 		 *  small sample of pointers.
 		 */
 		auto doAlloc = [&](HeapTestState &heap, size_t sz) {
-			CHERI::Capability p{heap_allocate(&t, heap.quota, sz)};
-
+			auto a = heap_allocate_cpp(&t, heap.quota, sz);
+			// TEST(!a.is_error(), "Fuzz allocation failed {}=>{}", sz,
+			// a.as_error());
```
nwf, line 492: I don't _think_ we want allocation tests in the fuzzer to be a failure by default, but it might be nice to have it as a mode of running it when we know things about the heap and its other parameters?  (It's probably worth factoring the fuzzer out to its own `.cc` that can be built into the test harness or as a `tests.extra` or somesuch?)

tests/allocator-test.cc
```diff
@@ -643,7 +641,8 @@ namespace
 		size_t allocSize       = 128;
 		auto   mallocQuotaLeft = heap_quota_remaining(MALLOC_CAPABILITY);
 		CHERI::Capability alloc{
-		  heap_allocate(&noWait, MALLOC_CAPABILITY, allocSize)};
+		  heap_allocate_cpp(&noWait, MALLOC_CAPABILITY, allocSize)
+		    .as_raw_capability()};
```
nwf, line 645: Stylistic, but I think this could be an assignment, `alloc = heap_allocate_cpp(...).as_raw_capability();` rather than invoking the copy constructor?

nwf (commented):
This seems to me like a nice improvement.

sdk/lib/event_group/event_group.cc
rmn30: Yes, I was thinking the same. It only applies if you are converting from `ErrorOr` to `int` success/fail but potentially useful.

sdk/lib/queue/queue.cc
rmn30:
```suggestion
	  [](int error) { return error; });
```

davidchisnall:
Adding comments here that I passed on elsewhere:

I don't like this.  `heap_allocate` is a low-level allocation API.  In C, these are commonly used and implicit conversion from `void*` to other pointer types means that they are sensible.  The lack of RAII means that these are the *only* sensible way of managing memory.

In C++, things like the C++ Core Guidelines recommend *never* using raw allocation functions.  Allocation should be done via `std::make_shared`, `std::make_unique`, and so on.  The only code that *should* be calling `heap_allocate` is code that implements smart pointers, and these already need to be careful with error handling.

Having a different return type, or a different API, for C and C++ will cause developer friction.  The place I did this elsewhere was to make code *compatible* between C and C++: C permitted implicit conversion, C++ did not so the C++ wrapper preserved the type information.  This is a very different goal.

Note that template type deduction should work for `ErrorOr`, so you can just write `ErrorOr foo = heap_allocate(...)` (if it doesn't, we should add deduction guidelines), so this is really just letting you save typing three letters, in exchange for having APIs that differ between C and C++.  That's not a good tradeoff in developer experience.

We absolutely should be adding better smart pointers or alternatives to `std::make_unique` and `std::make_shared` that improve error handling.

PR #621, opened 2025-11-04
Use more consistent error codes in queue-compartment

rmn30:
Return `-EPERM` for permission errors and `-EINVAL` for invalid arguments. Make the internal `unseal` function use `ErrorOr` and a new overload of `and_then` that returns `int` to facilitate this.

### DIFF
--- a/sdk/lib/queue/queue_compartment.cc
+++ b/sdk/lib/queue/queue_compartment.cc
@@ -37,13 +37,35 @@ namespace
 	 * unsealed handle or null if the handle is invalid or lacks the relevant
 	 * permissions.
 	 */
-	MessageQueue *unseal(CHERI_SEALED(MessageQueue *) handle, int permissions)
+	ErrorOr<MessageQueue> unseal(CHERI_SEALED(MessageQueue *) handle,
+	                             int permissions)
 	{
 		if ((queue_permissions(handle) & permissions) != permissions)
 		{
-			return nullptr;
+			return -EPERM;
 		}
-		return token_unseal(handle_key(), Sealed<MessageQueue>{handle});
+		auto queue = token_unseal(handle_key(), Sealed<MessageQueue>{handle});
+		if (queue)
+		{
+			return queue;
+		}
+		return -EINVAL;
+	}
+
+	/**
+	 * Helper to run a function with the unsealed queue if the handle has the
+	 * given permissions.
+	 *
+	 * Returns either:
+	 *  - -EPERM if the handle does not have all the requested permissions
+	 *  - -EINVAL if the handle cannot be unsealed
+	 *  - The return value of `fn` otherwise
+	 */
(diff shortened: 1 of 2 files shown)


sdk/lib/queue/queue_compartment.cc
```diff
@@ -77,105 +83,106 @@ int queue_destroy_sealed(Timeout            *timeout,
...
-			return -EPERM;
-		}
-		queue_stop(queue);
-		return token_obj_destroy(heapCapability, handle_key(), queueHandle);
-	}
-	return -EINVAL;
+	// timeout unused?!
+	return unseal(queueHandle, MessageQueuePermitDestroy)
+	  .and_then_int([&](MessageQueue *queue) {
+		  if (heap_can_free(heapCapability, queueHandle) != 0)
```
nwf: Should that be `token_obj_can_destroy`?

sdk/include/cheri.hh
```diff
@@ -1762,6 +1762,18 @@ namespace CHERI
...
 		}
 
+		/**
+		 * Variant on `and_then` that returns an int: either the value returned
+		 * by the callback, or the error value held by `*this`. This is useful
+		 * for propagating error values in functions where that return an int
+		 * where negative values indicate errors and zero or positive values
+		 * indicate success.
+		 */
+		int and_then_int(auto &&function)
```
nwf: This can be another type-directed overload of `and_then`, like the `void`-returning case, I think:
```suggestion
		int and_then(auto &&function)
		    requires std::is_same_v<int, std::invoke_result_t<decltype(function), T *>>
```

sdk/lib/queue/queue_compartment.cc
```diff
@@ -77,105 +83,106 @@ int queue_destroy_sealed(Timeout            *timeout,
...
-		if (heap_can_free(heapCapability, queueHandle) != 0)
-		{
-			return -EPERM;
-		}
-		queue_stop(queue);
-		return token_obj_destroy(heapCapability, handle_key(), queueHandle);
-	}
-	return -EINVAL;
+	// timeout unused?!
+	return unseal(queueHandle, MessageQueuePermitDestroy)
```
nwf: Since every use of `unseal()` is then followed by `.and_then_int`, we might (skip the introduction of `and_then_int` and) build a more specialized combinator for local use?

```
int with_unsealed(CHERI_SEALED(MessageQueue *) handle, int permissions, auto &&then)
{
  return unseal(handle, permissions).either(then, [](int e) { return e; });
}
```
or perhaps
```
template <int permissions>
int with_unsealed(CHERI_SEALED(MessageQueue *) handle, auto &&then)
{
  return unseal(handle, permissions).either(then, [](int e) { return e; });
}
```

sdk/lib/queue/queue_compartment.cc
rmn30: I guess so?

PR #624, opened 2025-11-05
Add tests and fixes for heap_free edge cases

rmn30:
heap_free and heap_claim: test, document and fix some edge cases.
  
  1) Use the capability base to look up the chunk to heap_free instead
  of the address. This guarantees the address is within the bounds of
  the original allocation, unlike the address. Same goes for heap_claim.
  
  2) Test that an attempt to free with a 'one-past-the-end' zero length
  capability fails.
  
  3) Test that an attempt to free fails if the bounds are not the full
  allocation.
  
  4) Test that it's OK to free a pointer to the interior of an
  allocation if it still has the full bounds.
  
  5) Don't allow claim or free of sealed capabilities: there's no reason
  to claim a sealed capability and for free the correct API is to use
  token_obj_destroy with the correct sealing key. Add the necessary
  checks and tests for this. Note that freeing fails for a different reason when
  using the unsealed handle to a sealed token: the unsealed handle does
  not encompass the whole allocation (including header) so the free
  fails with a different error code. Note that taking a claim to a
  suballocation and subsequently releasing it using heap_free is allowed.

### DIFF
--- a/sdk/core/allocator/main.cc
+++ b/sdk/core/allocator/main.cc
@@ -831,14 +831,22 @@ namespace
 			                                heapCapability);
 			return -EPERM;
 		}
+		// Validate the pointer. We do not permit freeing sealed pointers.
 		Capability<void> mem{rawPointer};
-		if (!mem.is_valid())
+		if (!mem.is_valid() || mem.is_sealed())
 		{
 			return -EINVAL;
 		}
 		check_gm();
-		// Find the chunk that corresponds to this allocation.
-		auto *chunk = gm->allocation_start(mem.address());
+		/*
+		 * Find the chunk that corresponds to this allocation.  We use the base
+		 * as it cannot be further than one-past-the-end (because CHERI bounds
+		 * are monotone non-increasing). For us, one-past-the-end is the next
+		 * chunk's header which will have the revocation bit set meaning
+		 * allocation_start would return null. Note: `rawPointer`'s address is
+		 * not used at all!
+		 */
+		auto *chunk = gm->allocation_start(mem.base());
 		if (!chunk)
 		{
 			return -EINVAL;
@@ -943,12 +951,17 @@ __cheriot_minimum_stack(0x1c0) ssize_t
 		Debug::log<DebugLevel::Warning>("Invalid heap cap");
 		return -EPERM;
 	}
-	if (!Capability{pointer}.is_valid())
+	Capability p{pointer};
+	if (!p.is_valid() || p.is_sealed())
 	{
 		Debug::log<DebugLevel::Warning>("Invalid claimed cap");
 		return -EINVAL;
 	}
-	auto *chunk = gm->allocation_start(Capability{pointer}.address());
+	/*
+	 * As with heap_free we use the base to lookup the chunk (see comments in
+	 * heap_free_internal).
+	 */
+	auto *chunk = gm->allocation_start(p.base());
 	if (chunk == nullptr)
 	{
 		Debug::log<DebugLevel::Warning>("chunk not found");
--- a/sdk/include/stdlib.h
+++ b/sdk/include/stdlib.h
@@ -221,9 +221,12 @@ void *__cheri_compartment("allocator")
                       uint32_t flags      __if_cxx(= AllocateWaitAny));
 
 /**
- * Add a claim to an allocation.  The object will be counted against the quota
- * provided by the first argument until a corresponding call to `heap_free`.
- * Note that this can be used with interior pointers.
+ * Add a claim to an allocation.  As with `heap_free` the *base* of the
+ * capability is used to find the allocation, not the address.
+ *
+ * The object will be counted against the quota provided by the first argument
+ * until a corresponding call to `heap_free`. Note that this can be used with
+ * interior pointers.
  *
  * This will return the size of the allocation claimed on success (which may be
  * larger than the size requested in the original `heap_allocate` call; see its
@@ -233,8 +236,8 @@ void *__cheri_compartment("allocator")
  * quota is too small to accomodate the claim, and `-EINVAL` if `pointer` is not
  * a valid pointer into a live heap allocation.
  *
- * Similarly to `heap_allocate`, `-ENOTENOUGHSTACK` may be returned if the
- * stack is insufficiently large to run the function. See `heap_allocate`.
+ * Similarly to `heap_allocate`, `-ENOTENOUGHSTACK` may be returned if the stack
+ * is insufficiently large to run the function. See `heap_allocate`.
  */
 ssize_t __cheri_compartment("allocator")
   heap_claim(AllocatorCapability heapCapability, void *pointer);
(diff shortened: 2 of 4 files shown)


tests/allocator-test.cc
```diff
@@ -977,6 +977,22 @@ namespace
...
 
+		// It should not be possible to free sealed capabilities
+		TEST_EQUAL(heap_can_free(SECOND_HEAP, sealedPointer),
+		           -EINVAL,
+		           "heap_can_free should return error for sealed capability");
+		TEST_EQUAL(heap_free(SECOND_HEAP, sealedPointer),
+		           -EINVAL,
+		           "heap_free should return error for sealed capability");
+		// It should not be possible to free unsealed capabilities to sealed
+		// tokens (because the address doesn't include the header)
```
nwf: "the header" is maybe a bit too succinct, since there are several headers involved in allocation.  "the TokenObjectType header" perhaps?

nwf (approved):
One nit but LGTM

sdk/include/stdlib.h
```diff
@@ -272,11 +272,17 @@ heap_claim_fast(Timeout         *timeout,
 }
 
 /**
- * Free a heap allocation.
+ * Free a heap allocation or release a claim made by `heap_claim`.
```
nwf: :+1:

sdk/include/stdlib.h
```diff
@@ -272,11 +272,17 @@ heap_claim_fast(Timeout         *timeout,
...
  *
- * Returns 0 on success, `-EINVAL` if `ptr` is not a valid pointer to the start
- * of a live heap allocation, or `-ENOTENOUGHSTACK` if the stack size is
- * insufficiently large to safely run the function.
+ * Returns:
+ *  - 0 on success
+ *  - `-EINVAL` if `heapCapability` is invalid or if  `ptr` is untagged, sealed,
+ *    is not a heap pointer or its address is outside its bounds.
```
nwf:
```suggestion
 *  - `-EINVAL` if `heapCapability` is invalid or if  `ptr` is untagged, is sealed,
 *    is not a heap pointer, or has its address is outside its bounds.
```

Nit-picky.

tests/allocator-test.cc
```diff
@@ -87,6 +87,65 @@ namespace
...
 	}
 
+	__noinline void test_heap_free()
+	{
+		debug_log("Testing heap_free edge cases");
+		Timeout          t{5};
+		Capability<void> p1 = heap_allocate(&t, MALLOC_CAPABILITY, 16);
+		TEST(p1.is_valid(), "Unable to make allocation for interior free test");
+
+		// Attempt to free with wrong heap capability
```
nwf: How had we not tested that already?

tests/allocator-test.cc
```diff
@@ -87,6 +87,65 @@ namespace
...
+		  -EPERM,
+		  "attempt to free with wrong heap capability did not fail with "
+		  "-EPERM");
+		TEST(p1.is_valid_temporal(),
+		     "pointer invalid after failed free attempt with wrong heap");
+
+		// Attempt to free a pointer with (just) out of bounds address
+		Capability<void> p2 = p1;
+		p2.address()        = p2.top();
+		TEST(p2.is_valid(), "pointer invalid after setting address to top");
```
nwf: "This (`!p2.is_valid()`) can't happen because CHERI capabilities always guarantee one-past-the-end is tagged for C language support" or something like that?

tests/allocator-test.cc
```diff
@@ -1005,6 +1066,11 @@ namespace
 		TEST(canFree != 0,
 		     "Should not be able to free a sealed capability with the wrong "
 		     "malloc capability but succeeded");
+		TEST_EQUAL(
+		  token_obj_destroy(
+		    MALLOC_CAPABILITY, sealingCapability, sealedPointer),
+		  -EPERM,
+		  "Freeing a sealed capability with the wrong allocator succeeded");
```
nwf, line 1083: "wrong quota"?

tests/allocator-test.cc
rmn30: You mean the interior pointer free? It's possible we already do -- there are a lot of allocator tests! Actually I just noticed one of the tests I added is a duplicate (free with wrong quota) -- will remove that.

tests/allocator-test.cc
rmn30: Added a comment.

tests/allocator-test.cc
rmn30: Removed this test as it's a duplicate.

sdk/core/allocator/main.cc
```diff
@@ -832,13 +832,23 @@ namespace
 			return -EPERM;
 		}
 		Capability<void> mem{rawPointer};
-		if (!mem.is_valid())
+		/*
+		 * Validate the pointer. We do not permit freeing sealed pointers, or
+		 * pointers that are out of bounds. Note that address >= base is implied
+		 * by the combination of valid tag and CHERIoT capability encoding.
+		 */
+		if (!mem.is_valid() || mem.is_sealed() || mem.address() >= mem.top())
```
davidchisnall: Rather than checking address < top, I think it would be cleaner above to do:

`mem.address() = mem.base();`

This ensures that any future use of address is sensible.

sdk/core/allocator/main.cc
```diff
@@ -832,13 +832,23 @@ namespace
...
 		check_gm();
-		// Find the chunk that corresponds to this allocation.
-		auto *chunk = gm->allocation_start(mem.address());
+		/*
+		 * Find the chunk that corresponds to this allocation.  Use the base as
+		 * that cannot be further than one-past-the-end (because CHERI bounds
+		 * are monotone non-increasing), for us one-past-the-end is the next
+		 * chunk's header.
+		 */
+		auto *chunk = gm->allocation_start(mem.base());
```
davidchisnall, line 849: With the above suggestions, this change is unnecessary.

sdk/core/allocator/main.cc
rmn30: That would prohibit freeing any interior pointer which is a change from the current behaviour although it is UB anyway so up to us. Freeing a claim on a sub allocation would still be possible provided `address==base`.

sdk/core/allocator/main.cc
rmn30: I agree that either `base` or `address` is fine here provided we have one of the checks on `address`. Note that `address < top` is actually slightly stricter than `base==address` because the latter allows a zero length capability to the header of the next allocation with `base==address==top`. As per @nwf 's comment that is OK.

sdk/core/allocator/main.cc
rmn30: OK, I misunderstood and thought you were suggesting to check that `mem.address == mem.base`. I think setting `mem.address` to `mem.base` is equivalent to using `mem.base` to look up the chunk below because the address of `mem` is not used otherwise. Setting the address to base would take an extra instruction vs. just calling `allocation_start` with the base so can't we just do that?

PR #632, opened 2025-12-04
Allow large stacks.

davidchisnall:
This requires a CHERIoT-Platform/llvm-project#289 to work (without that, the stacks are insufficiently aligned and the loader refuses to load them).

With this, you can have stacks up to 64 KiB.  Tested with the PQC code with a 60 KiB stack.  More than 64 KiB would require a small loader change to not use the 16-bit size (and an associated linker change to not refuse to link things with larger stacks).

The loader changes need careful review.  I believe I have checked the important things, specifically:

 - Stacks are restored using the old (pre-adjustment) stack pointer.
 - Spills happen before the stack pointer is moved down.
 - The callee stack check happens with the size of the rounded-down stack.

All of these should be confirmed in review.

### DIFF
--- a/sdk/core/switcher/entry.S
+++ b/sdk/core/switcher/entry.S
@@ -451,13 +451,18 @@ __Z26compartment_switcher_entryz:
 	cgetbase           s1, csp
 	csetaddr           csp, csp, s1
 	sub                s1, s0, s1
-	csetboundsexact    ct2, csp, s1
+	csetboundsrounddown ct2, csp, s1
+	cgettop            s0, ct2
+	// Round down the top address to make sure it is still 16-byte aligned.
+	// This is *probably* not needed, but we need to do a proof on the
+	// csetboundsrounddown behaviour before removing this.  If it is removed,
+	// please link to the proof and leave the following line commented out.
+	andi               s0, s0, 0xfffffff0
 	csetaddr           csp, ct2, s0
 	/*
 	 * Atlas:
-	 *  s0: address of stack boundary between caller and callee frames, that is,
-	 *      the lowest address of the register spill from
-	 *      .Lswitch_entry_first_spill)
+	 *  s0: address of stack boundary between caller and callee frames,
+	 *      possibly rounded down to preserve alignment.
 	 *  sp: pointer to stack, with its limit and address set to the address in
 	 *      s0.  The base and permissions have not been altered from sp at
 	 *      entry, and the tag remains set since all manipulations have been
--- a/sdk/xmake.lua
+++ b/sdk/xmake.lua
@@ -834,9 +834,11 @@ rule("cheriot.firmware")
 				"\n\t\tSHORT(.thread_${thread_id}_trusted_stack_end - .thread_${thread_id}_trusted_stack_start);" ..
 				"\n\n"
 
-		-- Stacks must be less than this size or truncating them in compartment
-		-- switch may encounter precision errors.
-		local stack_size_limit = 8176
+		-- Stacks must be less than this size or we cannot express them in the
+		-- loader metadata.  The checks in lld for valid stacks currently
+		-- reject anything larger than this, so provide a helpful error here,
+		-- rather than an unhelpful one later.
+		local stack_size_limit = 65280
 
 		-- Initial pass through thread sequence to derive values within each
 		local thread_priorities_set = {}


sdk/xmake.lua
```diff
@@ -847,12 +843,6 @@ rule("cheriot.firmware")
 			-- loader will fail.  If it is too big, we waste space.
 			thread.trusted_stack_size = loader_trusted_stack_size + (24 * thread.trusted_stack_frames)
 
-			if thread.stack_size > stack_size_limit then
-				raise("thread " .. i .. " requested a " .. thread.stack_size ..
-				" stack.  Stacks over " .. stack_size_limit ..
-				" are not yet supported in the compartment switcher.")
-			end
```
rmn30, line 854: Why did you remove the check and not update it with the new, 64 KiB size limit?

rmn30 (commented):
Leaving @nwf to review the switcher bits as they may want to run their special tooling.

sdk/xmake.lua
davidchisnall: I removed the check then discovered the new limit.  I'll add it back.

sdk/xmake.lua
```diff
@@ -836,7 +836,7 @@ rule("cheriot.firmware")
 
 		-- Stacks must be less than this size or truncating them in compartment
 		-- switch may encounter precision errors.
```
rmn30, line 838: This comment is now out of date.

davidchisnall: Fixed.

sdk/core/switcher/entry.S
```diff
@@ -451,13 +451,13 @@ __Z26compartment_switcher_entryz:
...
+	csetboundsrounddown ct2, csp, s1
+	cgettop            s0, ct2
 	csetaddr           csp, ct2, s0
 	/*
 	 * Atlas:
-	 *  s0: address of stack boundary between caller and callee frames, that is,
-	 *      the lowest address of the register spill from
-	 *      .Lswitch_entry_first_spill)
+	 *  s0: address of stack boundary between caller and callee frames,
+	 *      possibly rounded down to preserve alignment.
```
nwf, line 465:
```suggestion
	 *      possibly rounded down to preserve representability.
```

We don't yet have a proof that csetboundsrounddown is guaranteed to result in a top that is at least as aligned as the base and requested length (for exponents 14 or less, anyway).  (That is, that `forall cap, length . (cap.exp <= 14 & ctz(length) <= 14) -> ctz(csetboundsrounddown(cap, length).top) >= min{ctz(cap.base), ctz(length)}`.)  That _ought_ to be true...

davidchisnall: Hmm, for software to not die it also needs to preserve alignment.  I wonder if there are cases where the rounds down for an aligned top does not result in an aligned base.  I have added an and to guarantee the alignment.  We can remove this if someone does the right proof.

rmn30: For `CSetBoundsRoundDown` the resulting base is always exactly as requested. In this case it should be the stack base, which is necessarily well aligned for the original stack size. Given that the requested length will be a multiple of 16 then either:

1. the new length is large enough that we must use an E > 4 that gives coarser than 16-byte precision. The requested length will either be exactly representable using this E or it will be rounded down to the largest aligned length less than the requested one. In either case the top will be at least 2**E aligned so is definitely 16-byte aligned.
2. The new length is small enough that it allows 16-byte precision. In this case we will get exact bounds so no alignment issues.

There is a third case that shouldn't normally happen:

3. A malicious caller could set the stack base to some unaligned value. This could force `CSetBoundsRoundDown` to use a smaller exponent than can represent the requested length. This would result in a max length capability of that exponent. Fortunately we have a check on entry that the stack base and address are at least 16-byte aligned so we're guaranteed that even in this case we'll get at least 16-byte alignment.

davidchisnall: @nwf points out elsewhere [this is where we check that the stack base is aligned](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/ba53563686177fafb32e6d7231dc3dccfc65d500/sdk/core/switcher/entry.S#L313-L320).

davidchisnall: It would be nice to have a proof somewhere (Sail?) that the xor rounding down the address that we set the stack pointer to is unnecessary, but i'm also happy to have it there.

nwf: My attempt at a sail SMT proof that CSetBoundsRoundDown's resulting length is at least as aligned as the minimum of the alignments of the requested base and requested length is stalled on https://github.com/rems-project/sail/issues/1562 .

PR #634, opened 2025-12-11
Add rules to compile standalone Rust files and Rust crates

xdoardo:
The examples work locally (removed in [6903d8a](https://github.com/CHERIoT-Platform/cheriot-rtos/pull/634/commits/6903d8a9265977e4f14c8a25d8a8b8774755a76a) to make CI pass), but can't currently be reproduced in CI due to, of course, cheri-enabled `rustc` (and `cargo` I suppose) missing from the devcontainer.

This PR adds two new xmake rules to the RTOS: the first is to build standalone Rust files, and the other is to build Rust crates.

#### Adding standalone Rust files
Should be pretty straightforward. From the example: 

```lua
compartment("hello")
	add_files("rust.rs")
	...
```
Running `xmake config --rc=<path to CHERIoT-enabled rustc> --sdk=...` will check that the given `rustc` supports the CHERIoT target. Internally, the rule builds the standalone file as a static library and adds it to the list of objectfiles for the target.

#### Adding Rust crates
Still straightforward, needs a bit of a trick:
```lua
compartment("run_rust"
	add_files("./hello_world/Cargo.toml", { sourcekind = "cheriot_rust_crate" })
	-- Setting the sourcekind is necessary so that rules for toml files aren't taken into account.
```
Running `xmake config --rc=<path to CHERIoT-enabled rustc> --sdk=...` will check that the given `rustc` supports the CHERIoT target and that `cargo` is available. The `cargo` binary installed on the system is not required to be "cheri-aware"; it can also be the default from a package manager, since it then uses the `rustc` given with the `rc` flag. 

This rule requires, however, that the static library lib-type for the crate is set in the `Cargo.toml`, and is not passed as a flag to `rustc` (because that would otherwise build also every dependency as a static library, which is not what we want).

### DIFF
--- a/sdk/xmake.lua
+++ b/sdk/xmake.lua
@@ -228,6 +228,13 @@ toolchain("cheriot-clang", function ()
 		self:add("cxflags", default_clang_flags)
 		-- Assembly flags
 		self:add("asflags", default_clang_flags)
+
+		-- Rust flags
+		local default_rc_flags = {
+			"--target=" .. target,
+			"-Ctarget-cpu=" .. cpu,
+		}
+		self:add("rcflags", default_rc_flags)
 	end)
 end)
 
@@ -309,6 +316,8 @@ end)
 -- Common rules for any CHERIoT component (library or compartment)
 rule("cheriot.component")
 	add_deps("cheriot.toolchain")
+
+	add_deps("cheriot.rust", "cheriot.rust.crate")
 
 	-- Set some default config values for all cheriot components.
 	on_load(function (target)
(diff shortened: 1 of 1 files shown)


sdk/xmake.lua
```diff
@@ -1268,5 +1268,228 @@ function compartment(name)
...
+	set_extensions(".rs")
+	on_config(function(target)
+		local rc = target:compiler("rc")
+		local outdata, errdata = os.iorunv(rc:program(), { "--print=target-list" })
+		assert(not errdata or errdata == "", "failed to run `" .. rc:program() .. " --print=target-list':\n" .. errdata)
+
+		local cheriot_target = "riscv32cheriot-unknown-cheriotrtos"
+		local cheriot_target_needle = string.gsub(cheriot_target, "%-", "%%-")
+		local result = string.match(outdata:trim(), cheriot_target_needle)
+
```
davidchisnall: This looks like something that should live in a toolchain definition, not a build rule?

sdk/xmake.lua
```diff
@@ -1268,5 +1268,228 @@ function compartment(name)
...
+		-- path/to/rust.rs -> libpath_to_rust
+		local libname = "lib" .. string.gsub(sourcefile:sub(1, #sourcefile - 3), "//", "_")
+		local targetfile = target:objectfile(libname)
+		-- <buildir>/libpath_to_rust.o -> <buildir>/libpath_to_rust.a
+		local targetfile = targetfile:sub(1, #targetfile - 1) .. "a"
+		table.insert(target:objectfiles(), targetfile)
+
+		local compinst = compiler.load("rc", { target = target })
+		local compflags = compinst:compflags({ target = target })
```
davidchisnall, line 1551: We probably also need to add the flags for RISC-V extensions to the build flags, or we'll end up always targeting the minimum CHERIoT target for Rust.

sdk/xmake.lua
xdoardo: [a89a88a](https://github.com/CHERIoT-Platform/cheriot-rtos/pull/634/commits/a89a88a9bfd8f87d7c0d42bb6d507deb25020606) adds the `--target=riscv32cheriot-unknown-cheriotrtos` flag to the `rcflags` directly in the toolchain definition.

This is because [74c979b](https://github.com/CHERIoT-Platform/cheriot-rtos/pull/634/commits/74c979bf303e920cfc18a54a4cf0df563b176212) removes the `on_config` callbacks for the rust rules which, when added by default, were executed multiple times (thus checking that rustc works as we want multiple times).

Now, instead, the compilation will fail if `rc` isn't set or `cargo` isn't found.

sdk/xmake.lua
xdoardo: [6e9d6f1](https://github.com/CHERIoT-Platform/cheriot-rtos/pull/634/commits/6e9d6f15b65c65ff7fa2b67d382b8705e468b929) attempts to do that. I am not sure it covers all the board-flag configurations.

sdk/xmake.lua
```diff
@@ -1731,5 +1735,184 @@ function compartment(name)
 		add_deps("cheriot.board.interrupts")
 end
 
-includes("lib/")
+-- Helper to synthesize cpu-specific Rust flags for the given board.
+local rust_flags_for_board = function(board)
+	if board:match("sonata") or board:match("ibex") then
+		return { "+b" }
+	end
```
davidchisnall: This should come from the board description file, we shouldn't ever match on the board name here.  This will break for any derived board file.

xdoardo: [250aeb4](https://github.com/CHERIoT-Platform/cheriot-rtos/pull/634/commits/250aeb4fd16a3e2bf019efb49111e0ae5e4df384)

sdk/xmake.lua
```diff
@@ -1512,5 +1521,169 @@ function compartment(name)
...
+
+		io.flush()
+
+		dependinfo.files = {}
+		local flags = table.join("--crate-type=staticlib", compflags, "-o", targetfile, sourcefile)
+		if is_mode("release") then
+			table.insert(flags, "-Copt-level=z")
+		else
+			table.insert(flags, "-Copt-level=0")
+		end
```
nwf: FWIW, for C/C++, we almost always build `-Oz`, regardless of release/debug mode.  The loader C++ gets built `-O1` to be a little easier to follow in tracing and is deleted after it's done, so the code size matters less.

sdk/xmake.lua
```diff
@@ -1512,5 +1521,169 @@ function compartment(name)
...
 end
 
-includes("lib/")
+-- Rules for standalone Rust source files.
+-- Note that to make the implementation of this rule not too complex single source files are compiled to static libraries,
+-- which will also contain the the parts of Rust's standard library that CHERIoT supports.
+--
+-- It expects the user to pass a CHERIoT-enabled `rustc` with the `--rc` flag to `xmake config`.
+rule("cheriot.rust", function()
+	set_extensions(".rs")
```
nwf, line 1530: xmake question: given the `set_extensions` here, do we need this rule to be a dependency of the `cheriot.rust` rule or will it just get picked up by default?

sdk/xmake.lua
```diff
@@ -1512,5 +1521,169 @@ function compartment(name)
...
+		assert(errdata == nil or errdata == "" or errdata:match("Finished") or (not errdata:match("error")),
+			"failed to compile  " .. sourcefile .. ":\n" .. errdata)
+
+		dependinfo.values = depvalues
+		table.insert(dependinfo.files, sourcefile)
+		depend.save(dependinfo, dependfile)
+	end)
+end)
+
+-- Hack to override the default rule for Rust, which is not sensible for our use-case.
```
nwf, line 1579: This is _probably_ fine, but does it make the (purely hypothetical) case of users building a rust program on the host at the same time as they build a firmware?  (Say, a rust program to generate something used in the firmware build.)

sdk/xmake.lua
```diff
@@ -1512,5 +1521,169 @@ function compartment(name)
 		add_deps("cheriot.board.interrupts")
 end
 
-includes("lib/")
+-- Rules for standalone Rust source files.
+-- Note that to make the implementation of this rule not too complex single source files are compiled to static libraries,
+-- which will also contain the the parts of Rust's standard library that CHERIoT supports.
+--
+-- It expects the user to pass a CHERIoT-enabled `rustc` with the `--rc` flag to `xmake config`.
+rule("cheriot.rust", function()
```
nwf, line 1529: Forgive my ignorance, but what's in this rule that wouldn't happen if I used the default xmake rust language (https://github.com/xmake-io/xmake/blob/821545f6a78188e733b0abe84cdfcb6d07c9e082/xmake/languages/rust/xmake.lua) and a `target()` with `set_kind("static")`?

sdk/xmake.lua
xdoardo: f954034 removes those lines, and makes the script always use `-Oz` instead.

sdk/xmake.lua
xdoardo: If you mean the `cheriot.rust.crate` rule, then we don't want it to be a dependency because the rule for crates just calls `cargo` passing the correct `RUSTC` flag to it. Thus, `cargo` is the one that actually takes care of the `.rs` files and not `xmake`, technically.

sdk/xmake.lua
nwf: Having spent a day arguing with xmake and gotten nothing to show for it, I think the answer is "nothing good".  Apologies for the noise.

sdk/xmake.lua
xdoardo: It does, yes. But, then again, the normal Rust rule does things which I think aren't well thought out as the default rules for C might be. So I think that, at least for now, not using (or forcing the disuse of) the default Rust rule is our best chance to have something relatively understandable and maintainable.

nwf (approved):
I don't love it, but for reasons that are entirely `xmake`'s fault.  Given the mess that is `xmake` (and not just our use of `xmake`), I think this is probably the best we're going to get and it has the distinct advantage of working _now_.

PR #638, opened 2025-12-12
Add check to the loader for cap_relocs to privileged compartments

xdoardo:
This PR adds a check to the loader to inform the user that cap_relocs to privileged compartments were found. These, by default, fail because the `.. image.libraries_and_compartments() ..` [here](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/main/sdk/core/loader/boot.cc#L1031) does _not_ iterate over privileged compartments.

For example, adding something like 
```
static auto foo = "hello";
const char** bar[] = {&foo};
...
<use bar>
...
```
to `allocator/main.cc` makes the loader hit [this assertion](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/main/sdk/core/loader/boot.cc#L1039).

For clarity, this patch *does not* allow such relocations, it just adds a more explicative message to the loader.
This check is executed only when `--debug-loader=y` is passed to `xmake config`, since all the code is gated under the definition and non-nullity of `DEBUG_LOADER`, which is set [here](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/main/sdk/xmake.lua#L1168).

### DIFF
--- a/sdk/core/loader/boot.cc
+++ b/sdk/core/loader/boot.cc
@@ -1050,17 +1050,17 @@ namespace
 
 			if constexpr (DebugLoader)
 			{
-				auto error = false;
+				auto hasSeenForbiddenCapRelocs = false;
 				for (auto &pC : image.privilegedCompartments)
 				{
 					if (contains(pC.code, reloc.addr) ||
 					    contains(pC.data, reloc.addr))
 					{
-						error = true;
+						hasSeenForbiddenCapRelocs = true;
 						Debug::log(
 						  "Capreloc with address {} should be applied to "
 						  "a privileged compartment with code region: {} "
-						  "- {} and data region: ",
+						  "- {} and data region: {} - {}",
 						  reloc.addr,
 						  pC.code.start(),
 						  pC.code.start() + pC.code.size(),
@@ -1069,10 +1069,9 @@ namespace
 					}
 				}
 
-				Debug::Invariant(!error,
+				Debug::Invariant(!hasSeenForbiddenCapRelocs,
 				                 "Encountered forbidden relocations to "
 				                 "privileged compartment(s)");
-				__builtin_unreachable();
 			}
 
 			// Find the compartment that this relocation applies to.


sdk/core/loader/boot.cc
```diff
@@ -1047,6 +1047,35 @@ namespace
 		{
 			Debug::log(
 			  "Capreloc address: {}, base: {}", reloc.addr, reloc.base);
+
+#if defined DEBUG_LOADER && DEBUG_LOADER
```
davidchisnall: Please use constexpr if, not `#ifdef` for conditional code.  

Also, this formulation is entirely UB (the C preprocessor does not do short-circuit evaluation).

sdk/core/loader/boot.cc
```diff
@@ -1047,6 +1047,35 @@ namespace
...
+				}
+			}
+
+			if (error)
+			{
+				Debug::Invariant(false,
+				                 "Encountered forbidden relocations to "
+				                 "privileged compartment(s)");
+				__builtin_unreachable();
+			}
```
davidchisnall:
```suggestion
			Debug::Invariant(!error,
						     "Encountered forbidden relocations to "
				             "privileged compartment(s)");
```

xdoardo: Oh my. Sorry. Changed in [468ba58](https://github.com/CHERIoT-Platform/cheriot-rtos/pull/638/commits/468ba58bc41bf9882db7dbe15cd2d31e84d25f8b).

sdk/core/loader/boot.cc
xdoardo: Changed in [77dae23](https://github.com/CHERIoT-Platform/cheriot-rtos/pull/638/commits/77dae23213061eb9cb6b395da863dfc8f270818f). `DebugLoader` is defined [here](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/main/sdk/core/loader/debug.hh#L16).

sdk/core/loader/boot.cc
xdoardo: (call to `__builtin_unreachable()` removed here: https://github.com/CHERIoT-Platform/cheriot-rtos/pull/638/commits/529b8341decd48b7efe36035dd19093e6549ba63)

PR #639, opened 2026-01-05
Fix missing return type in  `__atomic_load_cap` libcall declaration

jacobpake:
Fixes typo in libcall declaration, should match signature:

```
void *__atomic_load_cap(void *const *ptr, int)
```


davidchisnall (approved):
Ooops, thanks!

PR #641, opened 2026-01-05
Add platform headers for Ibex's proprietary CSRs

nwf:
This is sat atop #558 and uses that machinery.

### DIFF
--- a/sdk/include/platform/ibex/platform-custom_csr.hh
+++ b/sdk/include/platform/ibex/platform-custom_csr.hh
@@ -1,0 +1,68 @@
+// Copyright Microsoft and CHERIoT Contributors.
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+/*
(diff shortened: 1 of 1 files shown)


nwf:
This is now rebased and ready to be merged, I think!

davidchisnall:
I'm a little bit uncertain about this because things in platform- are supposed to be platform *abstraction* layers, rather than just platform-specific things.  If we're going to put custom CSRs there, we should probably have a concept and a discovery mechanism, so we can ask constexpr questions such as:

 - Does the current CPU have this custom CSR?
 - Does the current CPU have a CSR for this functionality, and, if so, what is it and how do I set it?

nwf:
Hm.  I'm happy for this to be somewhere else, if we have / can add such a thing, since I'm not sure what the abstraction looks like, having only a single example from which to attempt generalization.

nwf:
Ah, and we need to rev the devcontainer to include the Kudu fix.

PR #642, opened 2026-01-06
loader: package as object for auditing

nwf:
Arrange for there to be a single file for "the loader" rather than it just being an "object"-kind xmake target whose objects get used in dependent targets' link phases.  This makes it slightly easier to reason about and investigate "the loader" as a thing.

### DIFF
--- a/sdk/core/loader/boot.S
+++ b/sdk/core/loader/boot.S
@@ -155,7 +155,7 @@
 	la_abs			s1, __thread_count
 	csetaddr		cs1, ca4, s1
 	clhu			s1, 0(cs1)
-	li			t0, -BOOT_THREADINFO_SZ
+	li			t0, -ThreadLoaderInfo_size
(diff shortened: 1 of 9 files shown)


davidchisnall:
I'd really like to never have .a files anywhere near our build flow.  If we need something like this, can we do `ld -r` and produce a .o file, like people who don't need to fit their linker in 64 KiB of RAM?

davidchisnall:
This needs rebasing after the refactor for the initialisation code.

davidchisnall (commented):
The changes look okay, as long as CI is happy (especially: with the compartment-initialisation tests), I'm happy.

nwf:
> This needs rebasing after the refactor for the initialisation code.

Whoop, I had that internally, just hadn't pushed it up.  Thanks for the reminder.

davidchisnall (approved):

PR #646, opened 2026-01-26
Only define away compartment name annotations for clang-tiday and bare-metal when no compartment name was specified to the compiler.

resistor:


davidchisnall (approved):

nwf (approved):
Seems sensible to me.

PR #649, opened 2026-01-29
Remove the cross-thread interrupt mechanism.

### LINKED ISSUE #372: switcher cross-thread signaling can race the cross-call and -return paths
While writing #358, working towards #47, and thinking more about the [`switcher_interrupt_thread`](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/de2db172fd5767566aee89f756f6f7c5cdefda89/sdk/core/switcher/entry.S#L1842) API, I realized it's possible to induce force unwinds rather than error handler invocation.  The cross-call path can be preemptive (both [before](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/de2db172fd5767566aee89f756f6f7c5cdefda89/sdk/core/switcher/entry.S#L585) and [after](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/de2db172fd5767566aee89f756f6f7c5cdefda89/sdk/core/switcher/entry.S#L618) the call), and can be off core and "within" the same compartment as the caller of `switcher_interrupt_thread`.

Prior to [the actual call at `.Lswitch_callee_call`](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/de2db172fd5767566aee89f756f6f7c5cdefda89/sdk/core/switcher/entry.S#L609), losing this race causes the call to be skipped, as [coming back on core](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/de2db172fd5767566aee89f756f6f7c5cdefda89/sdk/core/switcher/entry.S#L982) will see the injected error, then see that it is _in the switcher_, and, ultimately, jump ahead to [`switcher_after_compartment_call`](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/de2db172fd5767566aee89f756f6f7c5cdefda89/sdk/core/switcher/entry.S#L612).

*After* the actual call, things are a little more exciting, even if the control flow proceeds similarly:
1. we might just clobber the return value of a completed call, if we're before [the trusted stack frame update](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/de2db172fd5767566aee89f756f6f7c5cdefda89/sdk/core/switcher/entry.S#L711), because the instruction sequence is idempotent up to that point, or, more excitingly...
2. we might forcibly unwind out of the caller entirely, if we're after that point (just because we happened to be returning from a cross call at the time of being interrupted).

Probably we fix this at the same time as we fix #47 and maybe we do so in part by revising the interrupt mechanism (introduced, for reference, by #114).  It almost certainly means that the cross-call and cross-return paths will involve short sequences of IRQ-deferred code, and it may be more convenient to use `mret` rather than `cjalr` to exit the TCB so that we can pick up atomic changes in IRQ disposition.

davidchisnall:
We eventually want *something like* this, but not this.

It exists in the git history if we want to resurrect it later, but currently it's adding TCB complexity for a thing no one is using.  We should design something better *with some actual use case implemented* and then implement that thing.

Fixes #372


davidchisnall:
Yes, we should open an issue to track the desire to do this kind of thing (in particular, some kind of watchdog and for context switching of virtualised other operating systems).

PR #651, opened 2026-01-30
Fix an issue where `heap_free_all` would free claims.

davidchisnall:
In some circumstances, claims would be allocated earlier in the heap than the object that they referred to.  When calling `heap_free_all`, this would result in the claims themselves being deallocated.  This corrupted the claims list and typically put the allocator into an infinite loop (or caused it to crash) when it reached the object against which the claim was held.

This change reserves allocator ID 0 to signify allocator-owned objects. The allocator will still account claims to the quota that requires them but they will now be marked in the heap as owned by the allocator.  This prevents them from being freed by any external caller.

Claims already track the quota ID of the quota that they refer to, so storing that in the object header was redundant.

### DIFF
--- a/sdk/core/allocator/main.cc
+++ b/sdk/core/allocator/main.cc
@@ -274,6 +274,10 @@ namespace
 	                              uint32_t flags              = AllocateWaitAny)
 	{
 		check_gm();
+		Debug::Assert(
+		  capability->identifier != QuotaIdentifierAllocatorOwned,
+		  "Attempting to allocate a normal object with the allocator-owned "
+		  "identifier.  This should be impossible.");
 
 		do
 		{
@@ -411,6 +415,9 @@ namespace
 		  reinterpret_cast<PrivateAllocatorCapabilityState *>(capability);
 
 		// Assign an identifier if this is the first time that we've seen this.
+		// The value of 0 is used in allocator capabilities as a marker for
+		// uninitialised capabilities.  It is used internally in the heap for
+		// objects that are owned by the allocator.
 		if (state->identifier == 0)
 		{
 			static uint32_t nextIdentifier = 1;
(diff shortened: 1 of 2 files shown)


sdk/core/allocator/main.cc
```diff
@@ -583,8 +586,15 @@ namespace
...
-			return new (std::get<Capability<void>>(space))
-			  Claim(capability.identifier, next);
+			auto claim = std::get<Capability<void>>(space);
+			// Set the owner of the claim to 0.  The ID zero is reserved for
+			// allocator-owned structures.  The claim object stores the ID of
+			// the quota that this allocation should be returned to.
+			Capability heap{gm->heapStart};
+			heap.address() = claim.address();
+			auto chunk     = MChunkHeader::from_body(heap);
+			chunk->set_owner(0);
```
rmn30: Could you simplify this by passing 0 instead of `capability.identifier` in the call to `mspace_dispatch` above?

rmn30: Maybe also create a symbolic name such as `AllocatorOwnedID`?

davidchisnall: Done, both changes.

sdk/core/allocator/main.cc
```diff
@@ -411,6 +415,9 @@ namespace
 		  reinterpret_cast<PrivateAllocatorCapabilityState *>(capability);
 
 		// Assign an identifier if this is the first time that we've seen this.
+		// The value of 0 is used in allocator capabilities as a marker for
+		// uninitialised capabilities.  It is used internally in the heap for
+		// objects that are owned by the allocator.
```
nwf, line 420: Use `QuotaIdentifierAllocatorOwned` on the next line?  Or at least refer to the value by that name in the comment?

davidchisnall: Done.

PR #653, opened 2026-02-03
core/allocator: don't bound encoded claim refs

nwf:
The root of each chain of claim objects for a given heap object is an encoded position in the heap (concretely, a uint16_t) within the latter object's chunk header.  A tightly-bounded capability to this field will have its base within the chunk header, and since we mark those in the revocation bitmap, such a capability cannot survive being round-tripped through memory (either via preemption or compiler spill).


rmn30 (approved):
It took me several goes to understand the comment but I got there in the end. Not sure how I'd phrase it any more clearly!

PR #661, opened 2026-02-10
cheriot-audit in build

DjangoClouds:
Allows us to incorporate `cheriot-audit` into the xmake process. We've also modified `examples/10.audit` to demonstrate the usage. Essentially, we can now call `xmake audit` and it will go through the code and any submodules running and required audits.

#### Normal Use
The idea of this code is:
1. Extend `xmake` with a new `xmake audit` ability that runs `cheriot-audit` on the selected firmware.
2. You can set the audit queries and Rego files in the individual `xmake.lua` files for each library/compartment. For example: The `xmake.lua` file in the CHERIoT network stack could be modified to include code that can run the network audit code.
3. By default, `xmake audit` will run only the queries already coded in the `xmake.lua` files.
4. You can pass a single Rego statement or a table with multiple Rego statements.
5. If the Rego doesn't resolve to `true`, the build will fail reporting which Rego test failed.
6. You can place an audit on a specific compartment. If that compartment is called by multiple other compartments it will only run the test once for each firmware listed in the xmake file.
7. A good place to set the `cheriot.audit` value is in a `after_load` hook. This best done as a rule.
8. By applying tests to specific compartments, you can ensure that, if a compartment is reused elsewhere, it is being used correctly.
9. By default, `data.rtos.valid()` will always be executed (unless the query is overridden from the command line)
10. You can also pass in queries from the command line by using `xmake audit <<query>>`. In this case, it will still use all the Rego files in the `xmake.lua` files but will only run the query that you passed and will return the results for you to view.
11. You can also include additional Rego file(s) as a command line option, which will add the extra Rego file(s) to all the other defined in the `xmake.lua` files. It is called like this: `xmake audit -modules=<<file1|file2||file3>> <<query>>

### Examples Rules in xmake.lua
For all of these we're going to create a rule called "caesar.checks" which we can apply to a compartment (in this example, "producer") as follows:
```lua
compartment("producer")
    add_files("producer.cc")
    add_rules("caesar.checks", {private = true})
```

#### Example 1: A single, direct query
This will directly call "data.rtos.valid" and the build will fail if the result is not `true`.
```lua
rule("caesar.checks")
    before_build(function (target)
        local audit = {
            query='data.rtos.valid' -- A direct query
        }
        target:set("cheriot.audit", audit)
    end)
```

#### Example 2: A Query with a Rego File
This query uses the `valid` function in the `caesar.rego` file. Note: Here, we assume that the Rego file is in the same location as the xmake file.
```lua
rule("caesar.checks")
    before_build(function (target)
        local audit = {    -- A query with a rego file
            query='data.caesar.valid', 
            module=path.join(target:scriptdir(), "caesar.rego")
        }
        target:set("cheriot.audit", audit)
    end)
```

#### Example 3: Multiple Queries & Multiple Rego Files
This shows how to include multiple queries and multiple Rego files:
```lua
rule("caesar.checks")
    before_build(function (target)
        local audit = {  -- Multiple queries (mixed direct queries and rego file queries)
            query={'data.caesar.valid', 'data.compartment.mmio_allow_list("uart2", {"uart_man"})'},
            module={path.join(target:scriptdir(), "caesar.rego"), path.join(target:scriptdir(), "brutus.rego")}
        }
        target:set("cheriot.audit", audit)
    end)
```

#### Example 4: Command Line Variables
In this example we are adding two new module files and calling queries within them.
```sh
xmake audit --module="caesar/caesar.rego|brutus/brutus.rego" "data.caesar.valid" "data.brutus.valid"
```
or
```sh
xmake audit -m "caesar/caesar.rego|brutus/brutus.rego" "data.caesar.valid" "data.brutus.valid"
```
Notes:
1. The module files are separated with a `|` character. They are all encased in double quotes.
2. The queries are separated by a blank space. The queries are also placed in double brackets.

### Envisioned Use
We envision that individual modules would each have there own folder containing an `xmake.lua` file and, if needed, Rego file. Anytime this module is reused, the root `xmake.lua` file would include its folder and it will automatically pick up the module's `xmake.lua` file and automatically add any rules in that module to the build.

Consider the following directory structure based on the [cheriot-template](https://github.com/CHERIoT-Platform/cheriot-template) repo:
```
cheriot-template
|->src
|  |->main.cc
|  |->module1
|     |->xmake.lua
|     |->module1.rego
|->third_party
|  |->cheriot_rtos @ XXXXXXX
|  |->uart_man
|  |  |->xmake.lua
|  |  |->uart_man.rego
|  |->network-stack
|     |->xmake.lua
|     |->network-stack.rego
|->xmake.rego
```
Under this scheme, each additional module would have it's own Rego module and an `xmake.lua` file that tells us how to build the code and which audit query to run. This keeps the module and its required audits together in one place increasing the chance that audit will be correctly run.

### To Do
- [x] If the query is passed from the command line: collect all the rego modules together.
- [x] When adding a new module with "--module=" it is added to the existing modules.
- [ ] Test with the Network Stack
- [x] `data.rtos.valid` is tested by default.
- [x] Passing a query from the command line use all existing defined modules but ignores any existing queries in favour of the new ones.
- [ ] Maybe add a new example, showing how to include things in a modular way.


sdk/xmake.lua
```diff
@@ -1113,6 +1113,111 @@ rule("cheriot.firmware.linkcmd")
 		batchcmds:add_depfiles(target:get("cheriot.ldscripts"))
 		batchcmds:add_depfiles(objects)
 	end)
+    after_link(function (target)
```
nwf: Just to keep separation of concerns a little more clear, this _might_ want to be part of a different `rule()` than `"cheriot.firmware.linkcmd"`, perhaps `rule("cheriot.auditcmd")` or somesuch.

sdk/xmake.lua
```diff
@@ -1113,6 +1113,111 @@ rule("cheriot.firmware.linkcmd")
...
+                if regofile then
+                    raise("Audit failed! target: "..target:name()..", rego file: "..regofile..", query: "..query)
+                else
+                    raise("Audit failed! target: "..target:name()..", query: "..query)
+                end
+            end
+        end
+
+		-- local board_file = target:get("cheriot.board_file")
+        local board_file = path.join("$(buildir)", "board.json")
```
nwf: This file isn't always built.  You probably want to...
1. add, in this rule's `after_load()` script hook, `target:add("deps", "cheriot.board.file")`
2. use `target:dep("cheriot.board.file"):targetfile()` to get the path to the board file.

sdk/xmake.lua
```diff
@@ -1113,6 +1113,111 @@ rule("cheriot.firmware.linkcmd")
...
+		local compartment_report = path.join(target:scriptdir(), target:targetfile() .. ".json")
+
+        -- Work through the dependancy tree and call this only once for each dependancy
+        -- Note: When we PR this to cheriot:
+        --  1. We can use the existing visit_all_dependencies_of function
```
nwf: Please do, yes. :)

DjangoClouds:
Interesting. The failing build checks here could be because the data.rtos.valid is caesar.rego is causing the builds to fail. I imagine, if we commented it out, it would pass the tests?

sdk/xmake.lua
```diff
@@ -1113,6 +1113,111 @@ rule("cheriot.firmware.linkcmd")
 		batchcmds:add_depfiles(target:get("cheriot.ldscripts"))
 		batchcmds:add_depfiles(objects)
 	end)
+    after_link(function (target)
+        -- This function goes through each dependancy looking for an audit instruction and runs any audits found.
+        -- Assumptions:
+        -- 1. cheriot.board_file is set
```
nwf: That's (thankfully!) stale.

sdk/xmake.lua
```diff
@@ -1113,6 +1113,111 @@ rule("cheriot.firmware.linkcmd")
...
+        -- 2. Running post link means that the compartment report should be at target:targetfile() .. ".json"
```
nwf: If things in the build system are starting to depend on that, we might want to `target:set("cheriot.link_report", ...)` up in the `after_load` hook and use that in the `on_linkcmd` hook and here?

sdk/xmake.lua
```diff
@@ -1113,6 +1113,111 @@ rule("cheriot.firmware.linkcmd")
...
+        -- 3. Each target to be checked has a "cheriot.audit" set.
+        -- 4. "cheriot.target" can take a string, a single table or a table of tables
+        -- 5. If you pass a string it will be the query to run against the rego file. Since you build this in the compartment's xmake.lua file, you can make this as complex as you'd like.
+        -- 6. If the format is a table or table of tables, the format for each table is:
+        --    [1] The query to run against the rego file. Since you build this in the compartment's xmake.lua file, you can make this as complex as you'd like.
+        --    [2] (OPTIONAL) The path to the rego file.
```
nwf:
```suggestion
        -- 3. Each target to be checked has a "cheriot.audit" set to...
        --    * a string: the rego query to run. (Since you build this in the compartment's xmake.lua file, you can make this as complex as you'd like.)
        --    * an array of up to two strings:
        --      [1] The query to run, as above.
        --      [2] (OPTIONAL) The path of a rego file to load before the query.
        --    * an array of arrays of the above type; each query will be run sequentially, and all must pass
```

That said, I wonder if it's also worth supporting strings in that "array of array the above type" option (making it "an array of strings or arrays of the above type")?

sdk/xmake.lua
```diff
@@ -1113,6 +1113,111 @@ rule("cheriot.firmware.linkcmd")
...
+        -- 7. The code code looks for the final result of "true" follwed by a new line. Anything else is raised as an exception.
+        -- 8. cheriot-audit is in the PATH. If this is going to be an issue we'll have to add code to go looking for it (SIGH!).
```
nwf: You can use https://xmake.io/api/scripts/extension-modules/lib/detect.html#detect-find-tool to simplify, but probably not eradicate, the search, if needed.  I'm content with requiring it to be on $PATH at the moment, like we do for `clang` and friends already.  If this logic gets moved out to a `xmake` `task`, then making it an `option()` for that `task` might also be a solution.

sdk/xmake.lua
```diff
@@ -1113,6 +1113,111 @@ rule("cheriot.firmware.linkcmd")
...
+
+		-- local board_file = target:get("cheriot.board_file")
+        local board_file = path.join("$(buildir)", "board.json")
+		local compartment_report = path.join(target:scriptdir(), target:targetfile() .. ".json")
+
+        -- Work through the dependancy tree and call this only once for each dependancy
+        -- Note: When we PR this to cheriot:
+        --  1. We can use the existing visit_all_dependencies_of function
+        --  2. We can place this code in cheriot-rtos/sdk/xmake.lua
+        --  3. We should apply this rule in rule("cheriot.firmware")?
```
nwf: I wonder if this is better done as a separate `xmake` `task`, so that the flow would be `xmake config && xmake build && xmake audit`?  @davidchisnall?

nwf (changes requested):
Thank you!  I think this kind of integration would really help.  Some scattered notes, thoughts, and questions.

sdk/xmake.lua
davidchisnall: That would be nice, yes.

sdk/xmake.lua
DjangoClouds: Do you mean something like this?
```lua
local audit = {  -- Multiple queires (mixed direct queries and rego file queries)
            {'data.caesar.valid', path.join(target:scriptdir(), "caesar.rego")}, 
            'data.rtos.valid'
        }
        target:set("cheriot.audit", audit)
```
Where the second element is a string not a table (I think they are "tables" rather than "arrays" - this was my first experience of Lua and I need to update all my comments).
In which case, I _think_ it already supports this due to the hierarchical nature of the code but I can double check.

sdk/xmake.lua
DjangoClouds: Meaning that `cheriot.board_file` will always be set?

sdk/xmake.lua
DjangoClouds: I think people may not bother running it at all ("path of least resistance" and all that).

sdk/xmake.lua
DjangoClouds: It's not supported. I missed that. To be honest, I wasn't happy with relying on the table order as a basis for meaning but it is simple to understand. I think I've got a better idea though:

**A Better Idea**
We could introduce named members of the tables. If we have a `file` member that then becomes the rego file for all the remaining queries in that table. Tables can support mixing named an unnamed members so something like this would work:
```lua
local audit = {
  {  -- The file member becomes available for every other rego query in this sub-table.
    file=path.join(target:scriptdir(), "caesar.rego"),
    'data.caesar.valid',
    'data.caesar.valid2'
  },
  'data.compartment.mmio_allow_list("clint", {"scheduler"})'  -- This query is outside the above table and so has no file associated with it.
}
```
An added bonus, is that we could add additional named members in future to cover things that we haven't thought of yet (as Lehman's Second Law begins to take hold).

**A Possible Complexity**
Looking through the code, `cheriot-audit` can accept multiple module files.
1. That means that we should also be process `file` when it is an array/table of multiple Rego files.
2. We might also have to look an hierarchical file inclusion. Something like this: 
```lua
local audit = {
    {
        file=path.join(target:scriptdir(), "caesar.rego"),
        'data.caesar.valid',   -- Query has access to functions in caesar.rego
        'data.caesar.surprised', -- Query has access to functions in caesar.rego
        {
            file={ path.join(target:scriptdir(), "brutus.rego"), path.join(target:scriptdir(), "cassius.rego"),
            'data.brutus.valid', -- Query has access to functions in caesar.rego, brutus.rego and cassius.rego
            'data.cassius.valid' -- Query has access to functions in caesar.rego, brutus.rego and cassius.rego
        }
    },
    'data.compartment.mmio_allow_list("clint", {"scheduler"})'  -- This query has no access to any rego files.
}
```
It does get me a little worried that I'm beginning to overcomplicate this.
@nwf @davidchisnall Any thoughts?

davidchisnall: I prefer the second.  The shape of the table should respect the CLI for cheriot-audit, so something like:

```lua
{
   module = "some/path/to/network_stack.rego",
   query = "data.network_stack.valid(myNetworkDevice)"
}
```

The xmake convention for things with multiple options is that they should be singletons or arrays, so the following would also be valid:


```lua
{
   module = { "some/path/to/network_stack.rego", "some/other/path/to/something.rego" }
   query = { "data.network_stack.valid(myNetworkDevice)", "data.otherthing.valid" }
}
```

If we make this an xmake tool, it would be really nice if:

 - `xmake audit` ran the queries and reported success or failure in the exit code.
 - If there are multiple queries and only one failed, `xmake audit` printed the failing one for diagnostics.
 - You could provide alternative queries such as `xmake audit --query 'data.compartment.shared_object("allocator_hazard_pointers")' to run individual queries and see their output.

sdk/xmake.lua
davidchisnall: I think it will be annoying if it fails the build.  During development, I often do things that poke small holes in the security model to expose things for additional testing.  My debug builds will often fail the security checks for a release build for this reason, but even when I'm doing a build that the build system calls a release build, it's fine for the checks to fail for testing.  I want to use the audit checks in CI, for pre-merge checks, so that the version in tree is always respecting my security policy, not to ensure that every single build is doing so.

sdk/xmake.lua
DjangoClouds: That makes sense & cuts through the complexity nicely!
So `xmake audit --query` won't be providing a pass or fail so much as reporting the result for introspective queries.
We may be able to add a `--module` to overload the current module selection too, if that could be useful.

sdk/xmake.lua
DjangoClouds: I get it now. No problem. Makes sense.

sdk/xmake.lua
nwf: (Aside: Lua somewhat informally uses the term "array" to refer to a subset of Lua tables, specifically those whose keys are solely a possibly empty span of positive integers starting at 1.  See https://www.lua.org/pil/11.1.html .  The Lua specification is careful not to use the word in normative text, but some of its builtins sure _happen to be_ useful for tables of such shape.)

sdk/xmake.lua
DjangoClouds: That's now working. Just working on being able to override the audit queries from the command line.

PR #667, opened 2026-02-25
Fix thread_millisecond_wait and add thread_wait_until

rmn30:
1. [Fix a bug in thread_millisecond_wait.](https://github.com/CHERIoT-Platform/cheriot-rtos/commit/cabebd78b0526d4c3c3f7ea92252fb6a5171e95f)
`(end-current) > MS_PER_TICK` was comparing cycles against
milliseconds which makes no sense.  The result is that we would call
thread_sleep with a timeout of 0 when we need to wait for between
MS_PER_TICK (e.g. 10) and TIMERCYCLES_PER_TICK cycles.  This does no
real harm except waste time doing a compartment call (which could fail
if we're out of trusted stack etc.).  Fix this and also use the
existing TIMERCYCLES_PER_TICK macro rather than inventing a new name.

4. [Add thread_wait_until to wait or spin until given cycle timer value.](https://github.com/CHERIoT-Platform/cheriot-rtos/commit/c511ef88fd44d12aad68caeabee84d78b534720c) 
This is sometimes what you want if you're aiming for a particular time
(e.g. the start of the next frame start in an animation) rather than a
particular period.  I've also refactored thread_millisecond_wait in
terms of this.  We should probably make this a scheduler primitive but
this works as an interim measure.

### DIFF
--- a/sdk/include/thread.h
+++ b/sdk/include/thread.h
@@ -155,16 +155,15 @@ static inline uint64_t thread_millisecond_wait(uint32_t milliseconds)
 	return milliseconds;
 #else
 	static const uint32_t CyclesPerMillisecond = CPU_TIMER_HZ / 1'000;
-	static const uint32_t CyclesPerTick        = CPU_TIMER_HZ / TICK_RATE_HZ;
 	__if_cxx(
 	  static_assert(CyclesPerMillisecond > 0, "CPU_TIMER_HZ is too low"););
 	uint32_t cycles  = milliseconds * CyclesPerMillisecond;
 	uint64_t start   = rdcycle64();
 	uint64_t end     = start + cycles;
 	uint64_t current = start;
-	while ((end > current) && (end - current > MS_PER_TICK))
+	while ((end > current) && (end - current > TIMERCYCLES_PER_TICK))
 	{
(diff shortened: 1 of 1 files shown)


sdk/include/thread.h
```diff
@@ -138,6 +138,35 @@ static inline uint64_t thread_microsecond_spin(uint32_t microseconds)
...
+	uint64_t              start                = rdcycle64();
+	uint64_t              current              = start;
+	// sleep for periods > 1 tick
+	while ((end > current) && (end - current > TIMERCYCLES_PER_TICK))
+	{
+		Timeout t = {0, ((uint32_t)(end - current)) / TIMERCYCLES_PER_TICK};
+		(void)thread_sleep(&t, ThreadSleepNoEarlyWake);
+		current = rdcycle64();
+	}
+	// Spin for the remaining time.
```
nwf, line 162: Is it worth spinning like this or should we just round up the number of quanta in `thread_sleep`?

nwf (commented):
One nit, but I'd like to get this merged.

davidchisnall:
This should now be possible to do correctly using the new timeout model.

PR #669, opened 2026-02-27
Add a design document for sub quotas.

tolbooth:
This design works towards addressing #140. It is what was converged on after discussion with @hlef, @nwf, and @davidchisnall.

Putting this up as a PR so that feedback is easy to give, and so that we can set a plan in stone once it's in an appropriate state.

Some notes:

Since we only fold one level of the hierarchy each time, and we pass both parent and child to the fold API, we do not need, as originally discussed, to be able to differentiate a sub quota on the heap from an arbitrary sealed allocation. This is nice, because it means we don't need to free up any extra header bits for this to be viable.

Behavior when it comes to claims is dependent on how we resolve #654. Perhaps another motivational example as to why we might want to change the semantics: folding owner into claim and claim into owner are equivalent here. Is this what we want?

Regarding claims, since they are now marked as 'owned' (in their chunk header) by the allocator with ID=0, we need to treat them differently than the rest of the heap walk for `fold_quota()`. This may involve some delicate treatment--it would be nice to not walk the claims of all chunks on the heap, searching for those owned by the quota we're folding (`claim_find(quota_id, &chunk)`), but it's unclear if we can just walk chunks with ID=0 either, as they may be claimed-but-not-owned allocations, rather than claims themselves. This may motivate work towards #656? Curious as to if there's an easy approach I'm missing here.

Also--Is there interest in an implementation of this with the simplified concurrency story detailed in the document? 

Work can begin immediately on the implementation of these sub-quotas (subject to comments here), as the storage structure, APIs, and concurrency mechanism are mostly independent of these trickier points.

### DIFF
--- a/docs/SubQuotas.md
+++ b/docs/SubQuotas.md
@@ -1,0 +1,188 @@
+# Dynamically Allocated Sub Quotas
+
+[Sub quotas](https://github.com/CHERIoT-Platform/cheriot-rtos/issues/140) provide a mechanism for delegating only a portion of a heap quota from one compartment to another compartment.
+This is desirable, as currently, when a compartment delegates its quota, it does so in whole.
+This leaves the compartment open to resource exhaustion attacks from the delegatee, or integrity violations via dangerous Allocator APIs like `heap_free_all`.
+[Previously](https://github.com/CHERIoT-Platform/cheriot-rtos/pull/617), we implemented user-defined permissions on top of sealed capabilities to allow for the creation of less-privileged allocation capabilities, in an attempt to minimize the trust necessary to delegate, but this was not a complete solution.
+
+
+For one, barring the usage of an allocation capability for specific Allocator APIs is overly restrictive: You, as a delegator, may want to allow the delegatee the ability to perform quick cleanup on this quota via `heap_free_all`, but don't want to have to trust that the delegatee will not free your allocations from under you.
+There are also useful design patterns that sub quotas would enable, such as lending a portion of quota as collateral to a compartment that allocates objects for you.
+The delegatee can use your quota to claim allocations made with its quota, ensuring you won't exhaust its resources without also exhausting your own.
+You are also guaranteed the delegatee will not consume more of your resources than the portion explicitly shared.
+Supporting such a relationship of mutual distrust is at the core of the CHERIoT compartment model.
+
+
+Having the ability to dynamically allocate sub quotas solves these issues.
+The pattern becomes:
+
+
+```cpp
+DECLARE_AND_DEFINE_ALLOCATOR_CAPABILITY(parentQuota, 1024 * 1024);
+#define PARENT_QUOTA STATIC_SEALED_VALUE(parentQuota)
+
+// ...
+
+AllocatorCapability subQuota = split_quota(PARENT_QUOTA, 1024);
+
+CHERI::Capability sealedSocket = network_socket_create_and_bind(
+    timeout, subQuota, isIPv6, ConnectionTypeTCP);
+
+fold_quota(PARENT_QUOTA, subQuota); // allocation now owned by PARENT_QUOTA
+
+// ...
+
+network_socket_close(timeout, subQuota, sealedSocket);
+```
+
+The exact semantics are discussed later in the document, but the main contribution is in allowing the programmatic creation of minimally sized quotas that can be passed into quota-requiring APIs to isolate the callers main quota from a potentially compromised compartment.
+They establish an upper bound on the allocation done by the delegatee compartment, and prevent malicious usage of `heap_free_all` from compromising the rest of the delegator's allocations while still allowing its use within the delegatee, better respecting both the principle of intentionality and of least privilege.
+
+
+## Design
+
+At present, statically allocated quotas are assigned IDs lazily via a monotonically increasing counter, with a maximum of 2^13 = 8192 unique IDs.
+There is no provisioning for ID reuse.
+To that end, I propose running an allocator for quota IDs in the existing free space of `AllocatorCapabilityState` structs by threading them onto a sorted linked list.
+
+
+To do so, each quota will additionally maintain:
+- A doubly-linked list structure (prev/next pointers)
+- Initially assigned quota (provides a fast-path for destruction)       
+- (Optionally) A synchronization primitive to allow for protecting sub quotas during operations that drop the allocator lock.
+
+
+The list is maintained in sorted order by identifier.
+We use a next-fit allocation policy with a roaming `head` pointer to reduce clustering of IDs.
+
+
+Since the AllocatorCapabilityState is 24 bytes in size, and only 48 bits are currently used, we have 16 bits to store the initial quota value and 16 bytes to describe our storage datastructure.
+
+
+Using 16-bit compressed pointers for the linked list structure and a `futex` word for synchronization, we retain 8 bytes for future use while we incur no additional storage overhead to provide ID reuse.
+
+
+### Handling Static Quotas
+
+We know at build time exactly how many static quotas will ever exist.
+We reserve the first n IDs for these, allowing them to still be lazily assigned as the quotas are used.
+We store this counter in a static variable, as is done currently.
+
+
+We do not thread static quotas onto the linked list, since the allocator is expected not to retain references to them after usage with its APIs.
+Indeed, the allocator may only ever be passed a no-capture handle, and in such a case be fundamentally unable to track these quotas.
+
+
+### Creation of a sub quota
+
+Sub quotas are created via a call to `split_quota(parent, size)`, which splits off the requested size into a new child sub quota. This operation will fail if the required quota is not available, which includes the size of the sub quota object, as it will be allocated using the parents quota.
+It has no effect on existing allocations on the parent quota.
+
+
+Creation requires taking a new ID and updating the list of extant sub quotas.
+We find the interval of free IDs between `head` and `head->next`.
+If there is a free ID in the interval, we assign to our new quota the midpoint and set our new `head` to `head->next->next`.
+If are no free IDs in the interval, we advance head and try again.
+
+
+As long as the total number of active sub quotas stays low, allocation is a fast operation and should require chasing very few pointers.
+
+
+### Destruction of a sub quota
+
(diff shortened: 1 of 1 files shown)


docs/SubQuotas.md
```diff
@@ -0,0 +1,188 @@
...
+
+
+The list is maintained in sorted order by identifier.
+We use a next-fit allocation policy with a roaming `head` pointer to reduce clustering of IDs.
+
+
+Since the AllocatorCapabilityState is 24 bytes in size, and only 48 bits are currently used, we have 16 bits to store the initial quota value and 16 bytes to describe our storage datastructure.
+
+
+Using 16-bit compressed pointers for the linked list structure and a `futex` word for synchronization, we retain 8 bytes for future use while we incur no additional storage overhead to provide ID reuse.
```
davidchisnall, line 62: Note that compressed pointers are fine for dynamically allocated quotas (they are in the heap), but not for static ones.  Static quotas are allocated outside of the heap and the allocator does not hold a capability that allows it to rederive capabilities to them.

There is also no reason that static and dynamic quotas have the same layout.  They can be differentiated by the hardware sealing type used to seal them, so if we need more space in dynamic ones than static, that's fine.

docs/SubQuotas.md
```diff
@@ -0,0 +1,188 @@
...
+
+
+### Handling Static Quotas
+
+We know at build time exactly how many static quotas will ever exist.
```
davidchisnall, line 67: We know at *link* time how many static quotas will exist.  We do not know this when we build the allocator.

docs/SubQuotas.md
```diff
@@ -0,0 +1,188 @@
...
+We reserve the first n IDs for these, allowing them to still be lazily assigned as the quotas are used.
```
davidchisnall, line 68: We probably need to compile in a maximum number of static allocator capabilities, but that should be fine.

docs/SubQuotas.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Creation of a sub quota
+
+Sub quotas are created via a call to `split_quota(parent, size)`, which splits off the requested size into a new child sub quota. This operation will fail if the required quota is not available, which includes the size of the sub quota object, as it will be allocated using the parents quota.
+It has no effect on existing allocations on the parent quota.
+
+
+Creation requires taking a new ID and updating the list of extant sub quotas.
+We find the interval of free IDs between `head` and `head->next`.
+If there is a free ID in the interval, we assign to our new quota the midpoint and set our new `head` to `head->next->next`.
```
davidchisnall, line 84: Why the midpoint?  If we assign one below the second element and insert ourself as the second element, we reduce fragmentation.  I'm not sure what the tradeoffs are after a bunch of allocation and deallocation of these things is, it would be nice to have some modelling to see what the common-case traversal will need to be.

docs/SubQuotas.md
```diff
@@ -0,0 +1,188 @@
...
+
+As long as the total number of active sub quotas stays low, allocation is a fast operation and should require chasing very few pointers.
+
+
+### Destruction of a sub quota
+
+Because a sealed allocation can only be freed by the compartment that holds the sealing key, and a delegated sub quota may be used to make sealed allocations, we cannot support a notion of destruction for sub quotas that forcibly frees all associated allocations.
+
+
+Rather 'destroying' a sub quota, which would imply a freeing of then-orphaned allocations, we provide a `fold_quota(parent, child)` API.
```
davidchisnall, line 96:
```suggestion
Rather than 'destroying' a sub quota, which would imply a freeing of then-orphaned allocations, we provide a `fold_quota(parent, child)` API.
```

docs/SubQuotas.md
```diff
@@ -0,0 +1,188 @@
...
+The parent-child relationship can be verified by checking the ID of the heap chunk the child resides in.
+To transfer ownership, the allocator must walk the heap and modify the owner ID of each associated chunk.
+
+
+The parent assumes responsibility for freeing the allocations as it pleases.
+Any API that takes a delegated allocator capability and uses it for sealed allocations is already expected to expose an API to free said allocations.
+At which point delegator may pass a restricted copy of its parent capability that may only be used to free said allocations.
+
+
+Claims made on objects allocated by the child sub-quota will be unaffected by the fold, but claims made by the child sub-quota will be transferred as other heap allocations are.
```
davidchisnall, line 107: Minor not: I don't really like `fold` as a term here.  It implies a FP-like fold operation to me, which is not what it does.  Recombine is probably a better verb since it implies both combination and the fact that the combined things must have previously been combined and split (you can't, in the current design, pull out part of one quota and then combine it back into another.  And I think that's the right choice because otherwise it's easy to steal someone else's quota).

docs/SubQuotas.md
```diff
@@ -0,0 +1,188 @@
...
+
+
+## Concurrency
+
+### Simplified
+
+Allocator APIs may drop the lock midway through an operation, as in a call to allocate with a full heap.
+In light of this, the simplest case for handling concurrency for sub quota objects is to insert a validity check for allocator capabilities any time the lock is reaquired, at which point the operation should fail if the quota is no longer valid.
+This is likely sufficient for an initial implementation.
+Splitting quota will require holding the allocator lock.
```
davidchisnall, line 127: This should probably be tracked separately.  We have an issue for the fact that `heap_free_all` holds the lock for too long.  We probably need to fix that, then reuse the mechanism that `heap_free_all` uses for this.

docs/SubQuotas.md
```diff
@@ -0,0 +1,188 @@
...
+### Further Further Work
+
+As with `heap_free_all`, we'd like to be able to drop the allocator lock while executing a fold.
+Refcounting handles the quota object side of this story, but we still run into the issue of dropping the lock, yielding, and waking up with an invalid position, since the heap structure may have been mutated.
+To mitigate this issue, we require an extra bit in the header to mark a given chunk as an anchor for the heap walk.
+Before dropping the lock during the walk, we set the `anchor` bit, which signals that, should the chunk be 'freed', it shouldn't be marked revoked.
+Then, when our walk wakes back up, we handle the actual freeing of said chunk, and continue.
+
+
+To avoid starvation in a scenario where a low-priority thread starts a heap walk while another frees a very large object, we can, in the free path, split off a minimally sized chunk of the large object to serve as the anchor.
```
davidchisnall, line 187: We can also use a priority-inheriting futex for the lock and kick the sleeping thread at the relevant point if necessary.

davidchisnall (commented):
Sorry for the late review!

docs/SubQuotas.md
tolbooth: Do you see being able to differentiate them as being desirable?

The thought here re:compressed pointers is that the linked list won't track static quotas. This is due in part to the fact that we just can't use compressed heap pointers for objects that don't reside on the heap, but mostly that we can't guarantee the allocator will ever see a reference to a static quota with GL permission, right? A compartment may always strip this before making calls to heap APIs, so even storing full capabilities, we wouldn't be able to guarantee the completeness of the linked list, as far as I understand it.

docs/SubQuotas.md
tolbooth: I like 'recombine'. With that change, should 'create' stay as 'create' or does 'split' work better as a pairing?

davidchisnall: Split sounds nicer.

docs/SubQuotas.md
tolbooth: The thought with taking the midpoint is that, since these are always size 1 allocations, we actually want the space to be maximally fragmented to avoid the case of allocating a large chunk of contiguous IDs then having to walk over all of them to find the next free ID. 

I agree though that it's not entirely clear this is a realistic workload: `allocate allocate .... allocate free allocate` ... given the use case of these objects is to, e.g., create a socket, it's perhaps more likely that allocates and frees will be interleaved quite evenly. So, worth some time modeling. 

For now I'll implement both schemes, as they're quite easy to plug-in. Final design can have some data to back this up.

docs/SubQuotas.md
```diff
@@ -0,0 +1,188 @@
...
+```
+
+
+### Further Further Work
+
+As with `heap_free_all`, we'd like to be able to drop the allocator lock while executing a fold.
+Refcounting handles the quota object side of this story, but we still run into the issue of dropping the lock, yielding, and waking up with an invalid position, since the heap structure may have been mutated.
+To mitigate this issue, we require an extra bit in the header to mark a given chunk as an anchor for the heap walk.
+Before dropping the lock during the walk, we set the `anchor` bit, which signals that, should the chunk be 'freed', it shouldn't be marked revoked.
+Then, when our walk wakes back up, we handle the actual freeing of said chunk, and continue.
```
nwf, line 184: The use of a single bit would somewhat complicate heap walk by precluding two threads from anchoring on the same object, wouldn't it?

davidchisnall: I think it’s fine for two heap-walking operations to be serialised with respect to each other. These are infrequent and slow.

We could probably allow early return for both by tracking the amount of memory allocated against a quota and stopping when we’ve found all objects.

It’s a shame we don’t have header space for a double linked ring of objects allocated against a particular quota, because the only realms were actually doing the heap walk is to fins objects allocated with a specific quota. If applying violence to the headers frees up enough bits for this then we should consider that instead.

nwf: Sadly, that's a lot of bits... probably at that point we'd be talking about growing the header by another two 32-bit words (can't really do just one for alignment, right?), and I think that's a bridge too far... even a footer 32-bit word on allocated chunks is probably too big.

tolbooth: I agree w.r.t tracking 'current' memory allocated against a quota--having a fast path for when all allocations made with the quota have already been freed / cutting the walk short seems reasonable as a solution. That gives us a spectrum of use from infrequent and care-free: hand off the quota, what happens with it is opaque, then fold back in whenever, vs. a hot path where you coordinate both sides of this handoff to make sure all allocations are freed before the fold, letting you skip the heap walk entirely.

There's enough room in the existing padding to add this to each `AllocatorCapabilityState` even with a move to using 32 bits for each heap offset.

Issue #332, opened 2024-11-01
Allocator: apply violence to free up some bits in the chunk header


nwf:
The current allocator is a `dlmalloc`-like design with a header fronting every allocated object or contiguous span of unallocated space.  Specifically, as defined at https://github.com/CHERIoT-Platform/cheriot-rtos/blob/299de3a7942b00641eb6517d9186600098148c73/sdk/core/allocator/alloc.h#L289-L320 .  Of note, there are 16 bits in every header dedicated to finding the previous header.  However, the only operation that traverses the header linked list backwards are the...
- [`ok_in_use_chunk`](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/299de3a7942b00641eb6517d9186600098148c73/sdk/core/allocator/alloc.h#L1582) integrity check function, having already checked that the prior chunk was not in use, and
- [`mspace_free_internal`](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/299de3a7942b00641eb6517d9186600098148c73/sdk/core/allocator/alloc.h#L2395-L2398) when doing coalescing, gated on, again, the prior chunk not being in use.

There's a wonderful thing about chunks not in use: their bodies are free for our use.  That's where we put the [`MChunk`](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/299de3a7942b00641eb6517d9186600098148c73/sdk/core/allocator/alloc.h#L592-L672) and [`TChunk`](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/299de3a7942b00641eb6517d9186600098148c73/sdk/core/allocator/alloc.h#L711-L849) free object metadata.  We could, for the small price of increasing our minimum allocation size, including chunk header, from [16 bytes (two 32-bit addresses, `sizeof(MChunk)`)](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/299de3a7942b00641eb6517d9186600098148c73/sdk/core/allocator/alloc.h#L684) to 24 (since headers must be 8-byte aligned), put more stuff therein.  Notably, we can steal a trick from `malloc()`s past and stuff a pointer (or address or even relative offset, as desired), at the _foot_ of a free span, pointing back to its _header_.  `!MChunkHeader::isPrevInUse` then indicates that this footer is present and valid.

[...]

PR #672, opened 2026-02-27
Allocator violence #332

nwf:
This series slightly tidies allocation headers and then enacts the violence described in #332 to get us a bunch more free bits and larger chunk sizes.  By the end of the series, we support 18-bit single chunks, or 2 MiB heaps (since, at the moment, the entire heap must fit in one chunk if everything's free; it might be nice to -- separately! -- fix that).

Running the test suite in release configuration on the sail, this has a very small (~0.4%) performance decrease in the allocator, adding ~26K cycles to the test suite's ~6.98M.  I have not attempted to investigate the generated code (as yet), but can if there's desire.

### DIFF
--- a/sdk/core/allocator/alloc.h
+++ b/sdk/core/allocator/alloc.h
@@ -49,8 +49,6 @@ constexpr size_t MaxSmallSize = 1U << TreeBinShift;
 using SmallSize               = uint16_t;
 constexpr size_t MaxChunkSize = (1U << utils::bytes2bits(sizeof(SmallSize)))
                                 << MallocAlignShift;
-// the compressed pointer. Used to point to prev and next in free lists.
-using SmallPtr = size_t;
 // the index to one of the bins
 using BIndex = uint32_t;
 // the bit map of all the bins. 1 for in-use and 0 for empty.
@@ -750,14 +748,13 @@ class __packed __aligned(MallocAlignment) MChunk
 		                                  offsetof(MChunk, ring));
 	}
 
-	// the internal small pointer representation of this chunk
-	SmallPtr ptr()
+	ptraddr_t addr()
 	{
 		return CHERI::Capability{this}.address();
 	}
 
 	/**
-	 * The friend needs to access bk, fd and ptr() to rederive chunks.
+	 * The friend needs to access bk, fd and addr() to rederive chunks.
 	 * XXX: How do we grant access to only these?
 	 */
 	friend class MState;
@@ -1082,7 +1079,7 @@ class MState
 	/// Rederive a capability to a `T` from the heap range.  This does not
 	/// apply bounds to the result.
 	template<typename T>
-	T *rederive(SmallPtr ptr)
+	T *rederive(ptraddr_t ptr)
 	{
 		CHERI::Capability<T> cap{heapStart.cast<T>()};
 		cap.address() = ptr;
@@ -2090,7 +2087,7 @@ class MState
 		              small_index2size(i));
 
 		if (RTCHECK(&p->ring == bin->last() ||
-		            (ok_address(f->ptr()) && f->bk_equals(p))))
+		            (ok_address(f->addr()) && f->bk_equals(p))))
 		{
 			if (br == fr)
 			{
@@ -2099,7 +2096,7 @@ class MState
 				bin->reset();
 			}
 			else if (RTCHECK(&p->ring == smallbin_at(i)->first() ||
-			                 (ok_address(b->ptr()) && b->fd_equals(p))))
+			                 (ok_address(b->addr()) && b->fd_equals(p))))
 			{
 				ds::linked_list::unsafe_remove(&p->ring);
 			}
@@ -2129,7 +2126,7 @@ class MState
 		MChunk *p = MChunk::from_ring(b->unsafe_take_first());
 
 		Debug::Assert(
-		  ok_address(p->ptr()), "Removed chunk {} has bad address", p);
+		  ok_address(p->addr()), "Removed chunk {} has bad address", p);
 
 		if (b->is_empty())
 		{
@@ -2194,8 +2191,8 @@ class MState
 				{
 					TChunk *back =
 					  TChunk::from_ring(t->mchunk.ring.cell_prev());
-					if (RTCHECK(ok_address(t->mchunk.ptr()) &&
-					            ok_address(back->mchunk.ptr())))
+					if (RTCHECK(ok_address(t->mchunk.addr()) &&
+					            ok_address(back->mchunk.addr())))
 					{
 						t->ring_emplace(i, xHeader);
 						break;
@@ -2233,7 +2230,7 @@ class MState
 		{
 			TChunk *f = TChunk::from_ring(x->mchunk.ring.cell_next());
 			r         = TChunk::from_ring(x->mchunk.ring.cell_prev());
-			if (RTCHECK(ok_address(f->mchunk.ptr()) &&
+			if (RTCHECK(ok_address(f->mchunk.addr()) &&
 			            f->mchunk.bk_equals(&x->mchunk) &&
 			            r->mchunk.fd_equals(&x->mchunk)))
 			{
@@ -2277,7 +2274,7 @@ class MState
 					treemap_clear(x->index);
 				}
 			}
-			else if (RTCHECK(ok_address(xp->mchunk.ptr())))
+			else if (RTCHECK(ok_address(xp->mchunk.addr())))
 			{
 				if (xp->child[0] == x)
 				{
@@ -2294,13 +2291,13 @@ class MState
 			}
 			if (r != nullptr)
 			{
-				if (RTCHECK(ok_address(r->mchunk.ptr())))
+				if (RTCHECK(ok_address(r->mchunk.addr())))
 				{
 					TChunk *c0, *c1;
 					r->parent = xp;
 					if ((c0 = x->child[0]) != nullptr)
 					{
-						if (RTCHECK(ok_address(c0->mchunk.ptr())))
+						if (RTCHECK(ok_address(c0->mchunk.addr())))
 						{
 							r->child[0] = c0;
 							c0->parent  = r;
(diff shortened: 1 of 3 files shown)


sdk/core/allocator/alloc.h
```diff
@@ -272,65 +182,257 @@ namespace displacement_proxy
...
-		return displacement_proxy::
-		  Proxy<MChunkHeader, SmallSize, false, head2size, size2head>(this,
-		                                                              prevSize);
+		return PredecessorProxy(*this);
 	}
 
+	/**
+	 * A pointer proxy to encode the address of this chunk's predecessor.
+	 *
+	 * MChunkHeader predecessor pointers are... odd.  They exist only when the
```
nwf: Whoops, copypasta error.

sdk/core/allocator/main.cc
```diff
@@ -133,6 +133,15 @@ namespace
 	{
 		if (gm == nullptr)
 		{
+			/*
+			 * We'd love this to be static, but getting a constant to C++ and a
+			 * linker script sounds like a headache.  Settle for it being a
+			 * runtime assertion in debug builds.
+			 */
+			Debug::Assert(LA_ABS(__export_mem_heap_maximum_size) ==
```
davidchisnall: Maybe make this a `Debug::Invariant`?  It's checked only once, so the extra cost is low.

sdk/core/allocator/alloc.h
```diff
@@ -45,12 +45,16 @@ constexpr size_t TreeBinShift = MallocAlignShift + NSmallBinsShift;
...
-using SmallSize               = uint16_t;
-constexpr size_t MaxChunkSize = (1U << utils::bytes2bits(sizeof(SmallSize)))
-                                << MallocAlignShift;
-// the compressed pointer. Used to point to prev and next in free lists.
-using SmallPtr = size_t;
+/**
+ * The type of compressed sizes used in MChunkHeader-s (q.v.).
+ * See size2head and head2size for codecs.
+ */
+using CompressedSizeType            = uint32_t;
```
davidchisnall, line 53: The name is slightly confusing here, since this is the type that is used for the *bitfield* type, i.e. the register size that the real type can be sign / zero extended to be.  The comment probably needs to say something about this, specifically that the range of this type does not imply the range of the data that it can represent when stored in headers.

sdk/core/allocator/alloc.h
```diff
@@ -272,65 +182,275 @@ namespace displacement_proxy
...
+	 * Compressed size of this chunk.  See cell_next().
+	 */
+	CompressedSizeType currSize : CompressedSizeBits;
+
+	/**
+	 * The second 32-bit word of a MChunkHeader holds the current owner (if any)
+	 * and, if the chunk is in use, the head of a linked list of claims on it.
+	 * This field is the unused space in that word.
+	 */
+	uint16_t unused2 : (16 - OwnerIDWidth);
```
davidchisnall, line 260: Is this where we will be able to stash the 'being-traversed' flag, for when things like `heap_free_all` drop the lock during traversal and take responsibility for actually freeing the current object if it's freed before they reacquire the lock?

sdk/core/allocator/alloc.h
```diff
@@ -272,65 +182,275 @@ namespace displacement_proxy
...
+	 * The unique identifier of the allocator.  The ID 0 is reserved for
+	 * objects that are owned by the allocator, such as claims.
+	 */
+	uint16_t ownerID : OwnerIDWidth;
+
 	/// Head of a linked list of claims on this allocation
 	uint16_t claims;
 
+	/**
+	 * This must be a result not in the range of upper layer's encoding
```
davidchisnall: Do we have a static assert of this somewhere?

sdk/core/allocator/alloc.h
```diff
@@ -272,65 +182,275 @@ namespace displacement_proxy
...
+		public:
+		using Type = MChunkHeader;
+
+		__always_inline PredecessorProxy(MChunkHeader &h) : hdr(h) {}
+
+		__always_inline operator MChunkHeader *() const
+		{
+			switch (hdr.prevState)
+			{
+				case PrevState::Free:
```
davidchisnall, line 331: This algorithm could use some comments (which could just be references to where to find the data structure comment elsewhere).

sdk/core/allocator/alloc.h
nwf: Static assert is somewhat tricky, but I've added a `Debug::Assert` in `Claims::encode_address` and a cross-reference to its comment to help future-us not get this wrong.

sdk/core/allocator/alloc.h
nwf: Yes, here or `unused1` (which would keep it next to the other flag bits).

sdk/core/allocator/alloc.h
```diff
@@ -45,12 +45,17 @@ constexpr size_t TreeBinShift = MallocAlignShift + NSmallBinsShift;
...
 
-// the compressed size. The actual size is SmallSize * MallocAlignment.
-using SmallSize               = uint16_t;
-constexpr size_t MaxChunkSize = (1U << utils::bytes2bits(sizeof(SmallSize)))
-                                << MallocAlignShift;
-// the compressed pointer. Used to point to prev and next in free lists.
-using SmallPtr = size_t;
+/**
+ * The type of the bitfield member holding compressed sizes in
+ * MChunkHeader-s (q.v.) and in MChunkHeader::PrevFooter-s.
```
davidchisnall, line 50:
```suggestion
 * `MChunkHeader`s (q.v.) and in `MChunkHeader::PrevFooter`s
```

Not really happy with 'q.v.' though.  I've not seen it used in anything written in the last hundred years and we should make comments easy to read.

sdk/core/allocator/alloc.h
```diff
@@ -272,65 +183,291 @@ namespace displacement_proxy
  *       - body() is a MChunk (and not a TChunk)
  *       - Collected in a quarantine ring using body()'s MChunk::ring linkages
  *
+ *   - Free for allocation and tiny
```
davidchisnall, line 186: I think this is the first mention of 'tiny', so we should clarify what it means.  Perhaps we should have a small explanation of what 'tiny' and 'small' mean in terms of how they're encoded above this list.

sdk/core/allocator/alloc.h
```diff
@@ -443,12 +581,11 @@ __cheri_no_subobject_bounds MChunkHeader
 
 		auto newnext = new (newloc) MChunkHeader();
 		newnext->clear();
+		newnext->isCurrInUse = isCurrInUse;
```
davidchisnall, line 584: In the old version, we set `isPrevInUse`.  In the new version, this is a more complex multi-state thing, but I'm not sure I understand why setting it to 0 unconditionally in `->clear()` is always correct.  Perhaps a comment above `->clear()` could help?

I also wonder if we shouldn't remove `clear()` here and have the constructor of `MChunkHeader` initialise each field.  If we provide default values for each field, the synthesised constructor will do this, but I'd also be tempted to add an explicit constructor that sets the in-use state and fold the above three lines into `auto newnext = new (newloc) MChunkHeader(isCurrInUse);` (which really should be `isCurrentInUse`, but sadly some of this code predates sensible naming - we should fix that in a later PR).

sdk/core/allocator/alloc.h
davidchisnall, line 567: I think this bit of the comment is now stale.

It's also presumably unsafe to split things that are in the allocated state and not in quarantine.  Doing so would allow user code access to heap state, so we should probably have an assertion to check that we never do that.

sdk/core/allocator/alloc.h
```diff
@@ -495,10 +632,7 @@ __cheri_no_subobject_bounds MChunkHeader
 };
 static_assert(sizeof(MChunkHeader) == 8);
 static_assert(std::is_standard_layout_v<MChunkHeader>);
-static_assert(
-  offsetof(MChunkHeader, claims) == 2 * sizeof(SmallSize) + sizeof(uint16_t),
-  "Metadata is no longer 16 bits.  Update the OwnerIDWidth constant to correct "
-  "the space used for the owner ID to match the remaining space.");
+static_assert(offsetof(MChunkHeader, claims) == 6);
```
davidchisnall, line 635: Why does it need to be 6?

davidchisnall (approved):
Approving, because I think all of the changes I'd like are cosmetic and you can hit the merge button when they're done.

sdk/core/allocator/alloc.h
nwf: Alas, we presently `split` in the allocated state when we are returning [prefixes](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/be9d1e7c3c5a711a10b5f225da19c4be61651cb0/sdk/core/allocator/alloc.h#L2785) and [suffixes](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/be9d1e7c3c5a711a10b5f225da19c4be61651cb0/sdk/core/allocator/alloc.h#L2807) of aligned allocation requests.

sdk/core/allocator/alloc.h
nwf: The `newnext->prevState` field gets implicitly set within `ds::linked_list::emplace()` by virtue of the write to the `PredecessorProxy`.  I'll update the commentary.

sdk/core/allocator/alloc.h
davidchisnall: Do we then mark them as deallocated after the split?  Can we improve this API slightly to make sure that at least *one* of the prefix or suffix is not allocated?

davidchisnall: (Not necessarily now, but track this as something we should clean up later with an issue)

sdk/core/allocator/alloc.h
nwf: Doesn't.  That probably qualifies as "scaffolding code from development that got inadvertently committed".

nwf:
Rejecting my own code since this fails to update the encoding of `MChunkHeader::claims` and so will break in practice, even if the test suite passes right now.  Fixing that is going to be moderately annoying since `main.cc`'s implementation of `Claim`-s assumes that `MChunkHeader::claims` is a `uint16_t` to which it can take a reference.

PR #686, opened 2026-04-08
Add initialiser framework.

davidchisnall:
This is the MVP version, and allows compartments to expose entry points that are run in an order based on priority.

This includes some refactoring of the boot.S code to ensure that we've zeroed the heap and dropped privileges before the initialisers are called.

It also calls the scheduler initialisation routine as a cross-compartment call.  This gives us some additional defence in depth against accidental leakage from the loader into the scheduler.

### DIFF
--- a/sdk/core/loader/boot.S
+++ b/sdk/core/loader/boot.S
@@ -6,6 +6,19 @@
 #include "../switcher/trusted-stack-assembly.h"
 #include <cheri-builtins.h>
 
+
+/**
+ * Machine-mode stack high water mark CSR
+ */
+#define CSR_MSHWM  0xbc1
+
+/**
+ * Machine mode stack high water mark stack base CSR
+ */
+#define CSR_MSHWMB 0xbc2
+
+
+
 .include "assembly-helpers.s"
 
     .section .loader_start, "ax", @progbits
@@ -72,8 +85,7 @@
 	 *
 	 *  - a5/x15: holds the RW cap root without GL(obal) permission.
 	 *
-	 * In particular, ra/x1, a0/x10, and a1/x11 are free and may be clobbered,
-	 * as platform code is quite likely to want to use .Lfill_block, below.
+	 * In particular, ra/x1, a0/x10, and a1/x11 are free and may be clobbered.
 	 */
 #if __has_include(<platform-early_boot.inc>)
 #	include <platform-early_boot.inc>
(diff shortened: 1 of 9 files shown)


sdk/core/loader/boot.S
```diff
@@ -72,8 +85,7 @@
 	 *
 	 *  - a5/x15: holds the RW cap root without GL(obal) permission.
 	 *
-	 * In particular, ra/x1, a0/x10, and a1/x11 are free and may be clobbered,
-	 * as platform code is quite likely to want to use .Lfill_block, below.
+	 * In particular, ra/x1, a0/x10, and a1/x11 are free and may be clobbered.
```
nwf, line 88: Hm... I thought we had an in-tree consumer that did use `.Lfill_block`.... oh, indeed, and it's been replaced.  I won't quibble over a few instructions that stick around, I suppose. :)

sdk/core/loader/boot.S
```diff
@@ -131,40 +143,62 @@
 
 	// ca4 still has the RW memory root; nothing to change
 
+	// Reserve 16 bytes at the top of the stack to later hold the hazard pointer slot
+	cincoffset		csp, csp, -16
```
nwf, line 147: Shuffle some comments and use a symbolic name:
```suggestion
    /*
     * Reserve dummy heap hazard pointer space.
	 * The switcher wants to zero these on call and return,
	 * so we'll give it something to zero, to avoid a branch.
	 * See below for where this is installed into the current
	 * thread's trusted stack info.
	 */
	cincoffset		csp, csp, -ThreadHeapHazards_size
```
with
```
EXPORT_ASSEMBLY_EXPRESSION(ThreadHeapHazards_size, 2*sizeof(void*), 16);
```
somewhere appropriate (`sdk/core/switcher/misc-assembly.h`?)

sdk/core/loader/boot.S
```diff
@@ -131,40 +143,62 @@
...
+
 	/*
 	 * The first argument (ca0) is a SchedEntryInfo out parameter.  Make space
-	 * for it at the base of the stack, where we can easily find it again after
+	 * for it at the top of the stack, where we can easily find it again after
```
nwf: Well, hm.  It's now neither the top nor the base, but just the current stack pointer.

sdk/core/loader/boot.S
```diff
@@ -183,57 +217,52 @@
...
+	// Loop exit 
+	bge				s1, t1, .Ldone_initialisers
+
+	// Loop body
+	// Load the sealed capability for the entry point
+	clc				ct1, 0(cs1)
+	// Call the switcher
+	cjalr			cs0
+	// Increment the array pointer
+	cincoffset		cs1, cs1, 8
```
nwf, line 241: I think I'd prefer we viewed the initializer array as containing structures and not raw pointers in case we want to add metadata in the future; that is, can the `0(cs1)` up a few lines and the `8` here have symbolic constants?

sdk/core/loader/boot.S
```diff
@@ -183,57 +217,52 @@
...
+.Ldone_initialisers:
 
 	/*
-	 * The machine is now about to approximate the behavor of the idle thread
-	 * (mtdc, sp) making a compartment call into the scheduler compartment (ra
-	 * and gp) and in particular its initialization hook, scheduler_entry.
-	 * This is not a proper cross-call, especially in that neither the call nor
-	 * return will go via the switcher's cross-call path (or use the trusted
-	 * stack), because we are not a fully-fledged RTOS thread.
+	 * Load the two return values (pcc and cgp for the scheduler entry point).
```
nwf: Stale comment; we're loading the actual sealed entry point handle for the switcher's function (since we are now going via the switcher) and the saved pointer to the thread info structure.

sdk/core/loader/boot.S
```diff
@@ -183,57 +217,52 @@
...
 	 *
-	 * Clear all other registers before invoking.  This ensures that the
-	 * scheduler does not, for example, get access to capability roots.
+	 * Drop these elements from the on-stack SchedulerEntryInfo, leaving only
+	 * its array of ThreadInfo-s (pointed to by sp).
```
nwf: Stale comment, I think?

sdk/core/loader/boot.S
```diff
@@ -131,40 +143,62 @@
...
 	clhu			s1, 0(cs1)
 	li			t0, -BOOT_THREADINFO_SZ
 	mul			s1, s1, t0
-	addi			s1, s1, -SchedulerEntryInfo_offset_threads
 	cincoffset		csp, csp, s1
-	neg			s1, s1
-	csetbounds		ca0, csp, s1
+	// ca1 now contains the bounded capability to the threads array
+	neg				s1, s1
+	csetboundsexact	ct0, csp, s1
```
nwf, line 163: It might be nice if a debug build of the loader printed this value prior to trying to write to it, so we see the potential `v:0`?  Perhaps as part of `Debug::log("Creating boot threads\n");`?

sdk/core/loader/boot.S
```diff
@@ -131,40 +143,62 @@
...
+	// Make space for the scheduler info
+	cincoffset		csp, csp, -SchedulerEntryInfo_size
+	csc				ct0, SchedulerEntryInfo_offset_threads(csp)
+	// Make space for the thread infos
```
nwf: Stale comment.

sdk/core/loader/boot.S
davidchisnall: It's fewer instructions in the case where you aren't using it elsewhere, and only one more in the case where it is used.  The function was four instructions, including cret, and needed a cjalr to get to it, so two of the five instructions were overhead.

sdk/core/loader/boot.S
davidchisnall: We previously hard coded this in the C++.  I've now pulled it out and added an invariant check to make sure that the size of the thread space is correct.

sdk/core/loader/boot.S
davidchisnall: Fixed.

sdk/core/loader/boot.S
davidchisnall: Moved the log up.

sdk/core/loader/boot.S
davidchisnall: I think that adds complexity now for a feature that we don't know that we want, and which will require significant changes in this code anyway if we do want it.

sdk/include/cheri.hh
```diff
@@ -940,6 +940,24 @@ namespace CHERI
 			return *this;
 		}
 
+		/**
+		 * Postfix increment.  As pointer operation, adds the size of the
+		 * pointee.
+		 */
+		Capability operator++(int)
+		{
+			return {ptr + sizeof(T)};
```
nwf: Isn't this double-counting?  `ptr` is a `T *`, so `+ sizeof(T)` is going to add `sizeof(T)*sizeof(T)`, isn't it?

davidchisnall: I believe you are correct.  But these are both dead code in the current version, not actually used, so I will delete them and do a separate PR for them with a test.

nwf (approved):

ia0:
Maybe I'm doing something wrong, but this commit introduces the following confusing issue:

```
error: cheriot-rtos/sdk/core/loader/boot.S:167:23: error: immediate must be an integer in the range [0, 4095]
 csetbounds ca0, csp, SchedulerEntryInfo_size
                      ^
cheriot-rtos/sdk/core/loader/boot.S:199:23: error: immediate must be an integer in the range [0, 4095]
 csetbounds ca2, ca2, 16
                      ^
  > in cheriot-rtos/sdk/core/loader/boot.S
```

AFAICT 16 in between 0 and 4095 (and I would not be surprised if `SchedulerEntryInfo_size` too).

Any idea what's wrong?

This can be reproduced with the `cheri/examples/hello_world` of `cheri-rust`.

nwf:
Alas, we are having trouble reproducing that internally.  One of us has encountered it but the rest of us are quite puzzled.  Will advise as soon as we know something.  In the interim, are you using our docker devcontainer?  What's your LLVM version?  Host platform?

ia0:
I'm not using the docker devcontainer.

```shell
cheri-rust/cheri/examples/hello_world % ../../../build/host/llvm/bin/llvm-config --version
22.1.1-rust-dev
```

Host platform is x86_64 (Debian based).

PR #688, opened 2026-04-10
Add documentation for Ephemeral Claims.

tolbooth:
What it says on the tin.

### DIFF
--- a/docs/EphemeralClaims.md
+++ b/docs/EphemeralClaims.md
@@ -1,0 +1,63 @@
+Ephemeral Claims
+================
+
+Ephemeral claims are a fast-path claim mechanism in the CHERIoT RTOS.
+While claims are persistent, they are also expensive, since their creation entails a cross compartment call to the allocator.
+Ephemeral claims are temporary, but do not require a cross-compartment call and so avoid the associated overheads.
+The primary use case of ephemeral claims is to hold an object alive early in a compartment's entry path, keeping it valid while the compartment processes it, possibly including taking a persistent claim of its own.
+
+Ephemeral claims behave much like regular CHERIoT claims, with a few key differences:
+- They belong to a thread rather than a compartment.
+- They aren't bound to any quota object.
+- Each thread may possess up to two ephemeral claims at any one time, and they must be on different objects.
+- Ephemeral claim expire whenever the thread performs a compartment call or makes another ephemeral claim.
+
+This introduces subtle race conditions. For example, the following sequence is racy:
+
+```
+Thread 1: ephemeral_claim(X)
+Thread 1: claim(X)
+ (!) Thread 1 is interrupted in the allocator before grabbing the lock
+Thread 2: free(X)
+```
+
+In the above sequence, Thread 2 succeeds in its attempt to free `X`, since Thread 1 released its ephemeral claim on the object while making the cross-compartment call to the allocator to acquire a persistent claim.
+
+Hazard Pointers
+---------------
+
+Under the hood, a thread makes an ephemeral claim by setting its hazard slots.
+These are two `void *` hazard pointers stored in the thread's trusted stack, a per-thread data structure that resides in the switcher.
+A thread can retrieve a write-only capability to its hazard slots through a call to the switcher API `switcher_thread_hazard_slots()`.
+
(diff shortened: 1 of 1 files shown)


docs/EphemeralClaims.md
```diff
@@ -0,0 +1,63 @@
...
+- They belong to a thread rather than a compartment.
+- They aren't bound to any quota object.
+- Each thread may possess up to two ephemeral claims at any one time, and they must be on different objects.
+- Ephemeral claim expire whenever the thread performs a compartment call or makes another ephemeral claim.
+
+This introduces subtle race conditions. For example, the following sequence is racy:
+
+```
+Thread 1: ephemeral_claim(X)
+Thread 1: claim(X)
```
davidchisnall, line 19: This is the problem.  The claim is dropped as soon as the allocator is called.  We have an analyser warning for this.  This is identical to simply not doing the ephemeral claim, from a concurrency perspective.

docs/EphemeralClaims.md
```diff
@@ -0,0 +1,63 @@
...
+Thread 2: free(X)
+```
+
+In the above sequence, Thread 2 succeeds in its attempt to free `X`, since Thread 1 released its ephemeral claim on the object while making the cross-compartment call to the allocator to acquire a persistent claim.
+
+Hazard Pointers
+---------------
+
+Under the hood, a thread makes an ephemeral claim by setting its hazard slots.
+These are two `void *` hazard pointers stored in the thread's trusted stack, a per-thread data structure that resides in the switcher.
```
davidchisnall, line 30: Technically, all of them are stored contiguously, each thread has a pointer to its two slots that the switcher can retrieve.

docs/EphemeralClaims.md
tolbooth: Reworded here to be more explicit about that. My understanding though was that this is intentional, and that we just want to be very clear about the existence of this behaviour. Is it something you're looking to change?

docs/EphemeralClaims.md
```diff
@@ -0,0 +1,66 @@
+Ephemeral Claims
+================
+
+Ephemeral claims are a fast-path claim mechanism in the CHERIoT RTOS.
+While claims are persistent, they are also expensive, since their creation entails a cross compartment call to the allocator.
+Ephemeral claims are temporary, but do not require a cross-compartment call and so avoid the associated overheads.
+The primary use case of ephemeral claims is to hold an object alive early in a compartment's entry path, keeping it valid while the compartment processes it, possibly including taking a persistent claim of its own.
```
davidchisnall, line 7: That's the main use case for normal claims.  Ephemeral claims are dropped on cross-compartment calls and so, to make sure this doesn't happen, you should generally not hold them across function-call boundaries.

docs/EphemeralClaims.md
```diff
@@ -0,0 +1,66 @@
...
+
+Ephemeral claims behave much like regular CHERIoT claims, with a few key differences:
+- They belong to a thread rather than a compartment.
```
davidchisnall, line 10: They belong to a thread *executing in a specific compartment*.  As soon as you do a cross-compartment call, they're dropped.

docs/EphemeralClaims.md
```diff
@@ -0,0 +1,66 @@
...
+
+```
+Thread 1: ephemeral_claim(X)
+Thread 1: claim(X)
+ (!) Thread 1 is interrupted in the allocator before grabbing the lock
+Thread 2: free(X)
+```
+
+In the above sequence, Thread 2 succeeds in its attempt to free `X`, since Thread 1 released its ephemeral claim on the object while making the cross-compartment call to the allocator to acquire a persistent claim.
+In fact, this sequence is equivalent to simply not taking an ephemeral claim in the first place, and there is an analyser warning for this pattern.
```
davidchisnall, line 25: This is a pretty obvious failure mode.  The more subtle common one is if you do an ephemeral claim and then acquire a lock.  This is something that you might want to do because you want to hold the lock for as little time as possible and acquiring the ephemeral claim may yield.  If the lock is uncontended, then this works fine.  If the lock is contended then the lock implementation will call `futex_timed_wait` to block.

And this is unlikely to show up in testing, because unless another thread tries to free it from under you, it looks correct.  That's why we have a static analyser to check it.

PR #689, opened 2026-04-10
Add documentation for Privileged Libraries.

tolbooth:
What it says on the tin.

### DIFF
--- a/docs/PrivilegedLibraries.md
+++ b/docs/PrivilegedLibraries.md
@@ -1,0 +1,29 @@
+Privileged Libraries in the CHERIoT RTOS
+========================================
+
+Privileged Libraries are a mechanism in CHERIoT that allow libraries to hold loader-provided sealing keys.
+This enables unprivileged compartments to perform certain privileged operations without the need for a full compartment switch.
+Avoiding the switcher and its associated overheads altogether is desirable from a performance perspective, but requires care that we do not leak secrets.
+
+Mechanics of Privileged Libraries
+---------------------------------
+
+Just as CHERIoT libraries are compartments without a data segment, Privileged Libraries are privileged compartments without a data segment.
+A privileged compartment is a compartment that may have up to three sealing keys placed before its import table, accessible via PC-relative addressing.
+Privileged compartments are built with a different linker script that reserves space for the key slots, which the loader then populates at boot.
+The Allocator, the Scheduler, and the Software Revoker are all examples of such privileged compartments.
+
+Library calls in CHERIoT differ from compartment calls in that they do not need to go through the switcher.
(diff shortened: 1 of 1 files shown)


docs/PrivilegedLibraries.md
```diff
@@ -0,0 +1,29 @@
...
+========================================
+
+Privileged Libraries are a mechanism in CHERIoT that allow libraries to hold loader-provided sealing keys.
+This enables unprivileged compartments to perform certain privileged operations without the need for a full compartment switch.
+Avoiding the switcher and its associated overheads altogether is desirable from a performance perspective, but requires care that we do not leak secrets.
+
+Mechanics of Privileged Libraries
+---------------------------------
+
+Just as CHERIoT libraries are compartments without a data segment, Privileged Libraries are privileged compartments without a data segment.
```
davidchisnall, line 11: I'm not sure I like that framing.  There's a lot more to the difference between privileged libraries and libraries than there is to the difference between privileged compartments and normal compartments.

docs/PrivilegedLibraries.md
```diff
@@ -0,0 +1,29 @@
...
+A privileged compartment is a compartment that may have up to three sealing keys placed before its import table, accessible via PC-relative addressing.
```
davidchisnall, line 12: Note that we have a limited number of sealing otypes, and so the number of privileged libraries is also limited.

davidchisnall: Jumping into the definition of a privileged compartment here is confusing.

docs/PrivilegedLibraries.md
```diff
@@ -0,0 +1,29 @@
...
+
+Despite this semantic equivalence, the abstraction affords us some protection.
+Since sentries are sealed capabilities, they cannot be dereferenced or used for pointer arithmetic.
+The caller's only handle into the library is the sentry, so there is no way to derive a pointer into the library's PCC region and access the sealing keys that reside there outside the context of a call.
+
+There is one subtlety to this, however.
+Privileged libraries are libraries, and the compiler provides no guarantee that all registers are cleared on return from a library call.
+Ordinarily this is handled by the Switcher, but since privileged library calls bypass it entirely, secrets may be left visible to the caller in uncleared registers
+Due to this risk, privileged libraries like the Token Library are small, written in carefully audited assembly, and must manually manage their register state to avoid such leaks.
+Privileged library calls also run with interrupts disabled, preventing an interrupt handler from observing register contents mid-call.
```
davidchisnall, line 28: This isn't quite true.  They use the normal library export mechanism and so don't have to run with interrupts disabled.  In general, they *don't* need to, because thread isolation guarantees some useful non-interference properties..

PR #690, opened 2026-04-17
Stop using riscv32cheriot subarch. This hasn't been needed for at least eight months.

resistor:


davidchisnall (approved):

davidchisnall:
Looks like it is still needed for Rust? @xdoardo

xdoardo:
Yes, we decided the target nomenclature in https://github.com/CHERIoT-Platform/cheri-rust/discussions/17 and we haven't changed it yet. It's worth noting that the `riscv32cheriot-...` is used only in the front-end of the Rust compiler, which will pass to LLVM the correct `cheriot`-suffix-free target name. This to say that we don't depend on LLVM having the `riscv32cheriot-unknown-cheriotrtos` target defined. 

We discussed changing the target name to `riscv32-unknown-cheriotrtos`: this would work fine, but we didn't reach an agreement due to the consequences of applying this naming scheme to the bare-metal target. We were sort of putting the cart before the horses when discussing it, but we didn't understand what would happen when one downloads the Rust toolchain for `riscv32-unknown-unknown` with `rustup`: would it download the libs for our bare-metal CHERIoT target as well? Again, perhaps it's not necessary to think about this problem right now, as maybe we won't even want to add the bare-metal target to Rust and/or won't ever end up being upstreamed or part of the `rustup` infrastructure.

To get these changes merged, it should be enough to replace  the `target` [here](https://github.com/resistor/cheriot-rtos/blob/216e5938672e98104e5e9c65447caf840b1029ac/sdk/xmake.lua#L235) with `"riscv32cheriot-unknown-cheriotrtos"`. An issue can then be raised to the RTOS or Rust repo to remember to fix this when we reach an agreement to rename our Rust target.

resistor:
> To get these changes merged, it should be enough to replace the `target` [here](https://github.com/resistor/cheriot-rtos/blob/216e5938672e98104e5e9c65447caf840b1029ac/sdk/xmake.lua#L235) with `"riscv32cheriot-unknown-cheriotrtos"`. An issue can then be raised to the RTOS or Rust repo to remember to fix this when we reach an agreement to rename our Rust target.

Incorporated this change.

PR #691, opened 2026-04-21
Add sub quotas.

tolbooth:
MVP version of the sub quotas API, the design document for which can be seen at #669.

In particular, this implementation 1. Holds the allocator lock for the entire duration of a `recombine_quota`, and 2. cleaves the quota ID space in half naively, giving the top half to sub quotas, keeping the bottom for static quotas.

Addressing 1. should be done in tandem with `heap_free_all`, as they can rely on the same mechanism for dropping the lock mid-walk.

Addressing 2. can be done using #686 to compute the total number of static quotas and initializing a global in the allocator with this.

### DIFF
--- a/sdk/core/allocator/heap_offset.h
+++ b/sdk/core/allocator/heap_offset.h
@@ -1,0 +1,75 @@
+// Copyright Microsoft and CHERIoT Contributors.
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include <debug.hh>
+#include <stdint.h>
+
+/**
+ * Object representing a compressed pointer to a heap resident object of type T.
+ *
+ * A HeapOffset with `value = 0` represents a `nullptr`.
+ */
+template<typename T>
+struct HeapOffset
+{
+	/// The compressed pointer value as an offset into the allocator heap.
+	uint16_t value = 0;
+
+	/**
+	 * Return true if the underlying value is 0, that is, if this offset
+	 * represents a `nullptr`.
+	 */
+	[[nodiscard]] bool is_null() const
+	{
+		return value == 0;
+	}
+
+	/**
+	 * Decode this offset and return a pointer to T, or `nullptr` if null.
+	 */
+	[[nodiscard]] T *get() const;
+
+	/**
+	 * Encode a pointer to a heap-allocated T as a HeapOffset.
+	 * Asserts that `ptr` is correctly aligned and is within the
+	 * representable range of 16 bits.
+	 */
+	[[nodiscard]] static HeapOffset<T> from(T *ptr);
+
+	/**
+	 * Comparison operators compare the underlying compressed pointer value.
+	 */
+	bool operator==(const HeapOffset &other) const
+	{
+		return value == other.value;
+	}
+
+	bool operator!=(const HeapOffset &other) const
+	{
+		return value != other.value;
+	}
+
+	/**
+	 * Dereference operators allow this to be used as any other pointer.
+	 */
+	T *operator->() const
+	{
+		return get();
+	}
+
+	T &operator*() const
+	{
+		return *get();
+	}
+};
+
+template<typename T>
+struct DebugFormatArgumentAdaptor<HeapOffset<T>>
+{
+	__always_inline static DebugFormatArgument construct(const HeapOffset<T> &h)
+	{
+		return DebugFormatArgumentAdaptor<uint16_t>::construct(h.value);
+	}
+};
--- a/sdk/core/allocator/main.cc
+++ b/sdk/core/allocator/main.cc
@@ -35,25 +35,194 @@ using namespace CHERI;
 Revocation::Revoker revoker;
 namespace
 {
-	/**
-	 * Internal view of an allocator capability.
-	 *
-	 * TODO: For now, these are statically allocated. Eventually we will have
-	 * some that are dynamically allocated. These should be ref counted so that
-	 * we can avoid repeated validity checks.
-	 */
+	/// Internal view of an allocator capability.
 	struct PrivateAllocatorCapabilityState
 	{
 		/// The remaining quota for this capability.
 		size_t quota;
 		/// A unique identifier for this pool.
 		uint16_t identifier;
+		/// Pointers forming DLL for dynamically allocated sub quotas.
+		HeapOffset<PrivateAllocatorCapabilityState> next;
+		HeapOffset<PrivateAllocatorCapabilityState> prev;
 	};
 
 	static_assert(sizeof(PrivateAllocatorCapabilityState) <=
 	              sizeof(AllocatorCapabilityState));
 	static_assert(alignof(PrivateAllocatorCapabilityState) <=
 	              alignof(AllocatorCapabilityState));
+
+	/**
+	 * Represents the assignable ID space for heap resident quotas.
+	 */
+	class SubQuotaIdSpace
+	{
+		static constexpr uint16_t IdMax =
+		  (1 << (MChunkHeader::OwnerIDWidth)) - 1;
+		// FIXME: If we can determine the number of static quotas in the
+		// loader and provide this to the allocator, we can reclaim most of
+		// this ID space.
+		static constexpr uint16_t IdMin =
+		  (1 << (MChunkHeader::OwnerIDWidth - 1));
+		// Full span of the sub-quota ID range as a uint32_t to avoid
+		// truncation when projecting IDs when wrapping around.
+		static constexpr uint32_t IdRange = IdMax - IdMin + 1;
+
+		HeapOffset<PrivateAllocatorCapabilityState> head = {};
+
+		/**
+		 * Helper function that inserts the given quota object after `anchor`
+		 * and assigns it `id`.
+		 */
+		__always_inline void node_insert_after(
+		  PrivateAllocatorCapabilityState            *anchor,
+		  PrivateAllocatorCapabilityState            *node,
+		  HeapOffset<PrivateAllocatorCapabilityState> nodeOffset,
+		  uint16_t                                    id)
+		{
+			Debug::Assert(id >= IdMin && id <= IdMax,
+			              "Provided ID {} is outside sub-quota range [{}, {}]",
+			              id,
+			              IdMin,
+			              IdMax);
+			node->identifier = id;
+			node->next       = anchor->next;
+			node->prev =
+			  HeapOffset<PrivateAllocatorCapabilityState>::from(anchor);
+			anchor->next->prev = nodeOffset;
+			anchor->next       = nodeOffset;
+		}
+
+		/**
+		 * Helper function for picking the offset into a given range of free
+		 * IDs. Here we take the midpoint.
+		 */
+		[[nodiscard]] __always_inline static uint16_t offset_pick(uint32_t gap)
+		{
+			return static_cast<uint16_t>(gap / 2);
+		}
+
+		public:
+		[[nodiscard]] __always_inline static constexpr bool
+		is_sub_quota_id(uint16_t value)
+		{
+			return value >= IdMin;
+		}
+
+		[[nodiscard]] __always_inline bool is_empty() const
+		{
+			return head.is_null();
+		}
+
+		/**
+		 * Insert `node` into the sorted ring. Node is assigned an ID chosen by
+		 * policy described by `offset_pick`. Returns true on success, false if
+		 * the ID space is exhausted.
+		 */
+		[[nodiscard]] __noinline bool
+		id_allocate(PrivateAllocatorCapabilityState *node)
+		{
+			Debug::Assert(node != nullptr, "id_allocate called with null node");
+			Debug::Assert(node->next.is_null() && node->prev.is_null(),
+			              "Node {} already appears to be in ring",
+			              node->identifier);
+
+			auto nodeOffset =
+			  HeapOffset<PrivateAllocatorCapabilityState>::from(node);
+
+			// Ring is empty, so use full ID range for picking offset
+			if (head.is_null())
+			{
+				auto offset = offset_pick(IdRange);
+				Debug::Assert(
+				  offset < IdRange,
+				  "offset_pick returned offset {} outside IdRange {}",
+				  offset,
+				  IdRange);
+				uint16_t id      = static_cast<uint16_t>(IdMin + offset);
+				node->identifier = id;
+				node->next       = nodeOffset;
+				node->prev       = nodeOffset;
+				head             = nodeOffset;
+				return true;
+			}
+
+			auto *start   = head.get();
+			auto *current = start;
+
+			do
+			{
+				auto *next = current->next.get();
+				Debug::Assert(current == next ||
+				                current->identifier != next->identifier,
+				              "Duplicate identifiers in sub-quota ring: {}",
+				              current->identifier);
+
+				// project next's ID above current to handle wraparound nicely
+				uint32_t nextIdProjected = next->identifier;
+				if (nextIdProjected <= current->identifier)
+				{
+					nextIdProjected += IdRange;
+				}
+
+				uint32_t gap = nextIdProjected - current->identifier - 1;
+
+				if (gap > 0)
+				{
+					uint16_t offset = offset_pick(gap);
+					Debug::Assert(offset < gap,
+					              "offset_pick returned offset outside "
+					              "gap [{}, {})",
+					              current->identifier + 1,
+					              next->identifier);
+					uint32_t wideId = current->identifier + 1 + offset;
+					uint16_t id =
+					  (wideId > IdMax)
+					    ? static_cast<uint16_t>(IdMin + (wideId - IdMax - 1))
+					    : static_cast<uint16_t>(wideId);
+
+					// insert the node into the list and advance the head
+					// pointer to the inserted node, such that we reduce
+					// clustering.
+					node_insert_after(current, node, nodeOffset, id);
+					head = nodeOffset;
+					return true;
+				}
+
+				// no free ID range
+				current = next;
+			} while (current != start);
+
+			return false;
+		}
+
+		/**
+		 * Remove `node` from the sorted ring and clear its link pointers.
+		 */
+		void __noinline id_release(PrivateAllocatorCapabilityState *node)
+		{
+			Debug::Assert(node != nullptr, "id_release called with null node");
+			Debug::Assert(!head.is_null(), "release called on empty ring");
+			Debug::Assert(!node->next.is_null() && !node->prev.is_null(),
+			              "Node with ID {} appears not to be in the ring.",
+			              node->identifier);
+
+			if (head.get() == node)
+			{
+				head = (node->next.get() == node)
+				         ? HeapOffset<PrivateAllocatorCapabilityState>{}
+				         : node->next;
+			}
+
+			node->prev->next = node->next;
+			node->next->prev = node->prev;
+			node->next       = {};
+			node->prev       = {};
+		}
+	};
+
+	/// The sub quota ID space
+	SubQuotaIdSpace subQuotaIds;
 
 	// the global memory space
 	MState *gm;
@@ -391,6 +560,10 @@ namespace
 					// Failed to acquire lock within allowed timeout.
 					return -ETIMEDOUT;
 				}
+				if (!Capability{capability}.is_valid())
+				{
+					return -EPERM;
+				}
 				continue;
 			}
 		} while (may_block(timeout));
@@ -428,13 +601,13 @@ namespace
 		if (state->identifier == 0)
 		{
 			static uint32_t nextIdentifier = 1;
-			// Cleave the ID space in half, with upper range reserved for
-			// dynamically allocated subquotas.
-			// FIXME: If we can determine the number of static quotas in the
-			// loader and provide this to the allocator, we can reclaim most of
-			// this ID space.
-			if (nextIdentifier >= (1 << (MChunkHeader::OwnerIDWidth - 1)))
+
+			if (SubQuotaIdSpace::is_sub_quota_id(nextIdentifier))
 			{
+				// If this ID is a sub quota ID, then we've run out of IDs for
+				// our static quotas.
+				Debug::log<DebugLevel::Warning>(
+				  "Exhausted static quota ID space.");
 				return nullptr;
 			}
 			state->identifier = nextIdentifier++;
@@ -502,6 +675,27 @@ namespace
 		}
 
 		/**
+		 * Transfers ownership of this claim. Cannot be transferred to allocator
+		 * ownership.
+		 */
+		void ownership_transfer(uint16_t newOwner)
+		{
+			Debug::Assert(
+			  newOwner != QuotaIdentifierAllocatorOwned,
+			  "newOwner of claim cannot be QuotaIdentifierAllocatorOwned, "
+			  "since the allocator cannot make claims.");
+			allocatorIdentifier = newOwner;
+		}
+
+		/**
+		 * Returns the number of references for this claim.
+		 */
+		[[nodiscard]] uint32_t reference_count() const
+		{
+			return referenceCount;
+		}
+
+		/**
 		 * Returns the value of the compressed next pointer.
 		 */
 		[[nodiscard]] HeapOffset<Claim> encoded_next() const
@@ -518,12 +712,13 @@ namespace
 			/**
 			 * Placeholder value for end iterators.
 			 */
-			HeapOffset<Claim> endPlaceholder = {};
+			static inline const HeapOffset<Claim> EndPlaceholder = {};
 
 			/**
 			 * A pointer to the encoded next pointer.
 			 */
-			HeapOffset<Claim> *encodedNextPointer = &endPlaceholder;
+			HeapOffset<Claim> *encodedNextPointer =
+			  const_cast<HeapOffset<Claim> *>(&EndPlaceholder);
 
 			public:
 			/**
@@ -557,10 +752,9 @@ namespace
 			}
 
 			/// Iteration termination condition.
-			__always_inline bool operator!=(const Iterator Other)
+			__always_inline bool operator!=(const Iterator &other)
 			{
-				return encodedNextPointer->value !=
-				       Other.encodedNextPointer->value;
+				return *encodedNextPointer != *other.encodedNextPointer;
 			}
 
 			/**
@@ -830,6 +1024,7 @@ namespace
 			}
 			return 0;
 		}
+
 		return -EPERM;
 	}
 
@@ -846,6 +1041,7 @@ namespace
 			                                heapCapability);
 			return -EPERM;
 		}
+
 		// Validate the pointer. We do not permit freeing sealed pointers.
 		Capability<void> mem{rawPointer};
 		if (!mem.is_valid() || mem.is_sealed())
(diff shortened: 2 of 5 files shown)


nwf:
Sorry to have not had time to give this the look it deserves.

Quick question in the interim, tho'... in terms of merge strategy, this is going to conflict with #672's transition to larger compressed sizes in headers.  How reliant is this PR on the `uint16_t` nature of its `HeapOffset`-s?  Do we want to land that one and then this one or vice versa?  Is there a common bit of abstraction that'd be useful to land and rebase both atop?  (The `HeapOffset` class might be a good example?)

tolbooth:
> Sorry to have not had time to give this the look it deserves.
> 
> Quick question in the interim, tho'... in terms of merge strategy, this is going to conflict with #672's transition to larger compressed sizes in headers. How reliant is this PR on the `uint16_t` nature of its `HeapOffset`-s? Do we want to land that one and then this one or vice versa? Is there a common bit of abstraction that'd be useful to land and rebase both atop? (The `HeapOffset` class might be a good example?)

Apologies for the delayed response! I think landing #672 first, and then this would be the best order. I can split off the HeapOffset commit into its own PR, which would be that shared nugget, but I think it's also fine for the HeapOffset commit to come after #672 is merged to keep things simple on your end.

Mostly the same work either way, and the rest of this PR is *mostly* agnostic to those changes (32 bit offsets should still fit in existing padding of allocator cap state). What isn't already agnostic would probably be good to make agnostic, and that merge process will be good for forcing my hand there.

tests/allocator-test.cc
```diff
@@ -1471,6 +1471,466 @@ namespace
...
+			TEST(Capability{quotas[i]}.is_valid_temporal(),
+			     "split failed at index {}",
+			     i);
+			ptrs[i] = heap_allocate(&t, quotas[i], allocationSize);
+			TEST(Capability{ptrs[i]}.is_valid_temporal(),
+			     "allocation failed at index {}",
+			     i);
+		}
+
+		// FIXME: volatile here as temporary fix for
```
hlef, line 1857: Sounds like we can get rid of this one since the fix was merged.

tests/allocator-test.cc
```diff
@@ -1471,6 +1471,466 @@ namespace
...
+		debug_log("Beginning sub quota ID management structure tests");
+
+		Timeout               t{UnlimitedTimeout};
+		ds::xoroshiro::P32R16 rand = {};
+
+		/**
+		 * Basic functionality
+		 */
+
+		const size_t                     ArrSize = 64;
```
hlef, line 1485: Let's get rid of this abbreviation (https://github.com/CHERIoT-Platform/cheriot-rtos/blob/main/docs/CodingConventions.md).

tests/allocator-test.cc
```diff
@@ -1471,6 +1471,466 @@ namespace
...
+		std::vector<AllocatorCapability> quotas;
+		quotas.resize(ArrSize);
+		AllocatorCapability child, child2, c0, c1, c2;
```
hlef, line 1488: I think that we can re-use some of these to reduce the number of variables and make the code a bit more readable. Also the naming of these variables could probably be a bit more clear.

tests/allocator-test.cc
```diff
@@ -1471,6 +1471,466 @@ namespace
...
+		 * Failure cases
+		 */
+		debug_log("Testing sub_quota API parameter validation");
+		size_t quotaRemaining = 0;
+
+		// create sub quota
+		child = split_quota(nullptr, MALLOC_CAPABILITY, 0);
+		TEST(!Capability{child}.is_valid(),
+		     "Creating sub quota with null timeout succeded");
+		TEST_EQUAL(static_cast<int>(reinterpret_cast<intptr_t>(child)),
```
hlef, line 1526: I think that this could just be a `reinterpret_cast<int>(child)`. Or using `ErrorOr` and avoid the casts altogether.

hlef: (Same comment multiple times below).

hlef: Also, I think that we can get rid of the `is_valid()` check above as the two checks are a bit redundant. Similar observation a few times below.

tests/allocator-test.cc
```diff
@@ -1471,6 +1471,466 @@ namespace
...
+		child = split_quota(&t, nullptr, 0);
+		TEST(!Capability{child}.is_valid(),
+		     "Creating sub quota with null parent succeeded");
+		TEST_EQUAL(static_cast<int>(reinterpret_cast<intptr_t>(child)),
+		           -EPERM,
+		           "Creating sub quota with null parent succeeded");
+
+		// invalid sizes
+		child = split_quota(&t, MALLOC_CAPABILITY, MALLOC_QUOTA);
+		TEST(!Capability{child}.is_valid(),
```
hlef, line 1538: This test (split with `MALLOC_QUOTA`) feels a bit redundant with the test below (split with the remaining quota). I would get rid of this one.

tests/allocator-test.cc
```diff
@@ -1471,6 +1471,466 @@ namespace
...
+		TEST_EQUAL(recombine_quota(child, nullptr),
+		           -EPERM,
+		           "recombining sub quota with null parent succeeded");
+		child2 = split_quota(&t, MALLOC_CAPABILITY, 0);
+		TEST(Capability{child2}.is_valid(),
+		     "Sub quota creation failed for null parent recombine test");
+		TEST_EQUAL(recombine_quota(child, child2),
+		           -EPERM,
+		           "recombining sub quota with sibling as parent succeeded")
+		TEST_EQUAL(recombine_quota(child2, MALLOC_CAPABILITY),
```
hlef, line 1572: I would separate these two recombines visually as they are here for cleanup.

tests/allocator-test.cc
```diff
@@ -1471,6 +1471,466 @@ namespace
...
+		TEST(!Capability{child}.is_valid(),
+		     "Creating sub quota larger than parent succeeded");
+		TEST_EQUAL(
+		  static_cast<int>(reinterpret_cast<intptr_t>(child)),
+		  -ENOMEM,
+		  "Creating sub quota with quota greater than parent succeeded");
+
+		quotaRemaining = heap_quota_remaining(MALLOC_CAPABILITY);
+		child          = split_quota(
+		  &t, MALLOC_CAPABILITY, static_cast<size_t>(quotaRemaining));
```
hlef, line 1547: I think that we can call `heap_quota_remaining` directly in the call to `split_quota` to make things a bit more readable, since we don't use that information later.

tests/allocator-test.cc
```diff
@@ -1471,6 +1471,466 @@ namespace
...
+		TEST_EQUAL(recombine_quota(child2, MALLOC_CAPABILITY),
+		           0,
+		           "Cleanup recombine of child2 failed.");
+		TEST_EQUAL(recombine_quota(child, MALLOC_CAPABILITY),
+		           0,
+		           "Cleanup recombine child failed.");
+
+		/**
+		 * Quota Accounting
+		 */
```
hlef, line 1581: I think that we can combine all the tests in this category with the tests above. It will make the test a lot more compact.

tests/allocator-test.cc
```diff
@@ -1471,6 +1471,466 @@ namespace
...
+	/**
+	 */
+	__noinline void test_sub_quota_semantics()
+	{
+		debug_log("Beginning sub quota semantics tests");
+
+		Timeout               t{UnlimitedTimeout};
+		ds::xoroshiro::P32R16 rand = {};
+
+		const size_t                     ArrSize   = 64;
```
hlef, line 1667: Similar comment as above on the abbreviation.

tests/allocator-test.cc
```diff
@@ -1471,6 +1471,466 @@ namespace
 		debug_log("End of permission restriction tests");
 	}
 
+	__noinline void test_sub_quota_operations()
```
hlef, line 1474: Having a comment on what this test does would be helpful to understand why there are two tests. Essentially, one checks that the creation and deletion of sub-quotas works as expected, and the other checks that sub-quotas themselves work as expected.

tests/allocator-test.cc
```diff
@@ -1471,6 +1471,466 @@ namespace
...
+		// If a child has a claim on an object it doesn't own, that claim is
+		// transferred to the parent. Test with parent as owner and sibling as
+		// owner.
+
+		// Parent owns object
+		ptr = heap_allocate(&t, MALLOC_CAPABILITY, allocationSize);
+		TEST(Capability{ptr}.is_valid(), "Parent allocation failed");
+
+		child = split_quota(&t, MALLOC_CAPABILITY, splitSize);
+		AllocatorCapability child2 =
```
hlef, line 1733: Probably worth declaring this variable at the top to be consistent.

tests/allocator-test.cc
```diff
@@ -1471,6 +1471,466 @@ namespace
...
+		splitSize = allocationSize + (2 * sizeof(AllocatorCapabilityState));
+
+		quotas[0] = MALLOC_CAPABILITY;
+		ptrs[0]   = heap_allocate(&t, MALLOC_CAPABILITY, allocationSize);
+		TEST(Capability{ptrs[0]}.is_valid(), "Root allocation failed");
+
+		for (size_t i = 1; i < ArrSize; ++i)
+		{
+			quotas[i] =
+			  split_quota(&t, quotas[i - 1], (ArrSize - i) * splitSize);
```
hlef, line 1847: This line could use a comment: the ith quota is just large enough to hold `splitSize` worth of allocations and be split recursively into ArraySize-i more quotas that follow the same rule. The last quota cannot be further split and has a size of `splitSize`.

hlef (commented):
I had a first look at the tests. I love how exhaustive they are! Mostly minor comments.

sdk/core/allocator/heap_offset.h
```diff
@@ -0,0 +1,75 @@
...
+
+#pragma once
+
+#include <debug.hh>
+#include <stdint.h>
+
+/**
+ * Object representing a compressed pointer to a heap resident object of type T.
+ *
+ * A HeapOffset with `value = 0` represents a `nullptr`.
```
hlef, line 12: Probably worth a comment explaining why using 0 to encode this isn't a problem. Offset 0 in the heap corresponds to the MState header. There is no real use-case for encoding a pointer to the MState header as a heap offset because there is only one such object and we already have a pointer to that object at hand anyways.

sdk/core/allocator/main.cc
```diff
@@ -34,26 +35,195 @@ using namespace CHERI;
...
+			return head.is_null();
+		}
+
+		/**
+		 * Insert `node` into the sorted ring. Node is assigned an ID chosen by
+		 * policy described by `offset_pick`. Returns true on success, false if
+		 * the ID space is exhausted.
+		 */
+		[[nodiscard]] __noinline bool
+		id_allocate(PrivateAllocatorCapabilityState *node)
```
hlef, line 123: This should probably be called `allocate_id` instead (https://github.com/CHERIoT-Platform/cheriot-rtos/blob/main/docs/CodingConventions.md).

sdk/core/allocator/main.cc
```diff
@@ -34,26 +35,195 @@ using namespace CHERI;
...
+				current = next;
+			} while (current != start);
+
+			return false;
+		}
+
+		/**
+		 * Remove `node` from the sorted ring and clear its link pointers.
+		 */
+		void __noinline id_release(PrivateAllocatorCapabilityState *node)
```
hlef, line 202: This should probably be called `release_id` instead (https://github.com/CHERIoT-Platform/cheriot-rtos/blob/main/docs/CodingConventions.md).

sdk/core/allocator/main.cc
```diff
@@ -390,6 +560,10 @@ namespace
 					// Failed to acquire lock within allowed timeout.
 					return -ETIMEDOUT;
 				}
+				if (!Capability{capability}.is_valid())
```
hlef, line 563: This would need a dedicated comment. It's here to handle the case where we drop the lock and yield and the sub-quota we were using got freed in the meantime.

sdk/core/allocator/main.cc
```diff
@@ -1432,6 +1679,225 @@ __cheriot_minimum_stack(TokenObjDestroyStackUsage) int token_obj_can_destroy(
...
+{
+	STACK_CHECK(0x270);
+	if (!check_timeout_pointer(timeout))
+	{
+		return reinterpret_cast<AllocatorCapability>(-EINVAL);
+	}
+
+	LockGuard g{lock};
+
+	auto *parentCap = malloc_capability_unseal(parent, AllocatorPermitAllocate);
```
hlef, line 1693: It is not 100% obvious to me that `AllocatorPermitAllocate` is the right permission to require here. Shouldn't it be `AllocatorPermitFull`?

`split_quota` should not grant the caller compartment more permissions than it already had.

If the compartment has only allocation permissions, then it can allocate things, but cannot free them. But if the compartment uses `split_quota`, it suddently becomes able to allocate things *and* free them.

It's a minor increase in privileges since the compartment still cannot free the allocations that had been previously done by the parent allocation capability, but an increase still.

sdk/include/allocator.h
```diff
@@ -70,6 +70,55 @@ allocator_permissions_and(AllocatorCapability heapCapability, int permissions)
...
+ *
+ * The allocator capability returned from this function on success may be used
+ * in place of a static allocation capability in any heap API, but care should
+ * be taken: the quota is a heap allocated object, and so may be freed.
+ *
+ * This will return an AllocatorCapability that refers to the newly created sub
+ * quota on success.
+ *
+ * Returns `-EPERM` if `parent` is not a valid allocator capability or does not
+ * have `AllocatorPermitAllocate` permission, `-ENOMEM` if `parent` has
```
hlef, line 88: See my comment above on the permissions required on the parent allocator capability.

sdk/core/allocator/main.cc
davidchisnall: No, see the example [in the doc you lonked](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/main/docs/CodingConventions.md#start-with-the-most-generic-word)

sdk/core/allocator/main.cc
davidchisnall: As above, nope.

sdk/core/allocator/main.cc
hlef: Great, thanks!

sdk/core/allocator/main.cc
```diff
@@ -642,48 +848,16 @@ namespace
...
 	              "Claims should fit in the smallest possible allocation");
 
 	/**
-	 * Find a claim if one exists.  Returns a reference to the next pointer
+	 * Find a claim if one exists.  Returns a pointer to the next HeapOffset
 	 * that refers to this claim.
 	 */
-	std::pair<uint16_t &, Claim *> claim_find(uint16_t      owner,
-	                                          MChunkHeader &chunk)
+	std::pair<HeapOffset<Claim> *, Claim *> claim_find(uint16_t      owner,
```
hlef, line 859: What is the rationale for making this return a pointer instead of a reference? I think that we could make this return a `HeapOffset<Claim> &` instead and avoid a lot of awkward dereferences down the line.

sdk/core/allocator/main.cc
```diff
@@ -642,48 +848,16 @@ namespace
...
-			return address;
-		}
 	};
 	static_assert(sizeof(Claim) <= (1 << MallocAlignShift),
 	              "Claims should fit in the smallest possible allocation");
 
 	/**
-	 * Find a claim if one exists.  Returns a reference to the next pointer
+	 * Find a claim if one exists.  Returns a pointer to the next HeapOffset
 	 * that refers to this claim.
```
hlef, line 857: Looking at this function makes me realize it'd be nice to document the case where there is no claim here.

sdk/core/allocator/main.cc
```diff
@@ -1304,6 +1550,7 @@ __cheriot_minimum_stack(0x290) CHERI_SEALED(void *)
 	STACK_CHECK(0x290);
 	if (!check_timeout_pointer(timeout))
 	{
+		Debug::log<DebugLevel::Warning>("Timeout invalid.");
```
hlef, line 1553: Debug change that made it into the PR? No other function prints on invalid timeout. I'd remove that.

sdk/core/allocator/main.cc
```diff
@@ -489,10 +674,31 @@ namespace
...
+			  newOwner != QuotaIdentifierAllocatorOwned,
+			  "newOwner of claim cannot be QuotaIdentifierAllocatorOwned, "
+			  "since the allocator cannot make claims.");
+			allocatorIdentifier = newOwner;
+		}
+
+		/**
+		 * Returns the number of references for this claim.
+		 */
+		[[nodiscard]] uint32_t reference_count() const
```
hlef, line 693: Nit: We can move that down next to `reference_add` and `reference_remove`.

sdk/core/allocator/main.cc
```diff
@@ -534,21 +740,21 @@ namespace
...
 			 */
 			__always_inline Claim *operator->()
 			{
-				return Claim::from_encoded_offset(*encodedNextPointer);
+				return encodedNextPointer->get();
 			}
 
 			/// Iteration termination condition.
-			__always_inline bool operator!=(const Iterator Other)
+			__always_inline bool operator!=(const Iterator &other)
```
hlef, line 755: This change does not seem necessary? Not sure why we need to pass the iterator by reference.

sdk/core/allocator/main.cc
```diff
@@ -1155,62 +1363,22 @@ namespace
...
-			Debug::log<DebugLevel::Warning>(
-			  "Requested size {} is too large to include header",
-			  requestedSize);
-			return {nullptr, nullptr};
-		}
-
-		LockGuard g{lock};
-		auto     *capability =
-		  malloc_capability_unseal(heapCapability, AllocatorPermitAllocate);
 		if (capability == nullptr)
```
hlef, line 1378: Checking this in the internal function feels wrong. Also, the checks ends up duplicated on the `split_quota` path. This should probably be moved into `allocate_sealed_unsealed`.

sdk/core/allocator/main.cc
```diff
@@ -1432,6 +1679,225 @@ __cheriot_minimum_stack(TokenObjDestroyStackUsage) int token_obj_can_destroy(
...
+	// Allocate the quota to the child
+	parentCap->quota -= size;
+	unsealed->quota = size;
+
+	if (subQuotaIds.id_allocate(unsealed))
+	{
+		return sealed;
+	}
+
+	// the unsealed capability returned from allocate_sealed_unsealed_internal
```
hlef, line 1734: Very much a nit but let's make this start with a capital T 🙂

hlef: It would be nice to better document that this is the path where we failed to allocate a sub-quota ID.

sdk/core/allocator/main.cc
```diff
@@ -1432,6 +1679,225 @@ __cheriot_minimum_stack(TokenObjDestroyStackUsage) int token_obj_can_destroy(
...
+	Debug::log<DebugLevel::Warning>(
+	  "Failed to allocate ID in split_quota, freeing {}", unsealedFull);
+
+	int ret = heap_free_internal(parent, unsealedFull, true);
+
+	if (ret < 0)
+	{
+		Debug::log<DebugLevel::Warning>(
+		  "Failed to free allocated child after ID allocation failure. "
+		  "Free manually with recombine_quota.");
```
hlef, line 1750: It is unclear what "Free manually with recombine_quota" means here?

sdk/include/allocator.h
```diff
@@ -70,6 +70,55 @@ allocator_permissions_and(AllocatorCapability heapCapability, int permissions)
...
+ * insufficient quota to accommodate the requested `size` plus the overhead of
+ * the sub-quota object itself, or `-EINVAL` if either  `timeout` is not a valid
+ * timeout pointer or the internal allocation fails for any reason (including
+ * timeout).
```
hlef, line 92: We should probably document the error value used when we run out of quota IDs.

sdk/core/allocator/main.cc
```diff
@@ -1432,6 +1679,225 @@ __cheriot_minimum_stack(TokenObjDestroyStackUsage) int token_obj_can_destroy(
...
+		return reinterpret_cast<AllocatorCapability>(-EINVAL);
+	}
+
+	parentCap->quota += size;
+	return reinterpret_cast<AllocatorCapability>(-ENOMEM);
```
hlef, line 1755: Not sure if ENOMEM is the right error code here? See my comment in the API definition.

sdk/core/allocator/main.cc
```diff
@@ -34,26 +35,195 @@ using namespace CHERI;
...
+		// truncation when projecting IDs when wrapping around.
+		static constexpr uint32_t IdRange = IdMax - IdMin + 1;
+
+		HeapOffset<PrivateAllocatorCapabilityState> head = {};
+
+		/**
+		 * Helper function that inserts the given quota object after `anchor`
+		 * and assigns it `id`.
+		 */
+		__always_inline void node_insert_after(
```
hlef, line 77: Not sure if this helper is really necessary, since it is quite small and has just one caller (`id_allocate`). Also, passing `nodeOffset` doesn't really feel necessary since it's easy to compute from `node`.

sdk/core/allocator/main.cc
```diff
@@ -34,26 +35,195 @@ using namespace CHERI;
...
+			node->next       = anchor->next;
+			node->prev =
+			  HeapOffset<PrivateAllocatorCapabilityState>::from(anchor);
+			anchor->next->prev = nodeOffset;
+			anchor->next       = nodeOffset;
+		}
+
+		/**
+		 * Helper function for picking the offset into a given range of free
+		 * IDs. Here we take the midpoint.
```
hlef, line 98: It would be nice to explain why we take the midpoint. The idea is that, since these are always size 1 allocations, we want the space to be maximally fragmented to avoid the case of allocating a large chunk of contiguous IDs then having to walk over all of them to find the next free ID.

sdk/core/allocator/main.cc
```diff
@@ -34,26 +35,195 @@ using namespace CHERI;
...
+					uint16_t offset = offset_pick(gap);
+					Debug::Assert(offset < gap,
+					              "offset_pick returned offset outside "
+					              "gap [{}, {})",
+					              current->identifier + 1,
+					              next->identifier);
+					uint32_t wideId = current->identifier + 1 + offset;
+					uint16_t id =
+					  (wideId > IdMax)
+					    ? static_cast<uint16_t>(IdMin + (wideId - IdMax - 1))
```
hlef, line 181: The calculations look right to me, but I wish we had comments that explain why the -1 and +1 here and there.

sdk/core/allocator/main.cc
```diff
@@ -1432,6 +1679,225 @@ __cheriot_minimum_stack(TokenObjDestroyStackUsage) int token_obj_can_destroy(
...
+	  std::move(g), timeout, parentCap, key, sealedSize.value());
+
+	if (!sealedWrapped.is_valid())
+	{
+		return reinterpret_cast<AllocatorCapability>(-EINVAL);
+	}
+
+	// Slightly ugly solution. SealedTokenHandle doesn't expose .get() so we pun
+	// through pointer for correct return value.
+	auto sealed = *reinterpret_cast<AllocatorCapability *>(&sealedWrapped);
```
hlef, line 1720: @davidchisnall @nwf do we have a nicer way to do this?

sdk/core/allocator/main.cc
hlef: On another note, it's unclear to me in which cases this free could fail. @nwf @davidchisnall opinions?

sdk/core/allocator/main.cc
```diff
@@ -1432,6 +1679,225 @@ __cheriot_minimum_stack(TokenObjDestroyStackUsage) int token_obj_can_destroy(
...
+
+	// Release the child ID
+	subQuotaIds.id_release(childState);
+
+	// Now really free the child
+	ret = heap_free_internal(parent, unsealedChild, true);
+
+	if (ret != 0)
+	{
+		Debug::log<DebugLevel::Warning>("Failed to destroy sub quota {}",
```
hlef, line 1892: Apart from a bug, this should never happen as we checked if we can free earlier?

hlef:
Alright, done with a complete review. The PR is really good overall. Once @davidchisnall or @nwf had a chance to make a review, I can take some time to address the issues as @tolbooth likely won't be able to.

sdk/core/allocator/main.cc
hlef: As bonus for future works, we probably want to have a dedicated permission for this. We have one unused permission slot left we can use for this.

sdk/core/allocator/main.cc
```diff
@@ -427,7 +428,12 @@ namespace
 		if (state->identifier == 0)
 		{
 			static uint32_t nextIdentifier = 1;
-			if (nextIdentifier >= (1 << MChunkHeader::OwnerIDWidth))
+			// Cleave the ID space in half, with upper range reserved for
+			// dynamically allocated subquotas.
+			// FIXME: If we can determine the number of static quotas in the
+			// loader and provide this to the allocator, we can reclaim most of
+			// this ID space.
+			if (nextIdentifier >= (1 << (MChunkHeader::OwnerIDWidth - 1)))
```
nwf, line 436: Very nit-picky, but this hunk ended up in the `HeapOffset` commit and probably shouldn't have?

sdk/core/allocator/heap_offset.h
```diff
@@ -0,0 +1,75 @@
...
+ */
+template<typename T>
+struct HeapOffset
+{
+	/// The compressed pointer value as an offset into the allocator heap.
+	uint16_t value = 0;
```
nwf, line 18: This is going to make moving to anything strictly between 16 and 32 in size for the `HeapOffset`'s underlying type challenging, and that's sad, because I'd love to have this notion of "value used as absolute offset into the heap" abstracted out like this.

nwf (changes requested):

sdk/core/allocator/main.cc
```diff
@@ -35,26 +35,195 @@ using namespace CHERI;
...
+	/// Internal view of an allocator capability.
 	struct PrivateAllocatorCapabilityState
 	{
 		/// The remaining quota for this capability.
 		size_t quota;
 		/// A unique identifier for this pool.
 		uint16_t identifier;
+		/// Pointers forming DLL for dynamically allocated sub quotas.
+		HeapOffset<PrivateAllocatorCapabilityState> next;
+		HeapOffset<PrivateAllocatorCapabilityState> prev;
```
nwf, line 47: A good chunk of the allocator uses the `ds::linked_list` machinery already, and it might be nice to do that here, too.  It would avoid yet another implementation scattered through helpers and consumers, but I won't insist.

sdk/core/allocator/main.cc
```diff
@@ -1444,6 +1679,225 @@ __cheriot_minimum_stack(TokenObjDestroyStackUsage) int token_obj_can_destroy(
...
+							// parents and destroy the child's claim object. We
+							// loop here since claim counts are expected to be
+							// small, and we want to respect the saturation
+							// logic.
+							for (uint32_t i = 0;
+							     i < childClaim->reference_count();
+							     ++i)
+							{
+								parentClaim->reference_add();
+							}
```
nwf, line 1847: All the same, saturating add is sufficiently straightforward to do in constant time... I think I'd prefer `reference_add` be made to take a "how many" argument.

sdk/core/allocator/main.cc
```diff
@@ -1444,6 +1679,225 @@ __cheriot_minimum_stack(TokenObjDestroyStackUsage) int token_obj_can_destroy(
...
+	}
+
+	uint16_t childID  = childState->identifier;
+	uint16_t parentID = parentState->identifier;
+	if (!SubQuotaIdSpace::is_sub_quota_id(childID))
+	{
+		return -EINVAL;
+	}
+
+	// Sealed pointer was an internal address, we want to free from base.
```
nwf, line 1781: This cannot fail because `malloc_capability_unseal` returned non-`nullptr`, right?

sdk/core/allocator/main.cc
nwf: This should be a `Debug::Invariant`, yes.

sdk/core/allocator/main.cc
```diff
@@ -501,6 +674,27 @@ namespace
 			return allocatorIdentifier;
 		}
 
+		/**
+		 * Transfers ownership of this claim. Cannot be transferred to allocator
+		 * ownership.
+		 */
```
nwf, line 680: Add a comment noting that it is the caller's responsibility to ensure that an header's `Claims` chain has at most one entry for each claimant?

sdk/core/allocator/main.cc
```diff
@@ -35,26 +35,195 @@ using namespace CHERI;
...
+		// FIXME: If we can determine the number of static quotas in the
+		// loader and provide this to the allocator, we can reclaim most of
+		// this ID space.
+		static constexpr uint16_t IdMin =
+		  (1 << (MChunkHeader::OwnerIDWidth - 1));
+		// Full span of the sub-quota ID range as a uint32_t to avoid
+		// truncation when projecting IDs when wrapping around.
+		static constexpr uint32_t IdRange = IdMax - IdMin + 1;
+
+		HeapOffset<PrivateAllocatorCapabilityState> head = {};
```
nwf, line 71: Document that this points to a sentinel-free ring of the dynamic quota objects and that we'll "rotate" our view of the ring to try to point to a quota whose ring-successor is far away in owner-ID space so that subsequent allocations of quota objects do not often need to hunt for a free ID?

sdk/core/allocator/main.cc
```diff
@@ -35,26 +35,195 @@ using namespace CHERI;
...
 namespace
 {
-	/**
-	 * Internal view of an allocator capability.
-	 *
-	 * TODO: For now, these are statically allocated. Eventually we will have
-	 * some that are dynamically allocated. These should be ref counted so that
-	 * we can avoid repeated validity checks.
-	 */
+	/// Internal view of an allocator capability.
```
nwf, line 38: Document that these may appear in the heap and that, if they do, the parent/child relationship is encoded by the child object being owned by the parent rather than anything in the object payload itself.

sdk/core/allocator/main.cc
```diff
@@ -35,26 +35,195 @@ using namespace CHERI;
...
+		HeapOffset<PrivateAllocatorCapabilityState> prev;
 	};
 
 	static_assert(sizeof(PrivateAllocatorCapabilityState) <=
 	              sizeof(AllocatorCapabilityState));
 	static_assert(alignof(PrivateAllocatorCapabilityState) <=
 	              alignof(AllocatorCapabilityState));
 
+	/**
+	 * Represents the assignable ID space for heap resident quotas.
```
nwf, line 56: I suspect this whole scheme is somewhat my fault, so I apologize, but...

If we do the violence in #672, do we have enough bits to instead make the initial claims "pointer" in the `MChunkHeader` a tagged union of a static identifier and a heap address?  We will have stripped out the encoded link to the predecessor object, which should give us enough room to instead...
* add a bit to distinguish static and dynamic owners,
* encode an absolute heap offset in the case of dynamic owner, or
* encode a sequential identifier for static owner.

In that case, I think we don't even need the ring of all dynamic claims?

nwf (changes requested):

sdk/core/allocator/main.cc
hlef: Sounds good, but any way we make this future works once #672 is merged?

sdk/core/allocator/heap_offset.h
hlef: Unfortunately yes... is that a deal breaker? This is technically a pre-existing limitation of the `from_encoded_offset` machinery. Can we get away with proper documentation?

nwf: I don't know that it's a deal breaker exactly, but it is certainly something to ponder.  I don't love the approach I took in #727, but possibly something like that could be generalized.

hlef: (Discussed off-channel, posting here for the record)

I'm going to strip away the `HeapOffset` (manually call `from_encoded_offset` and friends) and we can refactor this properly at the scale of the allocator at a later stage.

sdk/core/allocator/main.cc
hlef: (Discussed off-channel, posting here for the record)

We'll go with this for now to ensure that the PR doesn't linger forever, and can improve and optimize afterwards.

hlef:
Updated PR here: https://github.com/CHERIoT-Platform/cheriot-rtos/pull/732

We can close this one.

PR #692, opened 2026-04-27
Add `volatile` qualifier in MMIO_CAPABILTY_WITH_PERMISSIONS macro

xdoardo:
As noted here: https://github.com/CHERIoT-Platform/cheriot-rtos/commit/2d7a16ba6972f86a22f90e9545d3b0860eae0e67#r183545095.

### DIFF
--- a/sdk/include/compartment-macros.h
+++ b/sdk/include/compartment-macros.h
@@ -119,7 +119,7 @@
 	                                         permitLoadMutable,                \
 	                                         permitLoadGlobal)                 \
 		MMIO_CAPABILITY_WITH_PERMISSIONS_INNER(                                \
-		  type,                                                                \
+		  volatile type,                                                       \
 		  name,                                                                \
 		  permitLoad,                                                          \
 		  permitStore,                                                         \
--- a/tests/mmio-test.cc
+++ b/tests/mmio-test.cc
@@ -16,6 +16,11 @@ void check_permissions(Capability<volatile void> mmio, PermissionSet p)
 
 int test_mmio()
 {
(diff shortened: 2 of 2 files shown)


rmn30 (approved):
LGTM

davidchisnall (approved):

rmn30:
I'm a bit surprised this didn't break more things. I guess it just depends what the compiler chooses to optimise?

sdk/include/compartment-macros.h
```diff
@@ -119,7 +119,7 @@
 	                                         permitLoadMutable,                \
 	                                         permitLoadGlobal)                 \
 		MMIO_CAPABILITY_WITH_PERMISSIONS_INNER(                                \
-		  type,                                                                \
```
nwf, line 122: Not that I object too much, but why here and not line 109?

rmn30: Only because that is more similar to what the other version of this macro does at line 165.

sdk/include/compartment-macros.h
nwf, line 108: Do we want to follow David's suggestion and make this `extern type` (now that `type` is itself `volatile` qualified)?

rmn30: I tried this but it fails to compile because in `stdio.h` there is an `MMIO_CAPABILITY(void, uart)` and you can't have a global of type void. Any other ideas?

nwf: It's probably fine as is; I don't think the type of `cValue` is actually used given the C style cast in the result of the statement expression.

PR #693, opened 2026-04-27
Strengthen static_asserts on MMIO capability permissions.

rmn30:
During review of #692 I noticed that these static asserts do not make
sense: permitLoadStoreCapability is not permitted without either
permitLoad or PermitStore and permitLoadMutable is not permitted with
permitLoad and permitLoadStoreCapability. The previous assertions
would therefore pass incorrectly if one of the subordinate permissions
was requested without the required prerequisites. permitLoad or
permitStore is the minimum sensible set of permissions for an
MMIO_CAPABILITY so just check for this. We could extend the assertion
to check the dependencies of subordinate permissions but the compiler
checks this anyway.

### DIFF
--- a/sdk/include/compartment-macros.h
+++ b/sdk/include/compartment-macros.h
@@ -91,8 +91,7 @@
 	  cValue)                                                                  \
 		({                                                                     \
 			/* NOLINTBEGIN(bugprone-macro-parentheses) */                      \
-			_Static_assert(permitLoad || permitStore ||                        \
-			                 permitLoadStoreCapabilities || permitLoadMutable, \
+			_Static_assert(permitLoad || permitStore,                          \
 			               "Importing an MMIO capability with no permissions " \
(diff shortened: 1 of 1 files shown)


davidchisnall:
The compiler will reject invalid things.  The problem is that the compiler accepts no permissions as meaning all permissions (so you can just request everything without enumerating it all), but these macros don't want to accidentally turn an incorrect 'I didn't ask for any permissions' (which is likely a bug) into 'I now have all of the permissions'.

rmn30:
This change won't allow that as it requires either permitLoad or permitStore.

nwf:
See also https://github.com/CHERIoT-Platform/cheriot-rtos/blob/8af2cf65a46d09d1028c3551911be2804781cd48/tests/mmio-test.cc#L24-L33

nwf (approved):

PR #698, opened 2026-04-30
Remove memcpy from the loader.

davidchisnall:
To date, every use of memcpy in the loader has been a bug.  There was one left (actually two after inlining) and it was a bug.  And then, in debug builds, there were two more instances of the same bug.

Fix the bug (in all three places) and now we don't need it.


davidchisnall:
Pushed the wrong branch to the new branch :-(

PR #699, opened 2026-04-30
Add run-once, barriers and condition variables to the locks library.

davidchisnall:
These are all built on futexes.

Along the way, fix a bug where futex wait with a zero timeout was spuriously succeeding.  This is permitted by the API contract (spurious wakes are allowed) but makes it harder to write tests.

Also fix a missing cast in std::atomic that meant it didn't work with enums.


sdk/core/scheduler/main.cc
```diff
@@ -139,7 +139,7 @@ namespace
 	{
 		Thread::walk_thread_list(futexWaitingList, [&](Thread *thread) {
 			if ((thread->futexPriorityInheriting) &&
-			    (thread->futexWaitAddress = key))
```
rmn30, line 142: I thought we already fixed this little horror. Is this another instance of the same?

davidchisnall: I thought we had too.  Not sure what happened.

rmn30: Indeed 1e41fff9e05e36234e5b1066179abee37730fe91 fixed the one in `priority_boost_update` but left this one in `priority_boost_reset`. I thought I checked for more instances at the time but apparently not well enough 😢

sdk/include/locks.h
```diff
@@ -68,6 +68,66 @@ struct CountingSemaphoreState
...
+ * State-machine states for run-once functions.
+ */
+enum OnceStateMachineStates : uint32_t
+{
+	/**
+	 * The function has not run yet.
+	 */
+	OnceStateNotRun = 0,
+	/**
+	 * The function has not yet run, but is running.
```
rmn30:
```suggestion
	 * The function is running and has not completed yet.
```

sdk/include/locks.h
```diff
@@ -220,4 +280,86 @@ int __cheri_libcall semaphore_get(Timeout                       *timeout,
...
+/**
+ * Wait on a barrier.  Returns 0 or 1 on success, or `-ETIMEDOUT` on failure.
+ *
+ *  - 0 indicates that this was the last thread to reach the barrier.
+ *  - 1 indicates that this was a thread that blocked on the barrier and has now
+ *    woken.
+ *  - `-ETIMEDOUT` indicates that not all threads reached the barrier prior to
+ * the timeout.
+ *
+ * Note that barriers that are reached via
```
rmn30: don't leave us hanging!

sdk/include/locks.h
```diff
@@ -220,4 +280,86 @@ int __cheri_libcall semaphore_get(Timeout                       *timeout,
  */
 int __cheri_libcall semaphore_put(struct CountingSemaphoreState *semaphore);
 
+/**
+ * Wait on a barrier.  Returns 0 or 1 on success, or `-ETIMEDOUT` on failure.
+ *
+ *  - 0 indicates that this was the last thread to reach the barrier.
+ *  - 1 indicates that this was a thread that blocked on the barrier and has now
+ *    woken.
+ *  - `-ETIMEDOUT` indicates that not all threads reached the barrier prior to
```
rmn30, line 289: Are there any other possible return values? `-EINVAL` for bad parameter? -1 for failed compartment call?

sdk/include/locks.h
```diff
@@ -220,4 +280,86 @@ int __cheri_libcall semaphore_get(Timeout                       *timeout,
...
+{
+	Timeout t = {0, UnlimitedTimeout};
+	return barrier_timed_wait(&t, barrier);
+}
+
+/**
+ * Signal the condition variable and wake up to the number of waiters specified
+ * by the second argument.
+ *
+ * Returns 0 on success.  Currently has no failure conditions.
```
rmn30: Except perhaps `-ECOMPARTMENTFAIL`?

sdk/include/locks.h
```diff
@@ -220,4 +280,86 @@ int __cheri_libcall semaphore_get(Timeout                       *timeout,
...
+ * thread calling `condition_variable_wait` will sleep forever.
+ */
+int __cheriot_libcall
+condition_variable_wait(Timeout                       *t,
+                        struct ConditionVariableState *conditionVariable,
+                        void                          *mutex,
+                        int (*mutexLock)(Timeout *, void *),
+                        int (*mutexUnlock)(void *));
+
+int __cheriot_libcall run_once_slow_path(struct OnceState *state,
```
rmn30: I guess this isn't intended to be called directly. Could we indicate this either with a comment, naming convention or something else? I guess you can't declare it in `run_once` because `C`?

sdk/include/locks.hh
```diff
@@ -381,4 +381,75 @@ class LockGuard
...
+class ConditionVariable
+{
+	/**
+	 * The state for the underlying condition variable.
+	 */
+	ConditionVariableState state;
+
+	public:
+	/**
+	 * Wake one waiter.
```
rmn30: Would be good to include the documentation here. Perhaps with a link?

sdk/include/locks.hh
```diff
@@ -381,4 +381,75 @@ class LockGuard
...
+		};
+		return condition_variable_wait(t, &state, &mutex, lock, unlock);
+	}
+
+	/**
+	 * Atomically release `mutex`, wait until the condition variable is
+	 * signalled, and reacquire the mutex.  This variant uses locks with no
+	 * timeout and so may block for longer than the specified timeout while
+	 * reacquiring the lock.  It will not attempt to reacquire the lock if the
+	 * timeout happens while waiting, so can still fail in the same ways as the
```
rmn30: I found this confusing. How can the timeout happen if the locks are not using timeouts? I guess it can still timeout on the CV.

sdk/lib/locks/barrier.cc
```diff
@@ -0,0 +1,40 @@
...
+	{
+		Debug::log("Last thread arrived, waking barrier {}", barrier);
+		barrier->remaining.notify_all();
+		return 0;
+	}
+
+	// If there are still waiters, we're going to sleep.
+	while (remaining != 0)
+	{
+		Debug::log("No the last thread, sleeping on barrier {}", barrier);
```
rmn30:
```suggestion
		Debug::log("Not the last thread, sleeping on barrier {}", barrier);
```

sdk/lib/locks/condition_variable.cc
```diff
@@ -0,0 +1,32 @@
+#include "lock_debug.hh"
+#include <locks.h>
+
+int condition_variable_notify(ConditionVariableState *conditionVariable,
+                              uint32_t                waiters)
+{
+	conditionVariable->sequenceCounter++;
+	futex_wake(
```
rmn30: Why is it OK to ignore the return value of `futex_wake`?

tests/locks-test.cc
```diff
@@ -394,6 +394,88 @@ namespace
...
+	 * Test the barrier functionality.
+	 */
+	void test_barrier()
+	{
+		static BarrierState barrier{3};
+		Timeout             t{1};
+		int                 ret = barrier_timed_wait(&t, &barrier);
+		TEST_EQUAL(ret,
+		           -ETIMEDOUT,
+		           "Barrier wait with no rendzvousing thread should time out");
```
rmn30:
```suggestion
		           "Barrier wait with no rendezvousing thread should time out");
```

rmn30 (commented):
LGTM. Just typos and comments.

rmn30 (approved):

PR #701, opened 2026-05-04
Explicitly drop .note.cheriot.compartment-name sections in the final ELF.

resistor:
These notes are per-compartment, so propagating them into the final ELF doesn't result in anything meaningful. This change is backwards compatible even if the compiler is not emitting .note.cheriot.compartment-name sections.


resistor:
See https://github.com/CHERIoT-Platform/llvm-project/pull/376

nwf (approved):
Seems sensible enough to me.

davidchisnall (approved):

PR #704, opened 2026-05-15
Add read-write locks.

davidchisnall:
These are the most tricky of the synchronisation primitives to get right.  Readers and writers need some ordering and you want to avoid too many spurious wakes.  This is made more complex by the fact that waiters can time out trying to acquire the read or write lock, so any *other* waiters behind them may suffer.

The approach here is described in detail in the comment in rwlock.cc. It requires the time-out path to potentially wake (and trigger spurious wakes in) all other waiters, but should make the common paths mostly sensible.


sdk/lib/locks/rwlock.cc
```diff
@@ -0,0 +1,438 @@
...
+	 * writer needs to be woken.
+	 *
+	 * When readers arrive, they can acquire the lock if both the reader count
+	 * is less than the maximum and there are no pending writers.  If this
+	 * happens, they increment the lock counter.
+	 *
+	 * If a reader cannot acquire the lock immediately, it sets the
+	 * readers-waiting flag and sleeps on the readers word.  When a reader
+	 * releases the lock, it checks to see whether it was the last reader and,
+	 * if so, and if there are writers waiting, it increments and notifies the
```
nwf, line 25: Given that readers can sleep if the maximum number of readers is hit, don't readers dropping the lock need to also signal (but not increment the writer notify word) if there are pending readers?

sdk/lib/locks/rwlock.cc
```diff
@@ -0,0 +1,438 @@
...
+	 * If a writer tries to acquire the lock, it checks whether there are zero
+	 * readers.  If so, it can acquire the lock directly.  If this fails, it
+	 * sets the writers-waiting flag and sleeps on the writer lock.  When
+	 * notified, it retries.
+	 *
+	 * All of this is moderately simple, except for timeouts.
+	 *
+	 * A *writer* timing out is the more complex case.  A writer may fail to
+	 * acquire the lock if either a reader (or multiple readers) or a writer
+	 * holds the lock.  If a reader holds the lock, other writers and readers
```
nwf:
```suggestion
	 * holds the lock.  If a writer holds the lock, other writers and readers
```

sdk/lib/locks/rwlock.cc
```diff
@@ -0,0 +1,438 @@
...
+		{
+			auto currentState = state.load(std::memory_order_relaxed);
+			while (true)
+			{
+				// Fast path.  If there are no current or pending writers, we
+				// can acquire the read lock immediately.
+				while (can_acquire_read_lock(currentState))
+				{
+					// The check in can_acquire_read_lock ensures that
+					// incrementing won't overflow into the higher bits.
```
nwf, line 166: But it might make the lock look write-locked?

sdk/lib/locks/rwlock.cc
```diff
@@ -0,0 +1,438 @@
...
+				if (is_write_locked(currentState) ||
+									reader_lock_count(currentState == 0))
+				{
+					Debug::log("Invalid lock to unlock as reader: {}", currentState);
+					return -EINVAL;
+				}
+				desired     = currentState - 1;
+				wakeReaders = false;
+				wakeWriters = false;
+				// If there are writers pending, we need to wake them.
```
nwf: Only if there are no readers still holding the lock?

sdk/lib/locks/rwlock.cc
```diff
@@ -0,0 +1,438 @@
...
+
+						// We have now dropped the reader-waiting flag.  Other
+						// readers may be waiting, so make sure that they wake
+						// up and reacquire the flag.
+						wake_readers();
+
+						// If there are writers waiting and the writer lock was
+						// dropped, we need to also wake writers
+						if (!is_write_locked(currentState) &&
+						    are_writers_pending(currentState))
```
nwf, line 218: Only if `reader_lock_count(currentState) != 0`, too?

sdk/lib/locks/rwlock.cc
davidchisnall: Hmm, I think the important thing for that case is that there are pending readers and *no* pending writers.  If there are pending writers, the next writer to wake will subsequently wake sleeping readers.

sdk/lib/locks/rwlock.cc
davidchisnall: No, unless there's a bug in `can_acquire_read_lock`.

sdk/lib/locks/rwlock.cc
davidchisnall: Actually, thinking about it, the time to wake readers is probably on the transition from the reader-count-saturated state to the reader-count-not-saturated state.

sdk/lib/locks/rwlock.cc
davidchisnall: Ah, no, I don't think so.  We need to defensively kick writers here because the ordering of our read and others waking up means that there is a gap when writers should wake but can't.

sdk/lib/locks/rwlock.cc
nwf: Assume no writers waiting. `can_acquire_read_lock(currentState)` compares `(reader_lock_count(currentState) < ReadersMask)`, and `reader_lock_count(currentState)` is just `currentState & ReadersMask` and `ReadersMask` includes the LSB of the state word.  `currentState + 1` can then equal `ReadersMask`, which is also the `WriteLocked` state.

I think the check in `can_acquire_read_lock` needs to compare against `ReadersMask - 1`?

sdk/lib/locks/rwlock.cc
```diff
@@ -0,0 +1,444 @@
...
+
+namespace
+{
+	/**
+	 * Reader-writer lock.  This uses two futex words:
+	 *
+	 * The first stores state, including the number of readers and whether there
+	 * are readers or writers pending.  The number of concurrent readers is
+	 * stored in this word.  When a writer holds the lock, it sets this to the
+	 * saturated value.
```
nwf, line 13: Is it an invariant that `reader_lock_count(s) == 0` implies both `!(s & ReaderWaitingFlag)` and `!(s & WriterWaitingFlag)`?

sdk/lib/locks/rwlock.cc
```diff
@@ -0,0 +1,444 @@
...
+		int unlock_reader()
+		{
+			auto     currentState = state.load(std::memory_order_relaxed);
+			uint32_t desired;
+			bool     wakeReaders;
+			bool     wakeWriters;
+			do
+			{
+				if (is_write_locked(currentState) ||
+				    reader_lock_count(currentState == 0))
```
jacobpake: Typo?

```suggestion
				    reader_lock_count(currentState) == 0)
```

sdk/lib/locks/rwlock.cc
```diff
@@ -0,0 +1,444 @@
...
+					desired &= ~WriterWaitingFlag;
+					wakeReaders = true;
+				}
+				// If there are writers blocked but not readers, we need to wake
+				// them.
+				else if (are_writers_pending(currentState))
+				{
+					desired &= ~ReaderWaitingFlag;
+					wakeWriters = true;
+				}
```
jacobpake, line 396: Are we clearing the right flags here?

sdk/lib/locks/rwlock.cc
davidchisnall: Yes, the check in `can_acquire_read_lock` was wrong.  See previous comment: EFQ.

sdk/lib/locks/rwlock.cc
davidchisnall: No, and also not above.

davidchisnall:
Not urgent to merge this, so I'm going to leave it for a few more review rounds - there are almost certainly still bugs.  Suggestions for more tests welcome!

sdk/lib/locks/rwlock.cc
```diff
@@ -0,0 +1,444 @@
...
+	 * readers.  If so, it can acquire the lock directly.  If this fails, it
+	 * sets the writers-waiting flag and sleeps on the writer lock.  When
+	 * notified, it retries.
+	 *
+	 * All of this is moderately simple, except for timeouts.
+	 *
+	 * A *writer* timing out is the more complex case.  A writer may fail to
+	 * acquire the lock if either a reader (or multiple readers) or a writer
+	 * holds the lock.  If a writer holds the lock, other writers and readers
+	 * may be blocked.  If a reader holds the lock, only other writers may be
```
marnovandermaas, line 38: Is the reason that you have the reader and writer role split just because multiple readers can then can hold the lock simultaneously?

sdk/lib/locks/rwlock.cc
```diff
@@ -0,0 +1,444 @@
...
+	 *    caller *may* be the only writer, but must assume that they are not
+	 *    (waiting writers are not reference counted).  On timeout, the caller
+	 *    must clear the writer-waiting flag and then wake writers.  Any other
+	 *    pending writers will reacquire the lock and, if they fail (which they
+	 *    almost certainly will), mark the lock as having waiters and sleep
+	 *    again.
+	 *  - Reader holds the lock.  Reader-waiting flag set.  At least one reader
+	 *    is blocked behind either us, or another higher-priority writer.
+	 *    Clear the writer-waiting flag, then signal the writer futex.  We are
+	 *    not trying to acquire the lock with the CAS at this point, so we may
```
marnovandermaas: Maybe useful to expand the acronym to compare and swap.

sdk/lib/locks/rwlock.cc
```diff
@@ -0,0 +1,444 @@
...
+	 * flag. If the writer has dropped the lock and there are pending writers,
+	 * it must *also* wake writers.  If the writer unlocked just before we
+	 * dropped the reader-waiting flag, then it's possible that we are the
+	 * *only* pending reader and the next writer is backed up behind us.
+	 */
+	class ReadWriteLock : public ReadWriteLockState
+	{
+		/// Flag in the state word indicating that at least one writer is
+		/// waiting
+		static const uint32_t WriterWaitingFlag = (1U << 31);
```
marnovandermaas, line 77: Is it really that important that you pack the reader count and the flags. I would find the code much easier to read if they were separate named fields in the struct.

tests/locks-test.cc
```diff
@@ -476,6 +476,67 @@ namespace
...
+		async([]() {
+			Timeout t{20};
+			otherThreadReturn = read_write_lock_acquire_as_writer(&t, &lock);
+			barrier_wait(&barrier);
+		});
+		// Try to make sure the background thread is blocked trying to release
+		// the lock.
+		sleep(2);
+		// Release the reader lock
+		expect(read_write_lock_release_as_reader, 0, "Releasing read lock (1)");
```
marnovandermaas, line 525: Maybe try another read acquire here to make sure the waiting writer is blocking that.

tests/locks-test.cc
```diff
@@ -476,6 +476,67 @@ namespace
...
+		expect(read_write_lock_release_as_reader, 0, "Releasing read lock (2)");
+		barrier_wait(&barrier);
+		TEST_EQUAL(otherThreadReturn, 0, "Releasing read locks woke writers");
+		expect(read_write_lock_release_as_reader,
+		       -EINVAL,
+		       "Releasing read lock that is write locked");
+		expect(read_write_lock_release_as_writer,
+		       0,
+		       "Releasing write lock that is write locked");
+		expect(read_write_lock_release_as_writer,
```
marnovandermaas, line 550: Check you can acquire the read lock again after the writer releases it.

tests/locks-test.cc
```diff
@@ -476,6 +476,67 @@ namespace
...
+		});
+		// Try to make sure the background thread is blocked trying to release
+		// the lock.
+		sleep(2);
+		// Release the reader lock
+		expect(read_write_lock_release_as_reader, 0, "Releasing read lock (1)");
+		expect(read_write_lock_release_as_reader, 0, "Releasing read lock (2)");
+		barrier_wait(&barrier);
+		TEST_EQUAL(otherThreadReturn, 0, "Releasing read locks woke writers");
+		expect(read_write_lock_release_as_reader,
```
marnovandermaas, line 539: Also try acquiring the lock as a writer when it is write locked.

sdk/lib/locks/rwlock.cc
nwf: Yes, precisely.  It allows for more concurrency, especially as reads dominate the workload.  (This is a classic synchronization primitive; the wikipedia page https://en.wikipedia.org/wiki/Readers%E2%80%93writer_lock is a reasonable place to start exploring its myriad variants, if you're curious.)

sdk/lib/locks/rwlock.cc
nwf: All but vital.  They must be part of the material subject to atomic compare-and-swap and compare-and-sleep (AKA futex wait) operations, and bit-packing is probably the easiest way to do that.

PR #705, opened 2026-05-16
sonata: fix PcIntModeBits definition in SonataGpioBase

hmelder:
This PR fixes an incorrect definition of the pin-change interrupt mode bitmask.


nwf:
Please feel free to suppress clang-tidy's warning with a pragma.  Assuming #558 lands, we should convert this file, too, but later.

davidchisnall (approved):

davidchisnall:
Looks like the NOLINT thing may need to be on the line above?

PR #706, opened 2026-05-18
Add kudu simulator to CI matrix.

rmn30:


rmn30:
I didn't add kudu for running the examples. I doubt it would add much value.

davidchisnall (approved):

rmn30:
This will have to wait until we have a dev container with a fix for https://github.com/microsoft/cheriot-kudu/issues/2.

PR #708, opened 2026-05-19
Initial work on generalising timeouts.

### LINKED ISSUE #703: Timeouts are not sufficient for some use cases
The timeout mechanism in CHERIoT RTOS was originally created to be a slightly nicer affordance on the FreeRTOS model.  In this model, you specify a number of ticks that may be used for blocking.  The `Timeout` structure contains the number of ticks you may wait for and the number that have elapsed.

#667 highlighted that there are some cases where you want to sleep until some deadline has passed.  Specifying this as a relative timeout is racy.  Unless you are the highest-priority thread, you look at the current time, compute the time until your next deadline as a relative time, are preempted, then sleep for longer than you expected.

The move to a tickless scheduler also means that the tick abstraction means that the tick model is purely an API property.  Internally, the scheduler uses 64-bit counter times and sets the timer interrupt to the next pending timeout.  Sleeping threads are stored in a sleep queue sorted by timeout.  This means that, internally, changing timeouts to be expressed in some other format is easy.  That said, the tick abstraction is useful for providing some rough time portability across different-speed clocks (one tick is 'the amount of time it takes to do a useful amount of work on this core', so waiting for two ticks is 'wait a little bit').

We have three use cases:

 - FreeRTOS-style wait for *n* ticks.
 - Wait until some deadline.
 - Wait for *n* cycles (note: this might be unnecessary because we don't guarantee any precision more than ticks for sleeps and so waiting for *n* cycles and *n* are equivalent).

There's also the possible use case for POSIX threading compatibility: timeouts relative to the wall clock instead of the monotonic clock.  This is a staggeringly bad idea, because wall clock time can move in arbitrary amounts (forward or backwards) in the background, and we probably should make things that want to wait on the wall-clock time do something like calculate the time relative to the monotonic clock, then wait for that timeout or one second, whichever is smaller, then recheck.

C11 mutexes suffer from the same API bug but doesn't have the excuse that it wasn't a known defect at the time of their standard.

We currently use a 32-bit counter for timeouts because ticks are typically on the order of 1-10 ms and so a 32-bit timeout would typically allow timeouts of at least a month and usually a year or more.  Waking up spuriously once a month and retrying is not likely to impact battery life for performance for anything.  For a 250 MHz core, counting cycles, a 32-bit timeout is only around 16 seconds.  Absolute timeouts would definitely need to be 64 bits, since the absolute time will overflow even a shifted 32-bit value quite quickly.

For absolute timeouts, we don't need the `elapsed` field of the current timeout: once the deadline is passed, callers can query the new monotonic time to see how far after the timeout we are.  

There are a few possibilities for how to implement this in a mostly (source) backwards compatible way.

Approach 1:  Allow timeouts to be specified as `Timeout *` or `uint64_t`
-----------------------------------------------------------

We could express absolute timeouts as an untagged 64-bit integer.  Things taking a timeout would take a union of either a timeout pointer or a 64-bit integer (which is passed in a capability register and differentiated by the tag bit).  If it's a timeout, the semantics are as today.  If the value is untagged, it is treated as an absolute timeout.

The benefits of this approach are:

 - No increase in timeout size
 - Can pass 0 to a timeout to mean 'don't block', which would simplify some existing code.
 - Code that reads the fields directly would need changing.

The downsides are:

 - Adding relative cycle-based timeouts would require more changes in the future (though there are a few top bits that we could steal from the 64-bit integer).
 - There's no way of having cycle-based relative timeouts that have an elapsed field.
 - C/C++ support for tagged unions using the CHERI tag bit is not great, for C++ we could probably hide a lot of the complexity in a `TimeoutParameter` class that held a `Timeout*` or `uint64_t`.

The network stack uses `t->remaining > 0` to check for timeout in a few places.  This would need replacing with a 'timeout_has_expired` helper function.

The allocator has a pattern (which may show up elsewhere) where it sleeps for one tick and elapses the time against the provided timeout.  This involves explicitly reading the `->remaining` field of the argument, so would need refactoring.  

Approach 2: Add a discriminator to the `Timeout`
--------------------------------------------------

We could redefine `struct Timeout` as something like:

```
struct Timeout
{
	union
	{
		uint64_t absolute;
		struct
		{
			uint32_t remaining;
			uint32_t elapsed;
		};
	};
	enum TimeoutKind : uint8_t
	{
		TimeoutKindTicksRelative = 0,
		TimeoutKindCyclesAbsolute,
	} kind;
};
```

The order of fields in the union means that the existing C initialiser of `{0, 0}` or `{[timeout], 0}` works, but any existing code that touches the fields directly will need changing.

The benefits of this approach are:

 - Extensible in the future (e.g. to support other clock sources).
 - Simple to understand.
 - Could be made binary backwards compatible by checking the length of the capability used to pass it (probably not worth the effort).

The downsides are:

 - Larger structure.  Given stack alignment requirements, this may add as much as 16 bytes more to a stack frame that passes a timeout down, even with the one-byte discriminator.
 - Is *almost* backwards compatible, but isn't in some cases.

I am slightly leaning towards approach 1, but I'm not sure.  Other perspectives welcome!

davidchisnall:
This extends the core (scheduler, locking, allocator) APIs to support either an absolute timeout relative to the platform's monotonic clock in addition to the FreeRTOS-inspired relative-tick-based waiting.

This is designed to be compatible with old APIs.  Deliberately, this PR *doesn't* update *all* of the APIs, so that we have callers that are still taking `Timeout*` and forward it.

Fixes #703

### DIFF
--- a/sdk/core/scheduler/thread.h
+++ b/sdk/core/scheduler/thread.h
@@ -158,28 +158,22 @@ namespace
 		 *
 		 * Returns true on timeout, false otherwise.
 		 */
-		bool suspend(Timeout     *t,
-		             ThreadImpl **newSleepQueue,
-		             bool         yieldUnconditionally = false,
-		             bool         yieldNotSleep        = false)
+		bool suspend(TimeoutArgument t,
+		             ThreadImpl    **newSleepQueue,
+		             bool            yieldUnconditionally = false,
+		             bool            yieldNotSleep        = false)
 		{
-			if (t->remaining != 0)
+			bool mayBlock = t.may_block();
+			if (mayBlock)
 			{
-				suspend(t->remaining, newSleepQueue, yieldNotSleep);
+				suspend_internal(t, newSleepQueue, yieldNotSleep);
 			}
-			if ((t->remaining != 0) || yieldUnconditionally)
+			if (yieldUnconditionally || mayBlock)
 			{
 				auto elapsed = yield_timed();
-				if (CHERI::Capability{t}.is_valid())
-				{
-					t->elapse(elapsed);
-					if (t->remaining > 0)
-					{
-						return false;
-					}
-				}
+				t.elapse(elapsed);
 			}
-			return true;
+			return !t.may_block();
 		}
 
 		/**
@@ -216,8 +210,6 @@ namespace
 		 */
 		void ready(WakeReason reason)
 		{
-			int64_t ticksLeft;
-
 			// We must be suspended.
 			Debug::Assert(state == ThreadState::Suspended,
 			              "Waking thread that is in state {}, not suspended",
@@ -278,32 +270,6 @@ namespace
 				visitor(thread);
 				thread = next;
 			}
-		}
-
-		/**
-		 * Suspend this thread. Take it off the ready list. If it is suspended
-		 * waiting on a resource, add it to the list of that resource. No
-		 * matter what, it has to be added to the timer list.
-		 */
-		void suspend(uint32_t     waitTicks,
-		             ThreadImpl **newSleepQueue,
-		             bool         yieldNotSleep = false)
-		{
-			isYielding = yieldNotSleep;
-			Debug::Assert(state == ThreadState::Ready,
-			              "Suspending thread that is in state {}, not ready",
-			              static_cast<ThreadState>(state));
-			list_remove(&priorityList[priority]);
-			state = ThreadState::Suspended;
-			priority_map_remove();
-			if (newSleepQueue != nullptr)
-			{
-				list_insert(newSleepQueue);
-				sleepQueue = newSleepQueue;
-			}
-			expiryTime = expiry_time_for_timeout(waitTicks);
-
-			timer_list_insert(&waitingList);
 		}
 
 		/**
@@ -624,6 +590,41 @@ namespace
 		CHERI_SEALED(TrustedStack *) tStackPtr;
 
 		private:
+		/**
+		 * Suspend this thread. Take it off the ready list. If it is suspended
+		 * waiting on a resource, add it to the list of that resource. No
+		 * matter what, it has to be added to the timer list.
+		 */
+		void suspend_internal(TimeoutArgument timeout,
+		                      ThreadImpl    **newSleepQueue,
+		                      bool            yieldNotSleep = false)
+		{
+			isYielding = yieldNotSleep;
+			Debug::Assert(state == ThreadState::Ready,
+			              "Suspending thread that is in state {}, not ready",
+			              static_cast<ThreadState>(state));
+			list_remove(&priorityList[priority]);
+			state = ThreadState::Suspended;
+			priority_map_remove();
+			if (newSleepQueue != nullptr)
+			{
+				list_insert(newSleepQueue);
+				sleepQueue = newSleepQueue;
+			}
+			if (timeout.is_relative())
+			{
+				expiryTime =
+				  expiry_time_for_timeout(timeout.relativeTimeout->remaining);
+			}
+			else
+			{
+				expiryTime = timeout.absoluteTimeout;
+			}
+			Debug::log("Sleeping until {}", expiryTime);
+
+			timer_list_insert(&waitingList);
+		}
+
 		/**
 		 * Helper to remove a thread from the priority map and update the
 		 * highest priority, if it was the last runnable thread at that
--- a/sdk/include/platform/generic-riscv/platform-time.h
+++ b/sdk/include/platform/generic-riscv/platform-time.h
@@ -1,0 +1,71 @@
+// Copyright SCI Semiconductor and CHERIoT Contributors.
+// SPDX-License-Identifier: MIT
+#pragma once
+
+#include <cdefs.h>
+#include <stdint.h>
+
+/**
+ * @file `platform-time.h` defines the interface for the timer that is used to
+ * implement a monotonic clock.  This is the generic implementation that
+ * assumes that the time value is exposed via a CSR.  If your board exposes
+ * this via an MMIO region then it must provide a different implementation.
+ */
+
+/**
+ * `CHERIOT_PLATFORM_TIME_HIGH` is a macro that can be defined by platform
+ * headers that include this via `#include_next`.  This defines the name for
+ * the CSR that provides the top 32 bits of a 64-bit counter.
+ */
+#ifndef CHERIOT_PLATFORM_TIME_HIGH
+#	define CHERIOT_PLATFORM_TIME_HIGH mcycleh
+#endif
+
+/**
+ * `CHERIOT_PLATFORM_TIME_LOW` is a macro that can be defined by platform
+ * headers that include this via `#include_next`.  This defines the name for
+ * the CSR that provides the top 32 bits of a 64-bit counter.
+ */
+#ifndef CHERIOT_PLATFORM_TIME_LOW
+#	define CHERIOT_PLATFORM_TIME_LOW mcycle
+#endif
+
+/**
+ * Platform-specific hook for reading the current monotonic time.  This should
+ * be implemented to return the time that increments with the rate defined by
+ * the `CPU_TIMER_HZ` macro and must be the same timer used by the scheduler.
+ */
+__if_c(static) inline uint64_t platform_monotonic_time_read()
+{
+	uint32_t mtimehBefore;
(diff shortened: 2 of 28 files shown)


sdk/include/timeout.h
```diff
@@ -86,3 +101,162 @@ typedef struct Timeout
...
+}
+
+/**
+ * Returns true if this is a valid `TimeoutArgument`: either an untagged value
+ * or a valid pointer to a `Timeout`.
+ */
+bool __cheriot_libcall timeout_is_valid(TimeoutArgument timeout);
+
+/**
+ * Returns true if the timeout is in the past, false otherwise.
```
rmn30: A relative timeout cannot be "in the past" so you need to describe what this does for relative timeouts.

davidchisnall: Yes it can.  If it has zero remaining ticks, it is in the past.

rmn30: I understand but I think it is clearer to be explicit.

sdk/include/platform/generic-riscv/platform-time.h
```diff
@@ -0,0 +1,32 @@
...
+	__asm__ volatile("csrr %0, " __XSTRING(CHERIOT_PLATFORM_TIME_LOW)
+	                 : "=r"(mtime));
+	uint32_t mtimehAfter;
+	__asm__ volatile("csrr %0, " __XSTRING(CHERIOT_PLATFORM_TIME_HIGH)
+	                 : "=r"(mtimehAfter));
+	uint32_t correction = (mtimehAfter - mtimehBefore) & (mtime >> 31);
+	// This header is included from C and C++, don't do C++-specific lints in
+	// it. NOLINTNEXTLINE(google-readability-casting)
+	return ((uint64_t)(mtimehAfter - correction) << 32) + mtime;
+}
```
rmn30, line 71: Two issues here:

1) This is only correct on platforms where `mtime == mcycle`. I'm not even sure that is true on Sail so we need to check.
2) This is not the canonical overflow safe way to read the 64-bit counter and I think a comment is warranted. I wasn't sure how it would know whether the counter overflowed before or after the read of the low bits but I think it relies on the high bit not being set if it has just overflowed? For reference the suggested sequence form the [RISC-V spec](https://docs.riscv.org/reference/isa/unpriv/counters.html#counters) is:

```
    again:
        rdcycleh     x3
        rdcycle      x2
        rdcycleh     x4
        bne          x3, x4, again
```

davidchisnall: > This is only correct on platforms where mtime == mcycle. I'm not even sure that is true on Sail so we need to check.

This is in the platform layer precisely to permit other implementations.

> This is not the canonical overflow safe way to read the 64-bit counter and I think a comment is warranted.

Really?  Because I copied this sequence from something that claimed it was the canonical way to do a branch-free read of cycle counters on RISC-V...

rmn30: I guess it's better to avoid a branch and I think it's correct as written.

nwf: Is `platform_monotonic_time_read` preemptable?  (In particular, `correction` is going to be either `0` or `1` since `mtime` is `uint32_t` and so `mtime >> 31` is also either `0` or `1`, but if, say, `mtimehAfter - mtimehBefore` is `2`, then we will be returning `(mtimehAfter << 32) + mtime`, which might be in the future, so the next read might not be a monotonic increase?)

sdk/include/timeout.h
```diff
@@ -86,3 +101,163 @@ typedef struct Timeout
...
+/**
+ * A union to hold timeouts that are passed down to end up in the scheduler.
+ * These are one of two things:
+ *
+ *  - An absolute timeout, measured in cycles of some platform-specific timer
+ *    (such as the `mtime` register) that increments at predictable rate.
+ *  - A relative timeout counted in elapsed scheduler ticks.
+ *
+ * This is a tagged union, with the CHERI tag bit used to differentiate between
+ * the two cases.
```
nwf, line 114: I saw there was some discussion in the Signal group around tags going missing.  The consequence is, AIUI, that the address and metadata, or some mangled form of it, gets interpreted as an absolute timeout, probably in the far future.  While unlikely that `__TimeoutArgument`-s get mangled like this, all it'd take is storing a cap to an on-stack relative `Timeout` out to non-stack memory and reloading it.

sdk/core/scheduler/thread.h
```diff
@@ -624,6 +590,41 @@ namespace
 		CHERI_SEALED(TrustedStack *) tStackPtr;
 
 		private:
+		/**
```
nwf, line 593: Just a nit, why did this move in the file?

tests/allocator-test.cc
```diff
@@ -1557,10 +1552,16 @@ int test_allocator()
 	    .as_error(),
 	  -EINVAL,
 	  "Allocating array with size overflow succeeded");
-	TEST_EQUAL(ErrorOr{heap_allocate_array(nullptr, MALLOC_CAPABILITY, 64, 2)}
-	             .as_error(),
-	           -EINVAL,
-	           "Allocating array with null timeout succeeded");
+	Timeout *invalidTimeout = static_cast<Timeout *>(
```
nwf, line 1555: Hmmm... this makes me realize that `nullptr` is an alias of `TimeoutNoWait`...  That's maybe unfortunate?

nwf (commented):
Seems fine, modulo some nits and the bigger question of using the CHERI tag like this.

sdk/include/timeout.h
davidchisnall: Timeouts *must be* on the stack.  We have an explicit check for that.

If you store a timeout to the stack and reload it, you will trap in existing code, so this is just a different failure mode.

If you pass the callee a pointer that overlaps the timeout pointer then you can probably make it destroy the timeout, but then it just blocks longer (or less long, depending on what it overwrites it with) and that's no different than your just passing it a different timeout.

sdk/include/platform/generic-riscv/platform-time.h
davidchisnall: Hmm, yes, I think that is an issue.  It's possible to be preempted for more than 2^32 timer ticks, which would cause that problem.

davidchisnall: > This is only correct on platforms where mtime == mcycle. I'm not even sure that is true on Sail so we need to check.

Oh, and this is only true if you don't read the entire file.  Look at lines 5-10.

sdk/include/timeout.h
nwf: I agree it's just a different failure mode, but all the same... should we have the scheduler require that a nominally integral/absolute timeout has an all-zero capability metadata half?  (That is, add basically `cgethigh; bnez`?)  Deliberately constructed integral values will pass this check, while capabilities that have lost their tag will, with high probability, not.

davidchisnall: So, there are two things:

1. We're passing the timeout pointer by value.  This means that only the compartment that receives it can accidentally overwrite it.  That, in turn, can happen only if you have a memory-safety bug that CHERI doesn't catch.  And, in that case, you're *so far* into UB that nothing we do will help it.
2. Related, the fact that timeout pointers need to be on the stack means that you can't clear the tag on the pointer as the effect of action at a distance unless you race in a really exciting way and allocate the timeout on the heap, then racily free it *before* the first `check_timeout` call.  And that's a really, really complex way of doing something that you could do by just passing an integer.

> should we have the scheduler require that a nominally integral/absolute timeout has an all-zero capability metadata half?

Only if we want all absolute timeouts to fail after about 20 seconds of uptime.  There's a reason that this is a 64-bit value.

tests/allocator-test.cc
davidchisnall: No, that's a feature.  Passing null as a timeout means don't wait.

sdk/core/scheduler/thread.h
davidchisnall: It's now a purely internal API and so is with the other private things.

sdk/include/platform/generic-riscv/platform-time.h
nwf: Should we fix this before merging?  Is the right fix just to use a loop or should we have an outlined irq-disabled helper?

nwf (approved):
LGTM

PR #709, opened 2026-05-20
allocator: refactor revoker interface

nwf:
Fuse the revoker and revocation region data structure ("shadow bitmap") into a single object; they're quite intertwined in practice and this allows for more exotic implementations than a literal bitmap and engine that applies that bitmap.

Some things move around: the `concept` of a revoker is generalized and moved out of sdk/include/platforms to sdk/include/allocator/revoker. The "direct bitmap" implementation of the revoke moves from being allocator-internal to the same place so that it can be used by platform revoker implementations.


davidchisnall (commented):
Looks good, but what is the rational for moving the concept out of platform/concepts?

nwf:
Nothing very strong; was just collecting revoker things into `sdk/include/revoker`, but thinking about it more, it probably makes sense for the hardware contract to stay where it was.

davidchisnall (approved):

PR #710, opened 2026-05-20
WIP: Doxygen

nwf:
I have been very slowly experimenting with getting Doxygen builds for the RTOS headers (notably, #558 tries to have sensible Doxygen output for `bitpack.hh`).

So far the result is a mess, despite our habitual use of doxygen-style comments... but it's a start.

I've been having Doxygen use a CHERIoT LLVM `libclang` to parse source; I'll go raise a PR on the devcontainer to build such a thing for general use.


Doxyfile
```diff
@@ -0,0 +1,3052 @@
...
+# cores available in the system. You can set it explicitly to a value larger
+# than 0 to get more control over the balance between CPU load and processing
+# speed. At this moment only the input processing can be done using multiple
+# threads. Since this is still an experimental feature the default is set to 1,
+# which effectively disables parallel processing. Please report any issues you
+# encounter. Generating dot graphs in parallel is controlled by the
+# DOT_NUM_THREADS setting.
+# Minimum value: 0, maximum value: 512, default value: 1.
+
+NUM_PROC_THREADS       = 1
```
davidchisnall: Should this be 0, or are there problems with threading in Doxygen?

Doxyfile
```diff
@@ -0,0 +1,3052 @@
...
+RECURSIVE              = YES
+
+# The EXCLUDE tag can be used to specify files and/or directories that should be
+# excluded from the INPUT source files. This way you can easily exclude a
+# subdirectory from a directory tree whose root is specified with the INPUT tag.
+#
+# Note that relative paths are relative to the directory from which Doxygen is
+# run.
+
+EXCLUDE                = sdk/include/c++-config sdk/include/libc++
```
davidchisnall: Mildly annoyingly, we have some non-standard extensions in `std::atomic` in `sdk/include/c++-config/atomic`, which would be nice to include.  I don't know if there's a good way to do that.

Doxyfile
nwf: Huh.  I didn't even notice and had just left this as default.

davidchisnall (approved):

.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,67 @@
+name: Build / deploy doxygen to github pages
+
+on:
+  # Build runs on pushes to any branch but deploy job runs only on main
+  push:
+    branches: ["*"]
```
davidchisnall: This probably should be branches: main, and pr:

```suggestion
  push:
    branches: [ main ]
  pull_request:
    branches: [ main ]
```

.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,67 @@
...
+      #   run: |
+      #       sudo apt-get update
+      #       sudo apt install graphviz cmake ninja-build
+      # Use a recent commit of doxygen from main that seems to work nicely.
+      # Released versions lack features or have bugs :-(
+      - name: Build doxygen
+        run: |
+          sudo apt-get update
+          sudo apt install graphviz cmake ninja-build bison flex
+          git clone https://github.com/doxygen/doxygen
```
nwf: You can cut down on network traffic and speed this up a bit with
```
mkdir doxygen
cd doxygen
git init
git remote add origin https://github.com/doxygen/doxygen
git fetch --depth 1 origin 744dfaeab267a00866a1fe065265f580f3060fac
git checkout FETCH_HEAD
```

(In sufficiently new versions of git, this is just `git clone --depth 1 --revision 744dfaeab267a00866a1fe065265f580f3060fac https://github.com/doxygen/doxygen` .)

rmn30: Unfortunately the version in the runner doesn't seem to be sufficiently new.

PR #714, opened 2026-05-26
Move guard.cc to using std::atomic.

davidchisnall:
This avoids the entire functions running with interrupts disabled and means that cores that implement atomics can avoid disabling interrupts entirely.

### DIFF
--- a/sdk/lib/cxxrt/guard.cc
+++ b/sdk/lib/cxxrt/guard.cc
@@ -1,10 +1,10 @@
 // Copyright Microsoft and CHERIoT Contributors.
 // SPDX-License-Identifier: MIT
 
+#include <atomic>
 #include <cdefs.h>
 #include <debug.hh>
-#include <futex.h>
-#include <limits>
+#include <locks.hh>
 #include <stdint.h>
 
 using Debug = ConditionalDebug<DEBUG_CXXRT, "cxxrt">;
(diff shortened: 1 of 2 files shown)


sdk/lib/cxxrt/guard.cc
nwf: This comment is stale now, yes?

sdk/lib/cxxrt/guard.cc
```diff
@@ -33,9 +31,9 @@ namespace
 	class GuardWord
 	{
 		/// The low half (first on a little-endian system).
-		uint32_t low;
+		std::atomic<uint32_t> low;
 		/// The high half (second on a little-endian system).
-		uint32_t high;
+		std::atomic<uint32_t> high;
```
nwf: Does this still need to be two words?

davidchisnall: Yes, it’s part of the ABI (an artefact of HP using long long when they meant `uintptr_t`, but hard to change).

@resistor, the Arm ABI uses a single 32-bit word. That would reduce code size. Ideally we’d move the flag bit to the top bit so we could use the low half as a PI futex.

I was thinking last night we should probably use the top word as a PI mutex anyway, so I’ll make that change before we merge.

davidchisnall:
Updated to use priority-inheriting locks.

nwf (approved):
Seems sensible to me.

PR #716, opened 2026-06-10
locks: FlagLock lock() docs & Invariant

nwf:
FlagLockGeneric<>::lock() invokes try_lock() (which wraps one of the lock library trylock() operations) with indefinite timeout and expects termination to imply success and has no way to indicate failure.

Unfortunately, these trylock()s can return after failure if the lock is PI and already held by the current thread (futex_wait(), and so std::atomic<>::wait, will fail), the lock has been "upgraded for destruction". or if the lock word has been corrupted.  Notably, the first case can happen if an attempt is made to recursively lock the FlagLock or the current thread was previously forcibly unwound out of a critical section with the lock held and is now attempting to reacquire it.

Document that lock() should not be used on locks that can be "upgraded for destruction" and add a Debug::Invariant that try_lock has indeed succeeded.

### DIFF
--- a/sdk/include/locks.hh
+++ b/sdk/include/locks.hh
@@ -39,31 +39,50 @@ __clang_ignored_warning_push("-Watomic-alignment");
  * requires mutual exclusion in the presence of mutual distrust should
  * consider an using a lock manager compartment with an API that returns a
  * single-use capability to unlock on any lock call.
+ *
+ * Flag locks are not recursive; if needed, use a `RecursiveMutex` instead.
  */
 template<bool IsPriorityInherited>
 class FlagLockGeneric
 {
 	FlagLockState state;
 
+	/**
+	 * Attempt to acquire the lock, blocking until a timeout specified by the
+	 * `timeout` parameter has expired.
+	 *
+	 * Returns an errno value.
+	 */
+	__always_inline int try_lock_internal(Timeout *timeout)
+	{
+		if constexpr (IsPriorityInherited)
+		{
+			return flaglock_priority_inheriting_trylock(timeout, &state);
+		}
+		else
+		{
+			return flaglock_trylock(timeout, &state);
+		}
+	}
+
 	public:
 	/**
 	 * Attempt to acquire the lock, blocking until a timeout specified by the
 	 * `timeout` parameter has expired.
+	 *
+	 * Returns `true` if and only if the lock transitioned from being unheld to
+	 * held by the current thread.
 	 */
 	__always_inline bool try_lock(Timeout *timeout)
 	{
-		if constexpr (IsPriorityInherited)
-		{
-			return flaglock_priority_inheriting_trylock(timeout, &state) == 0;
-		}
-		else
-		{
-			return flaglock_trylock(timeout, &state) == 0;
-		}
+		return try_lock_internal(timeout) == 0;
 	}
 
 	/**
 	 * Try to acquire the lock, do not block.
+	 *
+	 * Returns `true` if and only if the lock transitioned from being unheld to
+	 * held by the current thread.
 	 */
 	__always_inline bool try_lock()
 	{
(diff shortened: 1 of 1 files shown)


sdk/include/locks.hh
```diff
@@ -39,29 +39,44 @@ __clang_ignored_warning_push("-Watomic-alignment");
...
 
-	public:
 	/**
 	 * Attempt to acquire the lock, blocking until a timeout specified by the
 	 * `timeout` parameter has expired.
+	 *
+	 * Returns an errno value.
 	 */
-	__always_inline bool try_lock(Timeout *timeout)
+	__always_inline int try_lock_internal(Timeout *timeout)
```
rmn30, line 56: Should we leave it up to the compiler whether to inline?

sdk/include/locks.hh
```diff
@@ -73,11 +88,15 @@ class FlagLockGeneric
 
 	/**
 	 * Acquire the lock, potentially blocking forever.
+	 *
+	 * It is not safe to use this function on locks that may be put into
+	 * "destrucion mode" (see `flaglock_upgrade_for_destruction`).
```
rmn30: I think you should expand on why it is not safe. It's because the call could fail to obtain the lock and there's no way for the caller to know unless they check explicitly? It could also fail because of a failed compartment call to the scheduler due to stack overflow.

Are there callers of this that should have this warning propagated e.g. LockGuard?

sdk/include/locks.hh
nwf: I sort of assume the compiler would always agree: this is just a function call and register shuffling (copying `this` to `a1`).

rmn30: That's true. Didn't notice that the if was constexpt and was writing about code size.

rmn30 (approved):

rmn30:
I guess this is likely to conflict with #708 ?

sdk/include/locks.hh
```diff
@@ -39,31 +39,48 @@ __clang_ignored_warning_push("-Watomic-alignment");
...
+	 */
+	__always_inline bool try_lock(Timeout *timeout)
+	{
+		return try_lock_internal(timeout) == 0;
+	}
+
 	/**
 	 * Try to acquire the lock, do not block.
+	 *
+	 * Returns `true` iff the lock was not previously held and now is.
```
davidchisnall: As above.

sdk/include/locks.hh
```diff
@@ -39,31 +39,48 @@ __clang_ignored_warning_push("-Watomic-alignment");
...
+			return flaglock_trylock(timeout, &state);
 		}
 	}
 
+	public:
+	/**
+	 * Attempt to acquire the lock, blocking until a timeout specified by the
+	 * `timeout` parameter has expired.
+	 *
+	 * Returns `true` iff the lock was not previously held and now is.
```
davidchisnall: Please can we spell out 'if (and only if)': most people from a more applied software-engineering background get confused when I say 'iff'.

sdk/include/locks.hh
```diff
@@ -73,11 +90,22 @@ class FlagLockGeneric
...
+	 *
+	 * Because this function has no way to signal failure to acquire the lock
+	 * (other than by not returning, by tripping an Invariant),
+	 * it is not safe to use on locks that may be put into "destrucion mode"
+	 * (see `flaglock_upgrade_for_destruction`).
+	 * Attempting to recursively `lock` a held priority inheriting flag lock
+	 * (acquired by either `lock` or `try_lock`) will trip the same Invariant.
+	 * (Recall that flag locks explicitly do not support recursive acquisition;
+	 * the failure mode is, however, not necessarily one of non-termination!)
+	 * In general, defensive code should prefer `try_lock`.
```
davidchisnall: It would be nice if this was explicit that the API returns `void`, even though that looks like a bad idea, because the C++ standard library has locks that have `::lock()` methods and return `void`, and so callers in templated code will not check the return value, making it unsafe to add one.

sdk/include/locks.hh
```diff
@@ -73,11 +90,22 @@ class FlagLockGeneric
...
 	 */
 	__always_inline void lock()
 	{
 		Timeout t{UnlimitedTimeout};
```
davidchisnall, line 112: I think this will conflict with the timeout change, which we should land Real Soon Now™, because there's a newer and more efficient spelling of this.

sdk/include/locks.hh
```diff
@@ -269,14 +297,28 @@ class LockGuard
 	bool isOwned;
 
 	public:
-	/// Constructor, acquires the lock.
+	/**
+	 * Constructor, acquires the lock with unbounded wait.
+	 *
+	 * The lock to be wrapped must not be locked prior to calling.
```
davidchisnall: Must not be locked by the caller?  Or maybe must be able to be locked?  It's fine to use this for a lock that is locked by someone else, that's kind-of the point.

rmn30 (approved):
It's fine but what are we going to do about the conflicts with #708 ?

nwf:
> It's fine but what are we going to do about the conflicts with #708 ?

Pick one to move first and rebase the other?

nwf:
I have #708 rebased atop this locally.  Can we land this one, assuming CI is happy, and I'll push to that branch?

PR #717, opened 2026-06-16
Provide prototypes for runtime libcalls in header files.

resistor:
We want to have the toolchain emit warnings for functions annotated as libcalls that do not have a separate prototype provided in a header. This causes issues for CRT libcalls provided by the RTOS, where no header declaration is normally present. The simple fix is to add internal-only headers to provide such declarations and silence the warning.

### DIFF
--- a/sdk/lib/crt/arith64.c
+++ b/sdk/lib/crt/arith64.c
@@ -29,41 +29,11 @@
 // - Provide real implementations for __clzsi2() / __clzdi2() / __popcount* only when the RISCV Zbb extension is not present
 // - Use __builtin_clzg rather than calling __clzdi2 from other functions
 // - Replace clz*/ctz*/popcount* implementations with calls to clang builtins: they have slightly smaller code size without B, or will use B instructions if enabled and present
+// - Extracted declarations to arith64.h
 //
 // This file is *not* formatted, to make it easier to track changes from upstream (if there are any).
 
-#include "cdefs.h"
-#include <stdint.h>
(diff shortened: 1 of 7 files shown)


resistor:
See https://github.com/CHERIoT-Platform/llvm-project/pull/411

sdk/lib/crt/arith64.c
```diff
@@ -22,48 +22,24 @@
...
 // - Use __builtin_clzg rather than calling __clzdi2 from other functions
-// - Replace clz*/ctz*/popcount* implementations with calls to clang builtins: they have slightly smaller code size without B, or will use B instructions if enabled and present
+// - Replace clz*/ctz*/popcount* implementations with calls to clang builtins:
+// they have slightly smaller code size without B, or will use B instructions if
+// enabled and present
+// - Extracted declarations to arith64.h
 //
-// This file is *not* formatted, to make it easier to track changes from upstream (if there are any).
+// This file is *not* formatted, to make it easier to track changes from
+// upstream (if there are any).
```
rmn30: It looks like this got formatted despite the comment saying it should not be?

rmn30 (approved):
I think it's OK. Looks straightforward enough. I guess it makes it harder to keep track of changes in aritht64.c but there might be not be much we can do about that.

sdk/lib/crt/arith64.c
resistor: Fixed

PR #718, opened 2026-06-22
Fix allocator initialization arithmetics

hlef:
This fixes/improves a few things on the allocator initialization path:

- Don't add the size of `MChunkHeader` to the size of `MState` when calculating msize. This space is wasted, I suspect that the mistake comes either from re-using `pad_request` (which adds the `MChunkHeader` size), or from incorrectly assuming that `MinChunkSize` does not include `MChunkHeader`.

- The check for the minimal heap size does not include the hazard quarantine. This would let us initialize a heap that's too small.

- The documentation of `mspace_firstchunk_add` is outdated. I removed the `@brief` tags, it doesn't seem to be a standard practice (anymore) throughout the RTOS. Please let me know if you want it back.

- Add a comment to explain why we don't need to pad again after the hazard quarantine.

(This deserves a really good look by another pair of eyes to make sure I didn't break it more than it was!)

### DIFF
--- a/sdk/core/allocator/alloc.h
+++ b/sdk/core/allocator/alloc.h
@@ -1061,13 +1061,10 @@ class MState
 	}
 
 	/**
-	 * @brief Called when adding the first memory chunk to this MState.
-	 * At the moment we only support one contiguous MChunk for each MState.
+	 * Called when adding the first memory chunk to this MState. At the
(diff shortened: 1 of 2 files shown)


nwf (approved):
LGTM.  Thanks so much!

PR #723, opened 2026-06-30
queue_compartment.cc: removed unused helper has_permissions

nding0405:
Removed the unused `has_permissions` helper from `queue_compartment.cc`.
Permission checking is handled by `unseal(CHERI_SEALED(MessageQueue *) handle, int permissions)` now, so this removes dead code without changing behavior.
Testing:
- Ran the test suite successfully after the removal.


nwf:
This seems sensible to me, but I'm going to hold off merging until David has a chance to look.

davidchisnall (approved):

davidchisnall:
Yup, this was dead code left over from before a refactoring.

Issue #492, opened 2025-03-25
benchmarks are not clang-{format,tidy}-ed


nwf:
We (future me, really) should fix that, probably after landing #479.

PR #724, opened 2026-06-30
benchmarks: enable clang-format and clang-tidy

nding0405:
Added `benchmarks/*` to `generate_compile_commands.sh` and `run_clang_tidy_format.sh` so benchmark files are included in the clang-format and clang-tidy checks.
Fixed the resulting style issues in the benchmark files and confirmed that the benchmark files pass the checks.

### DIFF
--- a/benchmarks/compartment-call/callee.cc
+++ b/benchmarks/compartment-call/callee.cc
@@ -15,7 +15,7 @@ int callee_stack_using_return_metric(size_t stackUse)
 	return METRIC();
 }
 
-int callee_dereference(int *p)
+int callee_dereference(const int *p)
 {
 	return *p;
 }
--- a/benchmarks/compartment-call/caller.cc
+++ b/benchmarks/compartment-call/caller.cc
@@ -10,7 +10,7 @@
 
 using Debug = ConditionalDebug<DEBUG_CALLER, "Compartment call benchmark">;
 
-DECLARE_AND_DEFINE_ALLOCATOR_CAPABILITY(MALLOC_CAP_TWO, 1024);
+DECLARE_AND_DEFINE_ALLOCATOR_CAPABILITY(mallocCapTwo, 1024);
 
 __noinline int local_noop_return_metric()
 {
@@ -187,8 +187,7 @@ int __cheri_compartment("caller") run()
 		  [](int) { return local_check_pointer_return_metric(&nextResult); });
 
 		Timeout t{UnlimitedTimeout};
-		void   *alloc =
-		  heap_allocate(&t, STATIC_SEALED_VALUE(MALLOC_CAP_TWO), 10);
+		void *alloc = heap_allocate(&t, STATIC_SEALED_VALUE(mallocCapTwo), 10);
 
 		printf("#heap_claim_ephemeral\n");
 		benchWithMiddle(
--- a/benchmarks/timing.h
+++ b/benchmarks/timing.h
@@ -1,4 +1,3 @@
-#include "cdefs.h"
 #include <cheri.hh>
 #include <platform-rdcycle.h>
 
--- a/benchmarks/timing.h
+++ b/benchmarks/timing.h
@@ -46,18 +46,21 @@ namespace
 
 	void use_stack(size_t s)
 	{
-		volatile uint8_t stack_array[s];
-		stack_array[0] = 1;
+#pragma clang diagnostic push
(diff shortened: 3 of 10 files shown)


hlef:
This addresses https://github.com/CHERIoT-Platform/cheriot-rtos/issues/492.

nwf (approved):
LGTM; thank you!

benchmarks/compartment-call/callee.cc
```diff
@@ -15,7 +15,7 @@ int callee_stack_using_return_metric(size_t stackUse)
 	return METRIC();
 }
 
-int callee_dereference(int *p)
```
davidchisnall, line 18: These also need to change in the headers.  If you change the type of arguments, you'll get a linker failure.

benchmarks/compartment-call/caller.cc
```diff
@@ -10,6 +10,7 @@
 
 using Debug = ConditionalDebug<DEBUG_CALLER, "Compartment call benchmark">;
 
+// NOLINTNEXTLINE(readability-identifier-naming)
```
davidchisnall: We should probably fix the name, rather than turning off linting.

benchmarks/timing.h
```diff
@@ -46,18 +46,19 @@ namespace
 
 	void use_stack(size_t s)
 	{
-		volatile uint8_t stack_array[s];
-		stack_array[0] = 1;
+		volatile uint8_t *stackArray =
+		  static_cast<volatile uint8_t *>(__builtin_alloca(s));
+		stackArray[0] = 1;
```
davidchisnall: Why is this change?  We should never use `__builtin_alloca` directly.  The argument for using alloca instead of VLAs is that you should check for failure, but this does not check for failure (and will trap on failure anyway on CHERIoT, and CHERIoT lets you check for allocation failure with VLAs by checking the tag of the pointer).

There is an argument that you should avoid VLAs *and* alloca in embedded code because they mean you can't reason about stack usage, but that doesn't apply in a benchmark that's trying to measure the cost of variable-sized stack allocations.

davidchisnall (changes requested):

benchmarks/compartment-call/callee.cc
nding0405: Fixed by updating the .h to match the const int *.

benchmarks/compartment-call/caller.cc
nding0405: Fixed by renaming MALLOC_CAP_TWO to mallocCapTwo and updating the use site.

benchmarks/timing.h
nding0405: Fixed by reverting this back to the VLA form with a clang diagnostic suppression for the warning.

nding0405:
Thanks, I fixed the review comments: 
1. updated the compartment-call header declarations to match the const int * definitions, 
2. renamed the allocator capability, 
3. reverted the __builtin_alloca change back to the VLA with a scoped clang diagnostic ignorance.

benchmarks/timing.h
```diff
@@ -46,18 +46,21 @@ namespace
 
 	void use_stack(size_t s)
 	{
-		volatile uint8_t stack_array[s];
-		stack_array[0] = 1;
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wvla-cxx-extension"
```
davidchisnall:
```suggestion
__clang_ignored_warning_push("-Wvla-cxx-extension")
```

benchmarks/timing.h
```diff
@@ -46,18 +46,21 @@ namespace
...
+		volatile uint8_t stackArray[s];
+#pragma clang diagnostic pop
```
davidchisnall:
```suggestion
__clang_ignored_warning_pop()
```

davidchisnall:
We have some macros that wrap the diagnostic push / pop things that are gated on whether the compiler is clang and are a bit simpler than doing both pragmas.

benchmarks/timing.h
```diff
@@ -1,3 +1,4 @@
+#include "cdefs.h"
```
davidchisnall: This should definitely not use “, and should probably not be required (cdefs.h is included by all of our system headers).

davidchisnall (approved):
Thanks!

PR #726, opened 2026-07-02
Add an initial design document for clocks.

davidchisnall:
We currently maintain the wall-clock time in the network stack, this needs to be moved into the RTOS.

### DIFF
--- a/docs/Clocks.md
+++ b/docs/Clocks.md
@@ -1,0 +1,121 @@
+Design document for clocks
+==========================
+
+In CHERIoT RTOS today, we build everything in the core RTOS around a single clock source, which provides a monotonic clock.
+We also have a real-time clock (implementing `gettimeofday`) in the network stack, exposed from the SNTP compartment.
+
+POSIX defines a minimum of two clocks:
+
+ - `CLOCK_MONOTONIC`, which corresponds to the current clock.
+ - `CLOCK_REALTIME`, which is currently satisfied by the SNTP compartment but not integrated.
+
+Some APIs that we want to be able to implement, such as the POSIX threading APIs, take timeouts relative to the realtime clock, which means we need that to not come from an optional component.
+
+There is a single clock source for the monotonic clock, which is provided by the SoC and connected to an interrupt line so that the scheduler can wake processes based on it.
+In contrast, there are often multiple RTCs:
+
+ - A software-managed clock source from NTP or similar.
+ - An on-SoC RTC device that persists on low-power modes.
+ - An external RTC that has a separate battery backup and continues to increment time while the SoC is powered down.
+
+These have different properties.
+An off-chip RTC will always (after it's been set once) keep giving you time, but it might not be the correct time (drift of a second or more).
+SNTP will give you a very accurate time, as long as you're talking to an NTP server with a sensible stratum.
+Some clock sources are read-only (you can't set time from an NTP client) but some will want to persist time from higher-accuracy sources.
+The subsystem that manages clocks needs to support multiple sources with different properties.
+
+Exposing the realtime clock
+---------------------------
+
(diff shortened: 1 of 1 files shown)


docs/Clocks.md
```diff
@@ -0,0 +1,116 @@
...
+There is a shared memory object that contains a wall-clock time and the monotonic time at which it was correct, and an epoch.
+Callers can read both of these atomically using a small loop and then calculate the current monotonic time as a displacement from the wall-clock time.
+We may wish to make the epoch counter a futex: updates to the realtime clock are infrequent and it's useful to be able to watch for when the wall-clock time changes.
+
+Scheduler always uses monotonic time
+------------------------------------
+
+The scheduler is designed to be simple and so should continue to always use the monotonic clock for timeouts.
+APIs that want to wait for a realtime-clock-based update should convert the time to the monotonic time before calling scheduler APIs.
+This conversion is intrinsically racy, but so is anything using realtime clocks for timeouts.
```
rmn30, line 41: Could you elaborate on the race here? I guess you could be interrupted between reading the wall-clock and monotonic clock times from the shared object, but the epoch handles that. Once you have the pair you can accurately convert between given wall clock and monotonic without a race. I guess if you want to know what the time is "now" you have to read the monotonic clock and you could get interrupted before doing the conversion but that seems like an inherent problem with anything that wants to know the current time in a pre-emptible context.

rmn30: I guess if there's an update to the wall-clock time you could end up with an out of date conversion. You could always check the epoch again at the end of conversion if you really care?

docs/Clocks.md
```diff
@@ -0,0 +1,116 @@
...
+
+Wall-clock compartment
+---------------
+
+The wall-clock compartment should own the clock sources.
+Some clock sources are actually managed by other compartments, some are direct device access.
+Devices should look somewhat like this:
+
+```c++
+struct ExampleClockSource
```
rmn30, line 58: I guess you'll make this a concept?

docs/Clocks.md
```diff
@@ -0,0 +1,116 @@
...
+Each time source should be a separate header that defines a type that implements the time source.
+We need a mechanism in the board file to provide both the list of headers to include and the list of types to instantiate.
+
+*Open question*: What should this look like?
+Users can provide board files that patch an existing board file with JSON patch to add modify arrays quite easily, so if we list both the include files and patches as arrays, users can add and remove individual ones easily and xmake can synthesise a header that includes all of the types and provide their names.
+
+Out of scope functionality
+--------------------------
+
+Time zones.
```
rmn30, line 113: Fine, but you haven't specified an epoch for the wall clock time. Unix?

rmn30 (approved):
Sounds good to me.

docs/Clocks.md
davidchisnall: Yup.

docs/Clocks.md
davidchisnall: Oh, good point. It needs zero defined. Using the UNIX epoch is probably sensible. OS X defined the Cocoa epoch as the start of the millennium, which might be better (more useful range in a 32-bit value), but I think most APIs we want will use a POSIX timespec.

docs/Clocks.md
davidchisnall: Yup, you read wall-clock time, NTP sets wall-clock time with a big jump, you calculate a monotonic time relative to that, then wall-clock time moves before you sleep.

If the scheduler knows about wall-clock time, it may choose to wake things that waited on wall-clock time.

davidchisnall:
New version adds clarifications for Robert's comments, added to the merge queue.

docs/Clocks.md
rmn30: The RTC on the Adafruit HAT I used had a two digit BCD year! I made the assumption that it was a 21st century date...

PR #727, opened 2026-07-02
Allocator claims rework

nwf:
This changes the internal workings of `Claim` objects within the heap allocator, specifically so we no longer need to take a pointer to the claims chain head field within a `MChunkHeader`.  With that change made, we should be able to vary the widths of owner and claims (encoded) pointer fields within `MChunkHeader` and `Claim`s, and in particular make this width larger than 16 bits, which is one, and possibly the simplest, approach to getting us to supporting more than 512KiB heaps.

### DIFF
--- a/sdk/core/allocator/main.cc
+++ b/sdk/core/allocator/main.cc
@@ -456,12 +456,13 @@ namespace
 		/**
 		 * The identifier of the owning allocation capability.
 		 */
-		uint16_t allocatorIdentifier = 0;
+		uint16_t allocatorIdentifier : MChunkHeader::OwnerIDWidth = 0;
 		/**
 		 * Next 'pointer' encoded as a shifted offset from the start of the
 		 * heap.
 		 */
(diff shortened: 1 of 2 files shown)


sdk/core/allocator/main.cc
```diff
@@ -984,10 +943,10 @@ __cheriot_minimum_stack(0x220) void *heap_allocate(
 	  .as_raw();
 }
 
-__cheriot_minimum_stack(0x1c0) ssize_t
+__cheriot_minimum_stack(0x1d0) ssize_t
```
davidchisnall, line 956: Do you know why this got bigger?

davidchisnall (approved):

sdk/core/allocator/main.cc
nwf: My unverified suspicion is that it's the `ChainPredecessor` tagged union (a pointer and a flag) vs just the `uint16_t` we were carrying around before?

PR #728, opened 2026-07-06
Add more doxygen things

davidchisnall:
Two commits, one adds docs for the FreeRTOS compatibility layer, the second for the compartment-helper macros.


davidchisnall:
One more commit that adds a bunch more missing docs.

nwf (approved):
LGTM, so I'm happy when CI's happy.

PR #731, opened 2026-07-09
Complete the migration to TimeoutArgument.

davidchisnall:
PR #708 moved to using either relative timeouts in ticks or absolute timeouts in cycles on some clock.  That PR intentionally did not move *all* APIs to using the new interface, so that we could be certain that callers were not broken.  Now that this is done, move everything to the newer structure.

### DIFF
--- a/sdk/include/timeout.hh
+++ b/sdk/include/timeout.hh
@@ -12,8 +12,7 @@
 template<auto Fn, typename... Args>
 __always_inline auto blocking_forever(Args... args)
 {
-	Timeout t{UnlimitedTimeout};
-	return Fn(&t, std::forward<Args>(args)...);
+	return Fn(TimeoutWaitForever, std::forward<Args>(args)...);
 }
 
 /**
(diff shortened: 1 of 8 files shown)


sdk/include/timeout.hh
```diff
@@ -12,8 +12,7 @@
 template<auto Fn, typename... Args>
 __always_inline auto blocking_forever(Args... args)
 {
-	Timeout t{UnlimitedTimeout};
-	return Fn(&t, std::forward<Args>(args)...);
+	return Fn(TimeoutWaitForever, std::forward<Args>(args)...);
```
rmn30, line 15: Does the old version of this and `TimeoutNoWait` still work?

sdk/lib/queue/queue_compartment.cc
```diff
@@ -100,11 +100,11 @@ int queue_destroy_sealed(Timeout            *timeout,
 	  });
 }
 
-int queue_send_sealed(Timeout *timeout,
+int queue_send_sealed(TimeoutArgument timeout,
                       CHERI_SEALED(MessageQueue *) handle,
                       const void *src)
 {
-	if (!check_timeout_pointer(timeout))
+	if (!timeout.is_valid())
```
rmn30, line 107: Does existing code using `check_timeout_pointer` still work?

rmn30 (approved):
Looks fine but I'm just wondering about backwards compatibility. I guess this is already a concern with #708 but I just want to clarify what is an is not expected to work.

PR #732, opened 2026-07-09
Revised version of the sub-quotas

hlef:
Here is a revised version of Milo's sub-quota PR (https://github.com/CHERIoT-Platform/cheriot-rtos/pull/691).

Changes:
- Rebased on top of recent allocator changes.
- Eliminated `HeapOffset`.
- Addressed style and minor comments, more documentation.
- Among others, using `-ECHILD` for running out of quota IDs (I like the analogy with failing to create a child process, and the child terminology is very suitable).

Happy to give this another round of edits if needed. I'd keep the TODOs as TODOs though, as I do not have cycles to do these now.

### DIFF
--- a/sdk/core/allocator/main.cc
+++ b/sdk/core/allocator/main.cc
@@ -34,19 +34,79 @@ using namespace CHERI;
 Revocation::Revoker revoker;
 namespace
 {
+	// The global memory space
+	MState *gm;
+
 	/**
 	 * Internal view of an allocator capability.
 	 *
-	 * TODO: For now, these are statically allocated. Eventually we will have
-	 * some that are dynamically allocated. These should be ref counted so that
-	 * we can avoid repeated validity checks.
+	 * This is used for static allocator capabilities as well as for
+	 * dynamic (sub-) allocator capabilities. The latter are stored on the
+	 * heap; the parent/child relationship is encoded by the child
+	 * allocator capability object being owned by the parent allocator
+	 * capability on the heap, rather than anything in the object itself.
 	 */
 	struct PrivateAllocatorCapabilityState
 	{
-		/// The remaining quota for this capability.
+		/**
+		 * The remaining quota for this capability.
+		 */
 		size_t quota;
-		/// A unique identifier for this pool.
+		/**
+		 * A unique identifier for this pool.
+		 */
 		uint16_t identifier;
+		/**
+		 * Pointers forming a doubly-linked list for dynamically
+		 * allocated sub-quotas.
+		 *
+		 * TODO It would be great if we could re-use `ds::linked_list`.
+		 * This requires a custom `OffsetPtrAddr` class that offsets
+		 * relative to the heap start and that encodes and decodes on
+		 * list operations. This change should probably be done when we
+		 * increase the size of heaps since `uint16_t` will be too
+		 * small anyways.
+		 */
+		uint16_t next;
+		uint16_t prev;
+
+		/**
+		 * Decode an encoded offset and return a pointer to the
+		 * allocator capability state.
+		 *
+		 * TODO This duplicates similar functions in the Claim object.
+		 * It would be great if we could factor these out.
+		 */
+		static PrivateAllocatorCapabilityState *
+		from_encoded_offset(uint16_t offset)
+		{
+			if (offset == 0)
+			{
+				return nullptr;
+			}
+			Capability<PrivateAllocatorCapabilityState> ret{
+			  gm->heapStart.cast<PrivateAllocatorCapabilityState>()};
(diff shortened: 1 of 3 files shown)


hlef:
Intentionally not fixing the warning on the `reinterpret_cast<AllocatorCapability>(ret)` as I'd like to know what you think @nwf. Is that the right way of having the error code flow from `allocate_sealed_unsealed_internal` (without intrusive changes?).

sdk/core/allocator/main.cc
```diff
@@ -34,28 +34,267 @@ using namespace CHERI;
...
+		/**
+		 * Pointers forming a doubly-linked list for dynamically
+		 * allocated sub-quotas.
+		 *
+		 * TODO It would be great if we could re-use `ds::linked_list`.
+		 * This requires a custom `OffsetPtrAddr` class that offsets
+		 * relative to the heap start and that encodes and decodes on
+		 * list operations.
+		 */
+		uint16_t next;
```
nwf, line 70: Do we have room in the structure for these to get bigger than `uint16_t`-s?

sdk/core/allocator/main.cc
```diff
@@ -1446,6 +1773,215 @@ __cheriot_minimum_stack(TokenObjDestroyStackUsage) int token_obj_can_destroy(
...
+	auto sealed = *reinterpret_cast<AllocatorCapability *>(&sealedWrapped);
+	// Casting to Private lets us write to our prev/next pointers.
+	auto *unsealed = reinterpret_cast<PrivateAllocatorCapabilityState *>(
+	  unsealedWrapped.get());
+
+	// Allocate the quota to the child
+	parentCap->quota -= size;
+	unsealed->quota = size;
+
+	if (subQuotaIds.id_allocate(unsealed))
```
nwf, line 1832: Hrm.  This is _probably_ fine, but the need to allocate and then unwind is a bit unfortunate.  Could `subQuotaIds` have an idempotent `.can_allocate()` (name subject to bikeshedding) that does all the work of _finding_ but none of the _updating_ and returns either failure or enough state to know what to do once allocation succeeds?

sdk/core/allocator/main.cc
hlef: I think that this shouldn't be a problem? Should I let you update them when you push your PRs to support larger heaps?

sdk/core/allocator/main.cc
hlef: Agree, this would be a bit more elegant. Let me see if I can add that...

hlef: OK, thinking about it, one big downside of first checking if a subquota ID can be allocated, and then actually allocating it later, is that we end up walking the ID space twice. It's expensive, and I'd like to keep the allocation of sub-quotas as cheap as possible, since it may be done once per connection in the network use-case...

One could picture storing the ID found during the first walk and reuse it during the actual allocation, but it's a bit ugly and breaks the encapsulation of the SubQuotaIdSpace; I also see a lot of potential for things to go wrong there.

For now I suggest putting a comment at that place in the code to explain the choice, happy to do something else if you have a better idea.

nwf: As discussed out of band, but for the record.  My suggestion is that `.can_allocate()` indicates success with return of an instance of an opaque class, which `.allocate()` then takes, and which holds sufficient information about the search result so that `.allocate()` is O(1).  I _think_ it sufficient to hold a pointer to the to-be-prior node and the chosen new identifier.

sdk/core/allocator/main.cc
nwf: Yeah, is fine for the moment.  I wish we could have used something like `HeapOffset` to indicate what's going on, rather than needing to play the fun game of "what is this `uint16_t` actually encoding", but here we are and this PR isn't the right place to enlarge the types or start to fix it more generally.

hlef: I have added some of that in the comment.

sdk/core/allocator/main.cc
hlef: Discussed out of band again, for the record: I looked into it and while I agree that this is the right way to do it, I do not really have the cycles to do this as part of this PR. I left a comment, it'd be great if we could do this as a future improvement. I'd be happy to also create a bug report to track these little things that need to be fixed.

hlef:
> Intentionally not fixing the warning on the `reinterpret_cast<AllocatorCapability>(ret)` as I'd like to know what you think @nwf. Is that the right way of having the error code flow from `allocate_sealed_unsealed_internal` (without intrusive changes?).

For now, I used `__clang_ignored_warning_push` to silence that warning.

PR #733, opened 2026-07-09
Fixed multiwaiter_wait capability bound check

nding0405:
**Problem:**
`multiwaiter_wait` should check that the `events` capability covers the `EventWaiterSource` array. The old code used `sizeof(newEventsCount)`, which is equivalent to `sizeof(size_t)` instead of the actual element size.

**Solution:**
Changed the bounds check to use `sizeof(EventWaiterSource)`.

**Checks:**
1. Ran tests after the fix.
2. Ran clang-format and clang-tidy.


nwf (approved):
LGTM.  @davidchisnall, nothing subtle I'm missing here?

davidchisnall (approved):

PR #734, opened 2026-07-11
Add support for clocks as described in docs/Clocks.md.

davidchisnall:
This provides support for plugging in multiple clock sources. A parallel PR will refactor the existing SNTP support to use these interfaces.

### DIFF
--- a/docs/Clocks.md
+++ b/docs/Clocks.md
@@ -83,7 +83,7 @@ struct ExampleClockSource
 	 * this time source is that this is a good time.  Larger values indicate
 	 * more confidence.
 	 */
-    int get_time(uint64_t &outRealTime, uint64_t &outMonotonicTime, int &outPriority);
+    int get_time(TimeoutArgument timeout, uint64_t &outRealTime, uint64_t &outMonotonicTime, int &outPriority);
 
 	/**
 	 * Set the current wall-clock time.  This return 0 on succcess, some
--- a/sdk/include/time.h
+++ b/sdk/include/time.h
@@ -1,5 +1,136 @@
 // Copyright Microsoft and CHERIoT Contributors.
 // SPDX-License-Identifier: MIT
+#pragma once
 
-// libc++ headers expect this file to exist, but don't require any definitions
-// from it and so we can provide an empty version.
+/**
+ * \file .
+ *
+ * Standard clock support.
+ */
+
+#include <platform-time.h>
+#include <stdint.h>
+#include <sys/time.h>
+#include <thread.h>
+
+// The names in this file come from C or POSIX and so do not correspond to our
+// naming scheme.  This header is expected to be included in C, so should also
+// not provide warnings about void in function parameter lists.
+// NOLINTBEGIN(readability-identifier-naming,modernize-redundant-void-arg)
+
+/// ID for a clock.  Only monotonic and 'realtime' (wall clock) are supported.
+typedef enum __clockid_t
+{
+	/**
+	 * The monotonic clock. This is zero at system start and increments at a
+	 * fixed rate.
+	 */
+	CLOCK_MONOTONIC,
+	/**
+	 * The wall-clock time.
+	 *
+	 * This clock's value is meaningful only if `clock_update_wall_clock` has
+	 * been called at least once and there is at least one working clock source
+	 * in the system.
+	 */
+	CLOCK_REALTIME,
+	/**
+	 * CPU time consumed by the current thread, since boot time.  This clock is
+	 * equivalent to `CLOCK_MONOTONIC` if the scheduler is not compiled with
+	 * support for accounting.  Add `--scheduler-accounting=y` to your build
+	 * configuration line to enable this.
+	 */
+	CLOCK_THREAD_CPUTIME_ID,
+	/**
+	 * CPU time consumed by the current 'process'.  This value is defined by
+	 * POSIX, but CHERIoT RTOS does not have a direct equivalent of a process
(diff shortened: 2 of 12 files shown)


sdk/include/time.h
```diff
@@ -1,5 +1,129 @@
...
+
+/**
+ * A time value with up to nanosecond precision.
+ */
+struct timespec
+{
+	/// Seconds
+	time_t tv_sec;
+	/// Nanoseconds
+	long tv_nsec;
```
nwf: Do we want to use `uint32_t` here, since C23 gives us license to be more specific than `long`?

sdk/include/time.h
```diff
@@ -1,5 +1,129 @@
...
+ * are accounted to the current thread (POSIX specifies 'process' here, but
+ * CHERIoT RTOS does not have an directly analogous abstraction).
+ *
+ * Note: If scheduler accounting is not enabled, this API will return the
+ * total elapsed uptime instead.  Add `--scheduler-accounting=y` to your build
+ * configuration line to enable this.
+ */
+static inline clock_t clock(void)
+{
+#if SCHEDULER_ACCOUNTING == true
```
nwf, line 101: This is interesting, since some `--${component}-${option}` xmake options are specific to the `${component}` in question; some can propagate up xmake dependency chains; and some, like this, seem like they're global and need(?) to be pushed into every component from the top.

I don't... particularly love that, but I don't have a better solution off the top of my head.  Should we _generate_ this header (or an internal one included herein for `clock()`)?

sdk/lib/clock/README.md
```diff
@@ -0,0 +1,49 @@
...
+The concept is in the [platform/concepts/wall_clock_source.hh](../../include/platform/concepts/wall_clock_source.hh) header.
+Implementations may be entirely inline in the header (for example, if the clock is a hardware device with a simple interface) or may call other compartments (for example, if the clock source is NTP or similar).
+
+The `wall_clock` compartment will perform the entire update sequence with a lock held, so clock implementations do not need to handle reentrant invocation.
+They should gracefully recover if they trap and not leave themselves in an unrecoverable state.
+
+If your implementation is called `MyWallClockSource` then you may find it useful to add a line such as this to validate compliance with the concept:
+
+```c++
+static_assert(IsWallClockSource<MyWallClockSource>, "The SNTP wall-clock source must implement the required concept");
```
nwf: Probably something clang-format won't balk at?

```suggestion
static_assert(IsWallClockSource<MyWallClockSource>,
              "The SNTP wall-clock source must implement the required concept");
```

sdk/lib/clock/wall_clock.cc
```diff
@@ -0,0 +1,180 @@
...
+#include <timeout.h>
+#include <unwind.h>
+
+// Appease clang-tidy and LSP by providing a stub implementation if we don't
+// have a build configuration that provides this file yet.
+#if __has_include("rtc_sources.hh")
+#	include "rtc_sources.hh"
+#else
+#	include <tuple>
+struct NoWallClockSource
```
nwf, line 18: Add a comment that this needs to stay in sync with `rtc_sources.hh.in`'s `NoWallClockSource`?

sdk/lib/clock/xmake.lua
```diff
@@ -0,0 +1,68 @@
+-- Copyright Microsoft and CHERIoT Contributors.
+-- SPDX-License-Identifier: MIT
+
+includes("../freestanding", "../compartment_helpers", "../atomic")
```
nwf: Are these not already in scope by the time it matters?

sdk/include/time.h
davidchisnall: POSIX and pre-C23 require it to be long so I'm a bit conflicted on this one.  On the one hand, `uint32_t` is clearly the correct type.  On the other hand, there may be code that expects it to be `long` because that's what it's been required to be for the last 29 or so years.

sdk/include/time.h
davidchisnall: I really hate generated headers in xmake, so I'd like to avoid them where possible.  I think it's actually possible now in xmake to have options that add cxflags to everything without the walk-all-the-dependencies thing.

sdk/lib/clock/wall_clock.cc
davidchisnall: It doesn't actually need to stay in sync, just implement the concept.  I've added a static assert to make sure it does.

sdk/lib/clock/xmake.lua
davidchisnall: They actually don't need to be at all.  They're used for add_deps, but that resolves after loading everything.  Removed.

docs/Clocks.md
rmn30, line 5: Could you update this documentation to match the new state of things?

davidchisnall: What do you mean?  Design docs are design docs.  The design was updated in this PR.

We should separate design docs from user docs more, not make design docs less useful as design docs.  Design docs should *not* contain the current state, they should document the design process and requirements.

rmn30: OK, so where should a user look for documentation on how to use this? In sdk/lib/clocks? That seems rather hidden away. For one thing it won't be processed by our current doxygen configuration but docs/Clocks.md will be.

davidchisnall: sdk/lib/clocks is the right place to look.  All of the libraries have README, we should be pulling them into Doxygen.

rmn30 (approved):

PR #735, opened 2026-07-13
Fix for bitpacked UART (incorrect placing of fields in one of the registers)

venngar:
PR #558 have added bitpacked `platform-uart.hh` implementation, however `TransmitWatermark` and `ReceiveWatermark` are located in wrong position according to [OpenTitan documentation](https://opentitan.org/book/hw/ip/uart/doc/registers.html#fifo_ctrl). In addition to that, for `ReceiveWatermark` the max value enum is `Level62` instead of `Level64`.

This PR fixes both issues.


nwf (approved):
Thanks!

PR #737, opened 2026-07-18
Bitpacks: add required remove_reference_t in macro definitions

rmn30:
This is required when operating on a member of an array because
decltype adds ref to the type and the compiler refuses to allow it to
be used as a namespace. There may be more places where this is
required.


nwf (approved):
LGTM.  Thanks and doubly so for the added tests.

PR #738, opened 2026-07-20
Fix computation of PARALLEL_JOBS in run_clang_tidy_format.sh to work on macOS as well.

resistor:


scripts/run_clang_tidy_format.sh
```diff
@@ -21,11 +21,7 @@ if [ ! -x ${CLANG_FORMAT} ] ; then
 	exit 1
 fi
 
-if which nproc ; then
-	PARALLEL_JOBS=$(nproc)
-else
-	PARALLEL_JOBS=$(sysctl -n kern.smp.cpus)
-fi
+PARALLEL_JOBS=$(sysctl -n kern.smp.cpus 2>/dev/null || sysctl -n hw.ncpu 2>/dev/null || nproc)
```
davidchisnall: nproc also exists on macOS, it's a tiny shell script around the sysctl.  It might be better to try that as the first one to try and fall back to the alternatives if it doesn't exist.

resistor: There's no `nproc` on my macOS...

resistor: Went ahead and moved `nproc` to the first thing to try anyways.

davidchisnall: It's in homebrew.  I installed it because a bunch of scripts expect it to exist.

davidchisnall (approved):

PR #740, opened 2026-07-20
Remove redundant void parameter that clang-tidy 23 is unhappy about.

resistor:


sdk/include/thread.h
```diff
@@ -80,7 +80,7 @@ enum ThreadSleepFlags : uint32_t
  *
  * This is implemented in the switcher.
  */
-uint16_t __cheri_libcall thread_id_get(void);
+uint16_t __cheri_libcall thread_id_get();
```
nwf: Alas, these mean different things in C, and this header can be included in C code, not just C++.

sdk/include/thread_pool.h
```diff
@@ -161,7 +161,7 @@ namespace thread_pool
 	{
 		// If this is a stateless function, just send a callback function
 		// pointer, don't copy zero bytes of state to the heap.
-		if constexpr (std::is_convertible_v<T, void (*)(void)>)
+		if constexpr (std::is_convertible_v<T, void (*)()>)
```
nwf, line 164: This one seems fine; despite the `.h`, this bit is definitely C++-only.

sdk/include/thread.h
resistor: clang-tidy seems to believe they're no longer different in C23, which is what we're setting in xmake. If we want to retain support for old C standard, I can check this one to a NOLINT.

resistor: Changed to NOLINT on the assumption that we want to support pre-C23.

sdk/include/thread.h
```diff
@@ -80,6 +80,7 @@ enum ThreadSleepFlags : uint32_t
  *
  * This is implemented in the switcher.
  */
+// NOLINTNEXTLINE(modernize-redundant-void-arg)
```
davidchisnall: We also have `__if_c` and `__if_cxx` macros.  This might be more concise as `__if_c(void)` on the next line.

resistor: Oh nice, I wasn't aware of that macro. Swapped to that instead.

davidchisnall (approved):

sdk/include/thread.h
```diff
@@ -80,7 +80,7 @@ enum ThreadSleepFlags : uint32_t
  *
  * This is implemented in the switcher.
  */
-uint16_t __cheri_libcall thread_id_get(void);
+uint16_t __cheri_libcall thread_id_get(__if_c(void));
```
nwf: Maybe that deserves a comment as "here for pre-C23 compatibility" or something?  Can we gate on C version, not just language? ^^;

resistor: Comment added.

PR #741, opened 2026-07-20
NOLINT values with externally exposed names that clang-tidy23's readability checker doesn't like.

resistor:


sdk/core/loader/constants.h
```diff
@@ -41,10 +41,17 @@ EXPORT_ASSEMBLY_OFFSET(SchedulerEntryInfo, schedulerExportEntry, 8);
 EXPORT_ASSEMBLY_OFFSET(SchedulerEntryInfo, initialiserList, 16);
 EXPORT_ASSEMBLY_OFFSET(SchedulerEntryInfo, threads, 24);
 EXPORT_ASSEMBLY_SIZE(SchedulerEntryInfo, 32);
-EXPORT_ASSEMBLY_EXPRESSION(MIE_MEIE, priv::MIE_MEIE, 0x800);
-EXPORT_ASSEMBLY_EXPRESSION(MIE_MTIE, priv::MIE_MTIE, 0x080);
-EXPORT_ASSEMBLY_EXPRESSION(MSTATUS_MIE, priv::MSTATUS_MIE, 8);
-EXPORT_ASSEMBLY_EXPRESSION(ThreadHeapHazards_size,
-                           HazardPointersPerThread * sizeof(void *),
-                           16);
+EXPORT_ASSEMBLY_EXPRESSION(MIE_MEIE, // NOLINT(readability-identifier-naming)
```
nwf: Hrm... while the on-line `// NOLINT` form works at the moment, is it possible that changing the definition of this macro could change that?  If so, maybe the `// NOLINTBEGIN` form instead?

resistor: Done

nwf (approved):

PR #742, opened 2026-07-20
Fix instances where clang-tidy23 complains about else blocks after non-returning function calls.

resistor:


nwf (approved):
Seems OK by me if CI's happy.

PR #748, opened 2026-07-31
Add cancellation-token APIs.

davidchisnall:
These are modelled on their .NET equivalent, with mostly the same properties.  They provide a mechanism for coordinating cancellation, where one token can be shared with an arbitrary number of long-running tasks and they can all be notified that they're cancelled.

Also includes a commit that fixes some issues with another header that this exposed.

### DIFF
--- a/sdk/include/cancellation.h
+++ b/sdk/include/cancellation.h
@@ -1,0 +1,92 @@
+// Copyright SCI Semiconductor and CHERIoT Contributors.
+// SPDX-License-Identifier: MIT
+
+#pragma once
+#include <cdefs.h>
+#include <stdlib.h>
+#include <timeout.h>
+
+/**
+ * \file .
+ *
+ * Cancellation token APIs.
+ *
+ * Cancellation tokens are an abstraction that allows long-running tasks in
+ * other threads to be cooperatively cancelled.  A cancellation token is created
+ * from a *cancellation-token source*.  The holder of the cancellation-token
+ * source may signal cancellation.  Holders of the token can check whether it
+ * has been cancelled.
+ */
+
+/**
+ * State associated with a cancellation token.  The cancellation token and the
+ * cancellation-token source are both views of this object, but it is not used
+ */
+struct CancellationTokenState;
+
+/**
+ * The handle for a cancellation-token source.
+ */
+typedef struct CancellationTokenState
+  *__sealed_capability CancealltionTokenSource;
+
+/**
+ * The type for a cancellation token.  This can be queried, but cannot be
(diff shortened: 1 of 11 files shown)


sdk/include/cancellation.h
```diff
@@ -0,0 +1,92 @@
...
+ * Cancellation tokens are an abstraction that allows long-running tasks in
+ * other threads to be cooperatively cancelled.  A cancellation token is created
+ * from a *cancellation-token source*.  The holder of the cancellation-token
+ * source may signal cancellation.  Holders of the token can check whether it
+ * has been cancelled.
+ */
+
+/**
+ * State associated with a cancellation token.  The cancellation token and the
+ * cancellation-token source are both views of this object, but it is not used
```
rmn30, line 23: sentence truncated?

sdk/include/cancellation.h
```diff
@@ -0,0 +1,92 @@
...
+ */
+struct CancellationTokenState;
+
+/**
+ * The handle for a cancellation-token source.
+ */
+typedef struct CancellationTokenState
+  *__sealed_capability CancealltionTokenSource;
```
rmn30, line 31:
```suggestion
  *__sealed_capability CancellationTokenSource;
```
This typo is repeated throughout.

sdk/include/cancellation.h
```diff
@@ -0,0 +1,92 @@
...
+typedef const struct CancellationTokenState *CancellationToken;
+
+/**
+ * Create a cancellation-token source, allocated from the caller's quota.
+ *
+ * This returns either a tagged capability for a cancellation token or an
+ * untagged value that is an error code that corresponds to a value that the
+ * allocator would return on failure.
+ */
+CancealltionTokenSource __cheriot_compartment("cancellation_token")
```
rmn30, line 46: Didn't we make ErrorOr for this?

davidchisnall: For C++, this is a C API.

sdk/include/cheri-builtins.h
```diff
@@ -24,7 +24,7 @@
 #	include <cdefs.h>
 #	include <stddef.h>
 
-static inline __always_inline ptraddr_t cheri_address_get(void *x)
+static inline __always_inline ptraddr_t cheri_address_get(auto *x)
```
rmn30, line 27: Doesn't this need to work in C?

davidchisnall: Line 23 is `#ifdef __cplusplus`.

PR #750, opened 2026-08-05
Loader: warn about possibly wrong libcalls when lib size > 64KiB

xdoardo:


sdk/core/loader/boot.cc
```diff
@@ -697,6 +697,15 @@ namespace
 			{
 				if (contains<ExportEntry>(lib.exportTable, possibleLibcall))
 				{
+					auto libsize =
+					  lib.importTableSize + lib.code.size() + lib.data.size();
```
davidchisnall: I think this should just be `lib.code.size()`.  We don't care about the size of the other two, they're separate sections.

jacobpake:
We also look for exports in `image.privilegedCompartments`, `image.switcher` and `sourceCompartment`.

I am not sure about the first two, but I would like this check on the `sourceCompartment` also.

sdk/core/loader/boot.cc
xdoardo: [45fdfb6](https://github.com/CHERIoT-Platform/cheriot-rtos/pull/750/commits/45fdfb60e2253f3f9847dddca513198371ca93ae)

xdoardo:
> We also look for exports in `image.privilegedCompartments`, `image.switcher` and `sourceCompartment`.
> 
> I am not sure about the first two, but I would like this check on the `sourceCompartment` also.

[ac6fc88](https://github.com/CHERIoT-Platform/cheriot-rtos/pull/750/commits/ac6fc8885ee68a27dcc41744b0d344a6e7f79345). Will squash if accepted.

davidchisnall (approved):

PR #752, opened 2026-08-07
Add gmtime / timegm implementations.

davidchisnall:
These convert between human-usable UTC and UNIX timestamp representations.

### DIFF
--- a/sdk/lib/time/timegm.cc
+++ b/sdk/lib/time/timegm.cc
@@ -1,0 +1,337 @@
+// Copyright SCI Semiconductor and CHERIoT Contributors.
+// SPDX-License-Identifier: MIT
+
+/**
+ * This file provides implementations of the timegm and gmtime_r functions,
+ * which convert between a human-friendly time and a UNIX timestamp.
+ *
+ * UNIX timestamps ignore leap seconds, which means that this conversion can
+ * also do so,
+ */
+
+#include <array>
+#include <debug.hh>
+#include <time.h>
+
+namespace
+{
+	using Debug = ConditionalDebug<true, "Time">;
+
+	/// The number of seconds in one non-leap-second minute
+	constexpr int SecondsPerMinute = 60;
+	/// The number of minutes in one hour.
+	constexpr int MinutesPerHour = 60;
+	/// The number of seconds in one hour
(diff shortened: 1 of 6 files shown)


sdk/lib/time/timegm.cc
```diff
@@ -0,0 +1,337 @@
...
+		static constexpr std::array<int, 7> CenturyCodes = {
+		  4, 2, 0, 6, 4, 2, 0};
+		int yearInCentury = year % 100;
+		int yearCode      = ((yearInCentury + (yearInCentury / 4)) % 7);
+		int monthCode     = MonthCodes[month];
+		int century       = year / 100;
+		// Every day outside of the range this calculation works for is Monday.
+		if ((century < 17) || (century > 23))
+		{
+			return 0;
```
nwf, line 133: 0 is Sunday, not Monday?

sdk/lib/time/timegm.cc
```diff
@@ -0,0 +1,337 @@
...
+	{
+		static constexpr std::array<int, MonthsPerYear> MonthCodes = {
+		  0, 3, 3, 6, 1, 4, 6, 2, 5, 0, 3, 5};
+		static constexpr std::array<int, 7> CenturyCodes = {
+		  4, 2, 0, 6, 4, 2, 0};
+		int yearInCentury = year % 100;
+		int yearCode      = ((yearInCentury + (yearInCentury / 4)) % 7);
+		int monthCode     = MonthCodes[month];
+		int century       = year / 100;
+		// Every day outside of the range this calculation works for is Monday.
```
davidchisnall:
```suggestion
		// Every day outside of the range this calculation works for is Sunday.
```

nwf (approved):
I have fallen down a rabbit hole of NTP and POSIX and leap seconds and I think it all works out just fine for cases we care about.  LGTM.

davidchisnall:
> I have fallen down a rabbit hole of NTP and POSIX and leap seconds and I think it all works out just fine for cases we care about. LGTM.

I initially implemented all of the logic for leap seconds, before reading POSIX more carefully and learning that POSIX time just pretends that they never happened, and removed it.

PR #754, opened 2026-08-12
Add link-time check for negative heap size

jacobpake:
Asserts that `__export_mem_heap`, whose position is dynamic, comes before `__export_mem_heap_end`, whose position is statically defined in the board definition.

Under normal conditions this will already trigger the heap_maximum_size check below, but provides a more appropriate error message.

For board definitions where instruction_end == heap_end, such as for Sail, this also catches the case where we have exceeded the instruction memory.

This does not check that the heap size is above the minimum needed by the allocator.

### DIFF
--- a/sdk/firmware.rwdata.ldscript.in
+++ b/sdk/firmware.rwdata.ldscript.in
@@ -79,6 +79,8 @@ __export_mem_heap = @heap_start@;
 
 ASSERT(__export_mem_heap >= ., "Heap would overlap static data")
 
+ASSERT(__export_mem_heap < __export_mem_heap_end, "Heap has no available space")
+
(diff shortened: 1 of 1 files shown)


jacobpake:
I had assumed failing the assertion would cause an early exit but I see:

> error: ld.lld: error: Heap has no available space
> ld.lld: error: Heap arena is too large for the current allocator implementation

Is that a problem?

davidchisnall:
I think that is an early exit?  It's erroring before it produces the output.

davidchisnall (approved):

jacobpake:
> I think that is an early exit? It's erroring before it produces the output.

I mean an earlier exit, I wasn't expecting it to check the next assertion

PR #755, opened 2026-08-17
xmake board parsing: add boardBase path, tweak mixin resolution

nwf:
And while here, replace relative paths in the `sdk/boards` with ones using the `${sdk}` path substitution.

`${boardBase}` is of particular interest to out-of-tree BSPs: their base board files can use this to make it easier to combine them with (per-project, say) board patch files.


davidchisnall (approved):
Small patch, big improvement, thanks!

PR #761, opened 2026-08-24
Improve documentation of multiwaiters.

davidchisnall:


sdk/include/multiwaiter.h
```diff
@@ -17,11 +17,18 @@
...
- * write to it from another thread.
+ * to a fixed limit.  Each wait call provides a set of futexes to wait on, along
+ * with their expected values, and will suspend if all currently hold their
+ * expected values at the time of the call.  It will then sleep until either one
+ * (or more) of the futexes is woken or a timeout is reached.  On return, the
+ * caller-provided list will be updated, replacing the expected values with 0
+ * if this futex is not woken or 1 if it is.  Note that this, like the
+ * single-futex futex-wait API, is inherently racy: a futex value may be changed
+ * and the futex notified immediately after the multiwaiter returns and before
+ * the caller has inspected the results.
```
rmn30, line 28: I feel like this should have an explanation of why that is OK.

davidchisnall: What would you suggest?

sdk/include/multiwaiter.h
```diff
@@ -84,9 +91,33 @@ typedef CHERI_SEALED(struct MultiWaiterInternal *) MultiWaiter;
...
+ * happens then the multiwaiter will return immediately.  If not, then the
+ * multiwaiter will block until either one of the futexes is signalled with
+ * `futex_wake` or until the timeout (specified in `timeout`) expires.
+ *
+ * On successful return, the `value` field for each element in `newEvents` will
+ * be set to zero or one.  A value of one indicates that the futex identified
+ * by the corresponding `eventSource` field was signalled, or had changed
+ * before the initial check.  A value of zero indicates that this futex was not
+ * signalled, but does not imply that the event did not happen immediately
+ * after return.
```
rmn30, line 111: To be clear, is it possible for more than one value other be set to 1? Under what circumstances? Does that happen only in the case where there was a futex mismatch before waiting? If we resume due to a futex_wake then will only one be `value` be 1?

davidchisnall: It depends on scheduling.  If two futexes have `futex_wake` called on them before this thread is scheduled, they may both show as signalled.  But that's a scheduler implementation detail and subject to change, not an API policy.

sdk/include/multiwaiter.h
rmn30: "Therefore the caller should examine the value of the futex or event source to determine whether action is required."

Something to indicate how to correctly use the API. Makes me wonder whether setting it to 1 is useful at all? The caller probably has to examine all the futex values anyway...

PR #762, opened 2026-08-24
Correctly use array size rather than scalar size in randombytes() helper.

resistor:


nwf (approved):
I think this is right, since the type of `output` is a reference to `T[N]` and `sizeof` on an expression with a reference type returns the size of the referenced type.

PR #766, opened 2026-08-24
interrupt.h: add C++ wrapper utility method

nwf:
(As yet untested, sorry.  Posting for stylistic opinions early.)

### DIFF
--- a/sdk/include/interrupt.h
+++ b/sdk/include/interrupt.h
@@ -133,3 +133,103 @@ __cheri_compartment("scheduler") const uint32_t *interrupt_futex_get(
  */
 __cheri_compartment("scheduler") int interrupt_complete(
   InterruptCapability interruptcapability);
+
+#if defined(__cplusplus)
+
+#	include <futex.h>
+
+/**
+ * Packaged common design pattern for handlers of edge-triggered interrupts for
+ * seamless transitions between polling and waiting for an interrupt.  As long
+ * as events arrive quickly, this pattern avoids interaction with the scheduler
+ * and interrupt controller.  Only once processing looks to have no more events
+ * to process will the thread call in to the scheduler to atomically notify the
+ * interrupt controller and then sleep until the next event (interrupt) arrives.
+ *
+ * Roughly, this method invokes `body()` every time `check()` returns `true`.
+ * When `check` returns `false`, this method instead enters the scheduler to
+ * atomically, indefinitely wait for the arrival of an interrupt that notifies
+ * `futex`.  If such an interrupt arrives between the invocation of `check()`
+ * and this thread sleeping, the attempted wait will instead immediately return.
+ *
+ * In fact, the return type of `check` is templated; if it is not `bool`, then
+ * the result from `check()` first interpreted by explicit conversion to `bool`
+ * to determine whether or not `body` is invoked and, if `true`, then it is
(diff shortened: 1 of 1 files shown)


nwf:
I struggled a bit with whether this should always do infinite wait or not.  `check()` could `return` a `std::variant<T, TimeoutArgument>` to indicate whether to call `body` or `futex_timed_wait` (with `T = std::monostate` if it has nothing of interest to pass to `body`), but that felt less pleasant for the simple case?  Maybe it will become clearer in use.

I wondered, too, about a `multiwaiter` example, since either "a few IRQs" or "an IRQ and a cancellation notification" seemed plausible.

Last, is this the right place for this kind of thing, or should we have a separate header for such higher-level utilities?

sdk/include/interrupt.h
```diff
@@ -133,3 +133,69 @@ __cheri_compartment("scheduler") const uint32_t *interrupt_futex_get(
...
+ * example, the observed fill level of a FIFO that will be processed in `body`.
+ *
+ * The observed value of the futex is not passed to check or body, as its exact
+ * value is unlikely to be of interest.
+ */
+template<typename T, typename Check, typename Body>
+    requires std::is_nothrow_invocable_r_v<T, Check> &&
+	         requires { static_cast<bool>(std::declval<T>()); }
+__always_inline void
+interrupt_handler_with_futex(const uint32_t *futex, Check check, Body body)
```
davidchisnall: If there are no constraints on body’s type, I’d make it auto.

sdk/include/interrupt.h
```diff
@@ -133,3 +133,69 @@ __cheri_compartment("scheduler") const uint32_t *interrupt_futex_get(
...
+{
+	uint32_t futexValue;
+	T        checkResult;
+
+	while (true)
+	{
+		while ((futexValue = *futex),
```
davidchisnall: This is probably cleaner as a for loop, especially since that allows moving the scope of futexValue to the loop.

sdk/include/interrupt.h
```diff
@@ -133,3 +133,69 @@ __cheri_compartment("scheduler") const uint32_t *interrupt_futex_get(
...
+		       (checkResult = check()),
+		       static_cast<bool>(checkResult))
+		{
+			bool bodyResult;
+
+			if constexpr (std::is_same_v<bool, T>)
```
davidchisnall, line 182: It would be nice for body to be allowed to return void. I think most cases will want body to execute unconditionally, rather than have both check and body allow exit.

sdk/include/interrupt.h
```diff
@@ -133,3 +133,69 @@ __cheri_compartment("scheduler") const uint32_t *interrupt_futex_get(
...
+			{
+				static_assert(std::is_nothrow_invocable_r_v<bool, Body>);
```
davidchisnall: It would be nice if the various static asserts included a message for debugging.

davidchisnall: I think this one is tautological though, it’s repeating things enforced by the requires clause.

sdk/include/interrupt.h
nwf: There are some in-line `static_assert()`-s inside `if constexpr`-s, but I guess those can use `decltype(body)` instead.

PR #768, opened 2026-08-27
utils.hh: round_up and friends

nwf:
These are, like `bytes2bits`, the kind of thing that get holographically / openly coded, but it's probably nice to give them names.

### DIFF
--- a/sdk/core/loader/boot.cc
+++ b/sdk/core/loader/boot.cc
@@ -29,21 +29,6 @@ using namespace CHERI;
 
 namespace
 {
-
-	/**
-	 * Round up to a multiple of `Multiple`, which must be a power of two.
-	 */
-	template<size_t Multiple>
-	constexpr size_t round_up(size_t value)
-	{
-		static_assert((Multiple & (Multiple - 1)) == 0,
-		              "Multiple must be a power of two");
-		return (value + Multiple - 1) & -Multiple;
-	}
-	static_assert(round_up<16>(15) == 16);
-	static_assert(round_up<16>(28) == 32);
-	static_assert(round_up<8>(17) == 24);
-
 	__BEGIN_DECLS
 	static_assert(CheckSize<CHERIOT_LOADER_TRUSTED_SPILL_SIZE,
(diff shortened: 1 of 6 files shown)


sdk/include/utils.hh
```diff
@@ -13,14 +13,17 @@
 #include <stddef.h>
 #include <stdint.h>
 #include <type_traits>
+#include <utility>
 
 namespace utils
 {
+	/// Turn a byte count into a bit count
 	constexpr size_t bytes2bits(size_t in)
```
davidchisnall: Since we're doing cleanup here, can we also fix the name of this.  It reminds me of wondering why someone kept saying nate when he logged off IRC: he was German and was writing nacht.

We shouldn't use 2 to mean to.

We have one caller of this in tree and I hope we have no callers out of tree.

davidchisnall (approved):

sdk/core/loader/boot.cc
```diff
@@ -29,21 +29,6 @@ using namespace CHERI;
...
-	template<size_t Multiple>
-	constexpr size_t round_up(size_t value)
-	{
-		static_assert((Multiple & (Multiple - 1)) == 0,
-		              "Multiple must be a power of two");
-		return (value + Multiple - 1) & -Multiple;
-	}
-	static_assert(round_up<16>(15) == 16);
-	static_assert(round_up<16>(28) == 32);
-	static_assert(round_up<8>(17) == 24);
```
rmn30, line 45: Was this dead anyway!?

sdk/include/utils.hh
```diff
@@ -34,12 +37,58 @@ namespace utils
...
+	 *
+	 * If `value + multiple` overflows their type `T`, the result is
+	 * correct in a modular sense, but likely not useful in practice.
+	 */
+	template<typename T>
+	    requires std::is_unsigned_v<T>
+	constexpr T align_up(T value, T multiple)
+	{
+		return round_up_divide(value, multiple) * multiple;
+	}
```
rmn30, line 72: worth a few `static_asserts` here?

rmn30 (approved):
LGTM.

PR #769, opened 2026-08-28
Proposal: bitpack.hh to include __macro_map.h

nwf:
We hadn't done this from the beginning because, while __macro_map.h is a clever technical stunt, I'd been wary of its use becoming wide-spread, and of course the C preprocessor does not particularly lend itself to modularity or encapsulation.

However, in practice, this has resulted in headers coming to use the features of __macro_map.h without including it, making it a visible dependency in user code, which is unfriendly.

### DIFF
--- a/sdk/include/bitpack.hh
+++ b/sdk/include/bitpack.hh
@@ -287,6 +287,7 @@
  * @{
  */
 
+#include <__macro_map.h>
 #include <cdefs.h>
 #include <concepts>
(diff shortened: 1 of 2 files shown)


davidchisnall (approved):

davidchisnall:
Looks as if the test needs to be removed that checked for the old behaviour.

nwf:
Whoop, yep!

PR #770, opened 2026-09-02
Add math library

jacobpake:
This PR adds a math library containing implementations for a subset of functions normally provided by libm.

The implementations are taken from `freebsd/lib/msun` mostly unmodified, except that `volatile` globals are marked `const`. These globals seem to be marked `volatile` to ensure specific `fenv` related behaviour when e.g. dividing by 0. I don't know if we care about maintaining that behaviour, but marking these as `const` keeps the diff small and (I believe) preserves the intention.

The msun source files expected a number of header files we don't have, I don't know the conventional way to handle this so I have just added empty stubs with those names inside of `sdk/third_party/msun/include`. Any expected definitions which we didn't already provide I have added to `sdk/lib/math/compat.h`

The library currently provides the float and double variants of:

- `exp`
- `exp2`
- `log`
- `log10`
- `log2`
- `pow`
- `sqrt`
- `floor`
- `round`
- `ceil`
- `fmaximum_num`
- `fminimum_num`

These are all* expected to be available by the Rust `coretests` test suite, which runs in our CI, so they should all get *some* coverage. Other test suites will likely expect other libm functions, I plan to add those when we need them, but can add them earlier if preferred.

c7857a9d14290c4926c745b38ea82fdab288f7f9 adds the unmodified freebsd files
fd1dd852838fb09d7bb1f00dd4a83f9c17d5c1f7 applies the necessary changes to the above
c9d2df9d1ca21851b85f20fdf63107d6c4cd3fd6 adds the library and expected headers

*apart from the double variants of `fmaximum_num`/`fminimum_num` and the single variant of `ceil`, which are provided for symmetricalness.


davidchisnall:
Looks good.  Please can you add a README.md to the new directory under `third_party` like [this one](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/main/sdk/third_party/compiler-rt/README.md) including the git commit for FreeBSD that these were copied from, so we know if we need to pull in fixes.

jacobpake:
> Looks good. Please can you add a README.md to the new directory under `third_party` like [this one](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/main/sdk/third_party/compiler-rt/README.md) including the git commit for FreeBSD that these were copied from, so we know if we need to pull in fixes.

Yes, there is already [sdk/third_party/msun/README.md](https://github.com/jacobpake/cheriot-rtos/blob/86d50fc5814064554a3c8c528c057533ed848575/sdk/third_party/msun/README.md)

jacobpake:
Added `ceil`: https://github.com/CHERIoT-Platform/cheriot-rtos/compare/86d50fc5814064554a3c8c528c057533ed848575..c9d2df9d1ca21851b85f20fdf63107d6c4cd3fd6

davidchisnall:
Oh, GitHub's UI made it hard to find.  Thanks!

davidchisnall (approved):

davidchisnall:
Do you have any code-size numbers for these?  We will probably want to split them into smaller libraries at some point.

jacobpake:
> Do you have any code-size numbers for these? We will probably want to split them into smaller libraries at some point.

@davidchisnall, I do not - our test runner is very size sensitive, but I did not find the need to specify either `math32` or `math64` - i.e. it is not prohibitively big for us.

I will need to add another batch of functions here at some point, I will check sizes and split it up then

PR #771, opened 2026-09-04
Refactor KSZ8851 Ethernet SPI driver

nwf:
Cleave out the actual SPI and IRQ handling machinery into a `Provider` class, and provide that class for Sunburst.

This changes the names of some things to better namespace them, but that's meant to be purely syntactic.


davidchisnall:
Looks sensible.  I wonder if we want a `include/platform/external-chips` that is *never* in the default search path for a board, so platform-ethernet.hh can explicitly include <platform/external-chips/{chip on your board}>.

nwf:
I'll hold off on merging until I have a working 2nd Provider type, just to make sure this isn't already overfit.

PR #772, opened 2026-09-16
Add per-compartment-invocation cpu feature control

nwf:
The first few commits are stage setting and tidying arising from "nwf is looking at the switcher again" syndrome (and David's permission to expunge `CONFIG_MSHWM`).

The salient ones are the last two.
"core: per-compartment-invocation cpu features" introduces an opaque `uint16_t` word in each compartment invocation frame.  See the documentation in `sdk/include/switcher.h` for its general behavior.  The actual runtime behavior is dependent on some assembler macros optionally provided by a `platform-switcher_cpu_features.hh` header; see the documentation at the top of `sdk/core/switcher/entry.S` for details.  Absent such a header, the field is unconstrained but also uninterpreted, which is moderately nice for testing.  This commit also adds simple tests of cross-call/-return and cross-thread management of this value.

"ibex: add platform-switcher_cpu_features" does what it says on the tin, and uses the above machinery to expose Ibex's I$ and "data independent timing" features.


sdk/core/loader/boot.cc
```diff
@@ -1087,6 +1085,10 @@ namespace
 			// Special case: The first frame has the initial csp.
 			threadTStack->frames[0].csp = stack;
 
+#if __has_include(<platform-switcher_cpu_features.hh>)
```
davidchisnall: This could probably be simplified by having a generic version that defines this macro as 0.

sdk/core/loader/boot.S
```diff
@@ -6,6 +6,9 @@
 #include "../switcher/trusted-stack-assembly.h"
 #include <cheri-builtins.h>
 
+#if __has_include(<platform-switcher_cpu_features.hh>)
```
davidchisnall: Similarly in these blocks, if generic RISC-V has a version of this file with the macros defined as nothing, we can avoid the include guards. This also provides a hook for documentation.

sdk/core/loader/boot.S
```diff
@@ -183,8 +186,27 @@
...
+	 */
+	li				a1, TrustedStack_size
 	csh				a1, TrustedStack_offset_frameoffset(ca3)
+
+	/*
+	 * If we're on a platform that takes advantage of the switcher's
+	 * platformSpecific functionality, install the default value for our idle
+	 * thread.
+	 */
+#if __has_include(<platform-switcher_cpu_features.hh>)
```
davidchisnall: This can also be unconditional: there’s no harm in a default 0 here.

sdk/core/switcher/entry.S
```diff
@@ -767,6 +864,20 @@ switcher_after_compartment_call:
...
+#if __has_include(<platform-switcher_cpu_features.hh>)
+	/*
+	 * Restore the cpuFeatures value from the caller's frame.  This value is
+	 * presumed already masked, so we need not spend the instructions here.
+	 *
+	 * TODO: It isn't clear whether this would benefit from comparing against
+	 * the current value first.  We might hope that the gateware is doing that
+	 * comparison for us.
+	 */
+	clhu               s0, TrustedStackFrame_offset_cpuFeatures(ct1)
```
davidchisnall: If this load is folded into the macro, which takes the address expression as a third argument, we can remove the preprocessor conditionals and just have the default implementation be nothing.

sdk/include/platform/ibex/platform-switcher_cpu_features.hh
```diff
@@ -0,0 +1,47 @@
+// Copyright SCI Semiconductor and CHERIoT Contributors.
```
davidchisnall, line 1: I think I would prefer splitting this into two, one with a way of determining at compile time what features exist for the target CPU and how to set / restore them, one that handles the implementation, so the one users include is distinct from the one the loader and switcher include.

sdk/include/switcher.h
```diff
@@ -12,9 +12,6 @@
 /**
  * Returns true if the trusted stack contains at least `requiredFrames` frames
  * past the current one, false otherwise.
- *
```
davidchisnall, line 15: Why is this removed?

tests/compartment_calls-test.cc
```diff
@@ -6,6 +6,7 @@
 #include "compartment_calls.h"
 #include "tests.hh"
 #include <cheri.hh>
+#include <cheriot-atomic.hh>
```
davidchisnall: This file is for legacy compatibility, it shouldn’t be used in new code.

tests/compartment_calls-test.cc
```diff
@@ -44,6 +45,134 @@ void test_number_of_arguments()
...
+		TEST(switcher_invocation_cpu_features_set(-1, 0) == featuresAll,
+		     "switcher invocation CPU features: bad state after call w/ all");
+
+		/*
+		 * Check that a cross-thread switch picks up defaults for that thread,
+		 * (rather than our "all" value, if that's different) and comes back to
+		 * us with our "all" still OK.
+		 */
+
+		static cheriot::atomic<uint32_t> state  = 0;
```
davidchisnall: std::atomic

davidchisnall (commented):
Both of the locations we store this have space for a uint32_t, which would give us a bit more headroom in the future. Was there a reason for uint16_t?

sdk/core/loader/boot.cc
nwf: Sure, can do.

sdk/include/switcher.h
nwf: Those functions no longer exist?

nwf:
> Both of the locations we store this have space for a uint32_t, which would give us a bit more headroom in the future. Was there a reason for uint16_t?

There's just the one, since I think we can track it like errorHandlerCount.

But as to `uint16_t`, mostly an aversion to committing bits from working on the allocator.  The _interfaces_ to the application are all in terms of `uint32_t` and it would be straightforward to update the switcher to use more later.

sdk/include/switcher.h
davidchisnall: Oh, huh.  Ooops.

sdk/core/switcher/entry.S
nwf: I took a stab at that (and also passed the "rest of the load mnemonic" as an argument to make changing to a uint32_t-s worth of state if needed later easier).

davidchisnall (approved):
It would be nice to have a test that runs in Ibex SAFE only that checks that divide by 1 takes a different number of cycles with and without the constant time things set, so that we know we're actually setting them.

compile_flags.txt
rmn30, line 45: Should this be checked in at all? What uses it?

davidchisnall: This file?  clang-tidy uses it in CI.  This predefined macro?  Nothing anymore.

sdk/core/loader/boot.S
```diff
@@ -183,7 +183,13 @@
...
 	csetaddr		ca3, ca0, a3
 	csetboundsexact	ca3, ca3, a1
-	li				a1, TSTACKOFFSET_FIRSTFRAME + TrustedStackFrame_size
+	/*
+	 * Fib to the switcher that we've been created by the loader and so have an
+	 * active 0-th TrustedStackFrame.  This matters only for invoking the
+	 * initializers below and keeps the switcher from taking a cross-return out
+	 * of the 0th frame as a thread exit.
+	 */
+	li				a1, TrustedStack_size
```
rmn30, line 192: I'm not sure what's going on here. Why did we lost TSTACKOFFSET_FIRSTFRAME from the add? This is a functional change?

sdk/core/switcher/trusted-stack.hh
```diff
@@ -87,14 +86,11 @@ struct TrustedStackGeneric
...
-#ifdef CONFIG_MSHWM
-	  12
-#else
-	  4
-#endif
-	];
+	/**
+	 * Pad back up to alignof(void *)
+	 */
+	uint32_t pad;
```
rmn30, line 92: Do we have a static assert for this? We require that offsetof(TrustedStack, frames) % alignof(void*) == 0?

rmn30:
Nice! Just minor queries.

sdk/core/loader/boot.S
nwf: It should not be a functional change.

The prior add was of `TSTACKOFFSET_FIRSTFRAME + TrustedStackFrame_size`, which was, IMHO, more confusing than just spelling out the components.  `TSTACKOFFSET_FIRSTFRAME` was, specifically, `(TrustedStack_offset_frameoffset + TSTACK_HEADER_SZ)`, with `TSTACK_HEADER_SZ` being the amount of space (occupied by `threadID` and `padding`) between `TrustedStack::frameoffset` and `TrustedStack::frames[0]`.  All told, that computes the offset of `TrustedStack::frames[1]`, which is also spelled `TrustedStack_size`.

IMHO `TSTACK_HEADER_SZ` and `TSTACKOFFSET_FIRSTFRAME` were ugly names and unchecked duplication of information already in the C++ structure.

sdk/core/switcher/trusted-stack.hh
nwf: Yes we have an assert for that, if indirectly.  `alignof(TrustedStackFrame)` is `alignof(void*)` from its contents, and `static_assert(std::has_unique_object_representations_v<TrustedStack>);` is going to require that there be no anonymous space between the end of `TrustedStack::pad` and the beginning of `TrustedStack::frames[0]`.

nwf:
Writing test; will push when it succeeds, hopefully momentarily.

sdk/core/loader/boot.S
rmn30: After conversation with Wes this works because `TrustedStack_size` means `sizeof(TrustedStack<1>)` where the template parameter is the number of frames allocated. This is the same as `offset(TrustedStack<1>, frames[1])`. I was also confused because I hadn't noticed the change from `TrustedStackFrame_size` to `TrustedStack_size`

nwf:
Who knew, writing tests reveals things.  Ibex, in particular, can be built without support for an I$ or data independent timing, so even though Ibex-the-platform supports these things, Ibex-an-instance-thereof may not.

As a result, I have overhauled this to not just write to the CSRs but also _read back_ from them to see what actually happened.  It turned out not to be possible to fold all the memory accesses into macro arguments, so we're back to `#if __has_header(<....inc>)` checks in a few places, sorry.

It _is_ possible to build a CHERIoT-Ibex with data independent timing support, at least for its division-by-zero logic, that boots, but attempting to set `SecureIbex` didn't work.  I resorted to this abomination.
```
diff --git a/rtl/ibex_core.sv b/rtl/ibex_core.sv
index 88999ea1..c13c2976 100644
--- a/rtl/ibex_core.sv
+++ b/rtl/ibex_core.sv
@@ -727,7 +727,7 @@ module ibex_core import ibex_pkg::*; import cheri_pkg::*; #(
     .priv_mode_i          (priv_mode_id),
     .csr_mstatus_tw_i     (csr_mstatus_tw),
     .illegal_csr_insn_i   (illegal_csr_insn_id),
-    .data_ind_timing_i    (data_ind_timing),
+    .data_ind_timing_i    (1'b0),
     .csr_pcc_perm_sr_i    (pcc_cap_r.perms[PERM_SR]),
 
     // LSU
@@ -1429,7 +1429,7 @@ end
   ibex_cs_registers #(
     .DbgTriggerEn     (DbgTriggerEn),
     .DbgHwBreakNum    (DbgHwBreakNum),
-    .DataIndTiming    (DataIndTiming),
+    .DataIndTiming    (1'b1),
     .DummyInstructions(DummyInstructions),
     .ShadowCSR        (ShadowCSR),
     .ICache           (ICache),
```

With that gross hack in place, we can see the effects in the test suite:
```
Compartment calls (main runner) test: switcher invocation CPU features: entry=0x0 zero=0x0 all=0x2                    
[...]
Test misc APIs test: Ibex data independent timing probe: with(0x2)=0x28 without(0x0)=0x5
```
whereas without (and without the I$, since SAFE doesn't enable that) we skip most tests on the values:
```
Compartment calls (main runner) test: switcher invocation CPU features: entry=0x0 zero=0x0 all=0x0
Compartment calls (main runner) test: switcher invocation CPU features: skip tests since all=0
[...]
Test misc APIs test: Ibex data independent timing probe: with(0x0)=0x5 without(0x0)=0x5                             
Test misc APIs test: Ibex not built with data independent timing support?
```

On the sail model, which does not provide a `platform-switcher_cpu_features.inc` file, we instead have all the bits there but without any meaning:
```
Compartment calls (main runner) test: switcher invocation CPU features: entry=0x0 zero=0x0 all=0xffff
```

